PR #4, opened 2015-03-24
Relax Poison version requirement

jasonwebster:
The public API hasn't changed enough to matter here. There's no reason to stay on 1.2.x as far as I can tell.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -22,7 +22,7 @@ defmodule Raven.Mixfile do
 
   defp deps do
     [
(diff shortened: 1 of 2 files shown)


jasonwebster:
The build is going to fail, but it does on `master` too...

johnhamelink:
Any reason why this hasn't been merged

b-murphy:
+1 on merge

HashNuke:
Even hackney requirement needs to be relaxed.

Issue #11, opened 2016-08-03
Refactor API Client into own module


jeregrine:
- Handle errors gracefully https://docs.getsentry.com/on-premise/clientdev/#handling-failures
- Handle DSN configuration entirely 
- Request Encoding
- Response Handling

mitchellhenke:
@jeregrine should we call this done? or do we want to wait for retries

jeregrine:
Yea

PR #20, opened 2016-08-04
Separate api client

mitchellhenke:
Will probably have to sort this out a bit better, as far as which module handles auth headers, etc., but  this is a good start for #11

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-elixir/20)

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -63,25 +63,6 @@ defmodule Sentry do
     {endpoint, public_key, secret_key}
   end
 
-  @sentry_version 5
-  quote do
-    unquote(@sentry_client "sentry-elixir/#{Mix.Project.config[:version]}")
-  end
-
-  @doc """
-  Generates a Sentry API authorization header.
-  """
-  @spec authorization_header(String.t, String.t, Integer.t) :: String.t
-  def authorization_header(public_key, secret_key, timestamp \\ nil) do
-    # X-Sentry-Auth: Sentry sentry_version=5,
-    # sentry_client=<client version, arbitrary>,
-    # sentry_timestamp=<current timestamp>,
-    # sentry_key=<public api key>,
-    # sentry_secret=<secret api key>
-    timestamp = if timestamp, do: timestamp, else: unix_timestamp
-    "Sentry sentry_version=#{@sentry_version}, sentry_client=#{@sentry_client}, sentry_timestamp=#{timestamp}, sentry_key=#{public_key}, sentry_secret=#{secret_key}"
-  end
-
   @doc """
   Parses and submits an exception to Sentry if DSN is setup in application env.
   """
@@ -89,7 +70,9 @@ defmodule Sentry do
   def capture_exception(exception) do
     case Application.get_env(:sentry, :dsn) do
       dsn when is_bitstring(dsn) ->
-        capture_exception(exception |> transform, dsn |> parse_dsn!)
+        parsed_dsn = parse_dsn!(dsn)
+        transform(exception)
+        |> capture_exception(parsed_dsn)
       _ ->
         :error
     end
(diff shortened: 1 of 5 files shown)


lib/sentry.ex
```diff
@@ -229,13 +207,7 @@ defmodule Sentry do
...
 
-  @spec unix_timestamp :: Number.t
-  defp unix_timestamp do
-    {mega, sec, _micro} = :os.timestamp()
-    mega * (1000000 + sec)
-  end
-
-  @spec unix_timestamp :: String.t
+  @spec iso8601_timestamp :: String.t
   defp iso8601_timestamp do
```
jeregrine: Use the elixir DateTime methods http://elixir-lang.org/docs/stable/elixir/DateTime.html

lib/sentry/client.ex
```diff
@@ -0,0 +1,42 @@
...
+    unquote(@sentry_client "sentry-elixir/#{Mix.Project.config[:version]}")
+  end
+
+  def request(method, url, headers, body) do
+    case :hackney.request(method, url, headers, body, []) do
+      {:ok, 200, _headers, client} ->
+        case :hackney.body(client) do
+          {:ok, body} ->
+            id = Poison.decode!(body)
+                  |> Dict.get("id")
```
jeregrine: Use Map

lib/sentry/client.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Sentry.Client do
+  @sentry_version 5
+  quote do
+    unquote(@sentry_client "sentry-elixir/#{Mix.Project.config[:version]}")
+  end
+
+  def request(method, url, headers, body) do
+    case :hackney.request(method, url, headers, body, []) do
+      {:ok, 200, _headers, client} ->
+        case :hackney.body(client) do
```
jeregrine: Break this up into multiple functions or use HTTPotion or something

lib/sentry.ex
```diff
@@ -57,39 +57,24 @@ defmodule Sentry do
...
-    "Sentry sentry_version=#{@sentry_version}, sentry_client=#{@sentry_client}, sentry_timestamp=#{timestamp}, sentry_key=#{public_key}, sentry_secret=#{secret_key}"
-  end
-
   @doc """
   Parses and submits an exception to Sentry if DSN is setup in application env.
   """
   @spec capture_exception(String.t) :: {:ok, String.t} | :error
   def capture_exception(exception) do
+    # TODO: better environment handling
     case Application.get_env(:sentry, :dsn) do
```
jeregrine: Move this to a module attribute, and possibly even do it in the Client.

lib/sentry.ex
```diff
@@ -57,39 +57,24 @@ defmodule Sentry do
   def parse_dsn!(dsn) do
```
jeregrine: Probably parse the DSN in the client.

lib/sentry/client.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Sentry.Client do
+  @sentry_version 5
+  quote do
+    unquote(@sentry_client "sentry-elixir/#{Mix.Project.config[:version]}")
+  end
+
+  def request(method, url, headers, body) do
```
jeregrine: Since you can only post I'd probably just do a function that posts the body. I'd also have this function turn it into json.

jeregrine:
Basically I'd move everything the API client cares about into the API client. Examples:
- JSON de/encoding
- DSN parsing and usage

lib/sentry.ex
mitchellhenke: this is configurable probably at runtime unfortunately

mitchellhenke:
@jeregrine part of the issue is that it uses the existence of the DSN config to send or not send

lib/sentry.ex
mitchellhenke: done

PR #29, opened 2016-08-19
Added Docs, Ran Credo

jeregrine:
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-elixir/29)

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -9,7 +9,7 @@ defmodule Sentry.Client do
   end
 
   @moduledoc """
-    TODO
+    Provides basic HTTP client request and response handling for the Sentry API. 
   """
 
   @spec send_event(%Event{}) :: {:ok, String.t} | :error
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -8,10 +8,13 @@ defmodule Sentry.Client do
     unquote(@sentry_client "sentry-elixir/#{Mix.Project.config[:version]}")
   end
 
+  @moduledoc """
+    TODO
+  """
+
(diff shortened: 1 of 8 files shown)


lib/sentry/client.ex
```diff
@@ -8,10 +8,13 @@ defmodule Sentry.Client do
...
 
+  @moduledoc """
+    Provides basic HTTP client request and response handling for the Sentry API. 
+  """
+
   @spec send_event(%Event{}) :: {:ok, String.t} | :error
   def send_event(%Event{} = event) do
-    {endpoint, public_key, secret_key} = Application.fetch_env!(:sentry, :dsn)
-                                          |> parse_dsn!
+    {endpoint, public_key, secret_key} = parse_dsn!(Application.fetch_env!(:sentry, :dsn))
```
mitchellhenke: I guess I don't understand credo's issue with this.  piping the result of a function into another function is readable enough?

lib/sentry/client.ex
```diff
@@ -25,11 +28,12 @@ defmodule Sentry.Client do
       {:ok, 200, _headers, client} ->
         case :hackney.body(client) do
           {:ok, body} ->
-            id = Poison.decode!(body)
-                  |> Map.get("id")
+            id = body
+              |> Poison.decode!()
```
mitchellhenke: darn, elixir.vim indents against the equal sign like 

```
            id = body
                  |> Poison.decode!()
                  |> Map.get("id")
```

lib/sentry/client.ex
jeregrine: Eh I don't mind either.

PR #35, opened 2016-08-24
add plug tests

mitchellhenke:
@jeregrine

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-elixir/35)

### DIFF
--- a/test/plug_test.exs
+++ b/test/plug_test.exs
@@ -19,8 +19,11 @@ defmodule Sentry.PlugTest do
   test "exception makes call to Sentry API" do
    bypass = Bypass.open
    Bypass.expect bypass, fn conn ->
-     assert "/api/1/store/" == conn.request_path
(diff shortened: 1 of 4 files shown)


test/plug_test.exs
```diff
@@ -0,0 +1,44 @@
...
+    get "/error_route" do
+      _ = conn
+      raise RuntimeError, "Error"
+    end
+  end
+
+  test "exception re-raises" do
+    exception = %RuntimeError{message: "Error"}
+
+   conn = conn(:get, "/error_route")
```
jeregrine: Can you do

```
assert_raises RuntimeError, fn ->
  ExampleApp.call(conn, [])
  |> conn(:get, "/error_route")
end
```

mitchellhenke: ah, good idea

PR #40, opened 2016-08-25
Updating Docs and Config

### LINKED ISSUE #12: Setup Configuration
- release
- environment
- user context
- tags context
- contexts

### LINKED ISSUE #33: Make included_environments, :environment_name not required

jeregrine:
Closes #33
Closes #12

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-elixir/40)

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -60,8 +60,12 @@ defmodule Sentry do
   end
 
   def send_event(event = %Event{}) do
-    included_environments = Application.fetch_env!(:sentry, :included_environments)
-    if Application.fetch_env!(:sentry, :environment_name) in included_environments do
+    included_environments = case Application.fetch_env(:sentry, :included_environments) do
+      {:ok, e} -> e
+      :error -> ~w(prod dev test)a
+    end
+
+    if event.environment in included_environments do
       Client.send_event(event)
     else
       {:ok, ""}
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -60,10 +60,7 @@ defmodule Sentry do
   end
 
   def send_event(event = %Event{}) do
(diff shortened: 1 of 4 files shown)


lib/sentry/event.ex
```diff
@@ -52,15 +54,33 @@ defmodule Sentry.Event do
       |> String.trim("*")
       |> String.trim
 
+    release = case Application.fetch_env(:sentry, :release) do
```
mitchellhenke: I think we can use `Application.get_env(:sentry, _)`.  should return nil if not set

lib/sentry.ex
```diff
@@ -60,8 +60,12 @@ defmodule Sentry do
   end
 
   def send_event(event = %Event{}) do
-    included_environments = Application.fetch_env!(:sentry, :included_environments)
-    if Application.fetch_env!(:sentry, :environment_name) in included_environments do
+    included_environments = case Application.fetch_env(:sentry, :included_environments) do
```
mitchellhenke: similar to below, I think we can simplify to this:

```elixir
included_environments = Application.get_env(:sentry, :included_environments, ~w(test dev prod))
```

lib/sentry/event.ex
```diff
@@ -52,15 +54,33 @@ defmodule Sentry.Event do
...
+      {:ok, r} -> r
+      :error -> nil
+    end
+
+    server_name = case Application.fetch_env(:sentry, :server_name) do
+      {:ok, r} -> r
+      :error -> nil
+    end
+
+    env = case Application.fetch_env(:sentry, :environment_name) do
```
mitchellhenke:
```elixir
env = Application.get_env(:sentry, :environment_name, String.to_atom System.get_env("MIX_ENV"))
```

I think we would also have to convert to atom, as environment variables will always be strings.

PR #56, opened 2016-09-08
Small updates to the docs

jeregrine:
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-elixir/56)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ try do
   ThisWillError.reall()
 rescue
   my_exception ->
-    Sentry.capture_exception(my_exception)
+    Sentry.capture_exception(my_exception, [stacktrace: System.stacktrace(), extra: %{extra: information}])
 end
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,10 +32,20 @@ config :sentry,
...
+
+Sometimes you want to capture specific exceptions, to do so use the `Sentry.capture_exception/3`.
+
 ```elixir
-Sentry.capture_exception(my_exception)
+try do
+  ThisWillError.reall()
+rescue
+  my_exception ->
+    Sentry.capture_exception(my_exception)
```
mitchellhenke: should add `System.stacktrace()` to the second arg, and maybe some optional stuff in the last arg to show a fuller example?

jeregrine: @mitchellhenke maybe in capture_exception we should grab System.stacktrace() if its null?
https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/event.ex#L45

mitchellhenke: oops, I got it wrong.  It would be `Sentry.capture_exception(my_exception, [stacktrace: System.stacktrace(), extra: %{extra: information}])`

Stacktraces are not super intuitive, so I'm not sure if we should refer to other documentation on them, or write some of our own.

PR #66, opened 2016-09-25
set "a" at the end of sigil w to make it atom

KazuCocoa:
Currently, `included_environments:` is not an atom in the following document.
https://docs.sentry.io/hosted/clients/elixir/#configuration

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-elixir/66)


mitchellhenke:
Thanks!

PR #70, opened 2016-10-07
Update to reflect apparent documentation error

alistairstead:
The documentation suggested to only add config to `config/prod.ex` however to run the mix tasks that are also documented or to run a local dev environment or tests then default DSN values are needed even if the service is disabled in all environments other than prod.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ defp deps do
 end
 ```
 
-Setup the application environment in your `config/prod.exs`
+Setup the application environment in your `config/conf.exs` you may overload these values or add additional values in other environment files.
 
 ```elixir
 config :sentry,


jeregrine:
Great! Could you also fiy. x this here https://github.com/getsentry/sentry-elixir/blob/master/docs/index.rst#configuration and I will merge right aw

alistairstead:
The config gets complex if you move any to `config/config.ex` as even though sentry is not active in other environments it attempts to load config and raises errors.

Might be worth keeping the config only in `prod.ex` but then limiting the `:sentry` application to just prod?

```elixir
{:sentry, "~> 1.0.0", only: [:prod]}
```

mitchellhenke:
It looks like documentation for this has been updated, so I'm going to close this for now.  Always happy to improve the docs if something is still unclear though 🙂

PR #72, opened 2016-10-15
Use Mix.env to get mix environment

alexgaribay:


mitchellhenke:
We're changing the way `environment_name` is handled to hopefully make it simpler and easier to work with in #76.  It's unlikely we'll use `Mix.env`, as the application may not be running in a Mix environment.

If you have any feedback on #76, would be happy to hear it :)

Thanks!

PR #77, opened 2016-10-19
better error logging

### LINKED ISSUE #69: `Failed to send sentry event` does not provide enough detail
I was trying sentry out in my project and got this in my logs:

```
13:24:48.123 module=Sentry.Client [error] Failed to send sentry event.
# ...
```

Problem is, it does not tell me why it failed or give me any clue about it.  I initially assumed it was a problem with my dsn, but that was correct.  Eventually, I edited the code in `Sentry.Client.request/4` to have it capture the unmatched values (instead of ignoring them with `_`) and print them with `IO.inspect`.  The result was included this line:

```
{"X-Sentry-Error", "Creation of this event was denied due to rate limiting"},
```

...which clearly showed me the problem.  Another one of our projects was reporting a lot of errors in a short period, triggering rate limiting.

It would be really nice if users did not have to edit the code of this library to figure this out.  Can `log_api_error` be updated to include some detail about what the failure was?

mitchellhenke:
@jeregrine 

closes #69

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -48,7 +48,7 @@ defmodule Sentry.Client do
     end
   end
 
-  defp request(method, url, headers, body) do
+  def request(method, url, headers, body) do
     case :hackney.request(method, url, headers, body, []) do
       {:ok, 200, _headers, client} ->
         case :hackney.body(client) do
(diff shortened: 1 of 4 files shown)


jeregrine:
LGTM

PR #78, opened 2016-10-20
Add more documentation about environment configuration

DavidAntaramian:
I needed to provide a custom environment name for sentry that is different from a static value that I could assign in `prod.exs`. I discussed what I needed briefly with @mitchellhenke, I figured would document what I'm doing.

I also took the liberty of shifting the documentation block two spaces left so that it is not indented in `@moduledoc`. When it is indented, the Markdown is not picked up and it is instead output as plain text.

I also reduced the header level by a factor of one since there was no other documentation at the h2 level.

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -6,28 +6,60 @@ defmodule Sentry do
 
 
   @moduledoc """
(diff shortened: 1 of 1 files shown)


mitchellhenke:
💖 ❤️

Issue #99, opened 2016-12-09
Timestamp parameter is invalid and ignored


doughsay:
Just updated to sentry-elixir 2.0.2, and my events are getting this error when displayed in the sentry UI.  Also, the "First Seen" and "Last Seen" are always "n/a".  See screenshot:

![screen shot 2016-12-08 at 5 56 13 pm](https://cloud.githubusercontent.com/assets/221693/21034888/0f82231c-bd70-11e6-96e4-214b573c202f.png)

I assume the two are related...

Also, I don't know if this has anything to do with version 2.0.2, I think it's just been not working the whole time.  But, I only recently started getting that red bar at the top of the sentry ui.

mitchellhenke:
I believe this was fixed as much as the library can fix the issue in #100.  If you continue to have problems, please just open another issue, and I'll see what can be done on client or server side. 🙂

doughsay:
Thanks!  Sorry for my lack of responses; I'm confident this is all we can do from the library side as well.  I'll keep my eye out on the misbehaving events to see if I can figure anything else out.  So far, things seem to have magically fixed themselves.  Thanks again!

mitchellhenke:
@doughsay no worries, thanks for reporting this!

PR #100, opened 2016-12-09
Fix timestamp format to match the SDK docs.

doughsay:
The SDK docs say the timestamp should not have a timezone, and the example timestamp doesn't include microseconds: https://docs.sentry.io/clientdev/attributes/

This pull fixes the format to match what the docs describe.  Addresses https://github.com/getsentry/sentry-elixir/issues/99.

### DIFF
--- a/lib/sentry/util.ex
+++ b/lib/sentry/util.ex
@@ -12,11 +12,13 @@ defmodule Sentry.Util do
   end
 
   @doc """
-    Generates a iso8601_timestamp
+    Generates a iso8601_timestamp without microseconds and timezone
   """
   @spec iso8601_timestamp :: String.t
   def iso8601_timestamp do
     DateTime.utc_now()
+    |> Map.put(:microsecond, {0, 0})
     |> DateTime.to_iso8601()
+    |> String.trim_trailing("Z")
   end
 end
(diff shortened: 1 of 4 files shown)


mitchellhenke:
Thanks for opening this issue.  It looks like the time is correct in some way, as it shows when it was sent:

image

I tested the changes made here, and was still unable to get "first seen" to show up.
This _may_ be an issue with the Sentry server?

doughsay:
Well that's disappointing... I was hoping this would fix it.  Does it at least cause the timestamp field to be accepted by the server?  See #99 to see my screenshot of the server UI indicating that the timestamp field was ignored.  If it at least fixes that then this pull should still be valuable.

mitchellhenke:
I haven't seen that error myself yet, and wasn't able to get it using the old format or the new format.  Does it happen every time for you?

The changes are still valuable, as I'd rather we follow what the docs say is the correct format.  Since it persists, it may be an upstream issue within Sentry itself.

doughsay:
I'm not consistently getting the error, no.  Funny enough, I just enabled release tracking in my app, and now the "first seen" and "last seen" are working...  but I still get the error at the top warning me that timestamp was disregarded.

PR #104, opened 2017-01-10
Add structs to context

olafura:
What is acceptable to send is mostly missing in regards to context.
This change helps a bit with that to at least to have structs and
types for two context items, user and breadcrumbs.

### DIFF
--- a/lib/sentry/context.ex
+++ b/lib/sentry/context.ex
@@ -57,8 +57,9 @@ defmodule Sentry.Context do
     Process.get(@process_dictionary_key) || %{}
   end
 
-  def add_breadcrumb(map) when is_map(map) do
-    map = Map.put_new(map, "timestamp", Sentry.Util.unix_timestamp())
+  @spec add_breadcrumb(Sentry.Context.BreadCrumb.t) :: nil
+  def add_breadcrumb(%Sentry.Context.BreadCrumb{} = map) do
+    map = Map.put(map, :timestamp, Sentry.Util.unix_timestamp())
     current_context = get_context()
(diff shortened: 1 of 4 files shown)


olafura:
It would be good to leave the line numbers out of the comparison. Because it breaks the tests. I can look at it if you want.

mitchellhenke:
Thanks for opening this!

The structs for giving some structure to the data is great, but I think it may only work for breadcrumbs.  Per [this documentation](https://docs.sentry.io/learn/context/),  User context allows you to set arbitrary keys/values, so it would have to remain a map.

I'm also not sure we want to add a dependency for deep merging.  Is there an issue with merging context maps currently?

olafura:
The deep merge was only needed for User struct and only because I only implemented merging structs with it and not normal merge. But I'll modify the patch to drop the User Struct and the dependency ;)

lib/sentry/context.ex
```diff
@@ -57,8 +57,9 @@ defmodule Sentry.Context do
     Process.get(@process_dictionary_key) || %{}
   end
 
-  def add_breadcrumb(map) when is_map(map) do
-    map = Map.put_new(map, "timestamp", Sentry.Util.unix_timestamp())
+  @spec add_breadcrumb(Sentry.Context.BreadCrumb.t) :: nil
+  def add_breadcrumb(%Sentry.Context.BreadCrumb{} = map) do
```
mitchellhenke: After some thinking, I'm not sure if we want to pattern match on the struct, as it will break backwards compatibility since people are using maps currently.  What do you think would be the best way to work around that?

Issue #107, opened 2017-01-14
Load dynamic values from system env


rafaeljesus:
The current [code](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/client.ex#L94) does not load dynamic env var using `:system`

```elixir
config :sentry,
  dsn: {:system, "SENTRY_DNS"},
```

This is useful when deploying in production environment using [exrm](https://github.com/bitwalker/exrm).

I would appreciate this feature asap as I am not receiving errors on my Sentry account or if you have any other suggestion please let me now,

Thank guys

mitchellhenke:
I don't have a huge amount of experience with exrm/distillery, so please correct me if I'm off.  My understanding is that for libraries supporting the `{:system, "CONFIG_KEY"}` convention, when the release is built, it will store that tuple normally without fetching from System. Upon the release being run, the library will see that pattern, and read from the config at runtime.

I believe Phoenix only supports it for a couple parts of the configuration (PORT and HOST).  Would adding it for the dsn be sufficient, or would it be necessary for all configuration keys?

rafaeljesus:
@mitchellhenke Thank you for your reply! Yes that's the behavior mate. I also request it on [ex_statsd](https://github.com/CargoSense/ex_statsd/pull/32) I can totally submit a PR you you want, wdyt?

mitchellhenke:
@rafaeljesus cool, sounds good.  I can likely get to it this week.

PR #109, opened 2017-01-19
Read dsn from system

rafaeljesus:
Allow read dsn system values for deployment

Closes https://github.com/getsentry/sentry-elixir/issues/107

I kindly ask you to merge it asap as I have a application running in production waiting this feature,

Thanks

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -1,7 +1,7 @@
 defmodule Sentry.Client do
   alias Sentry.{Event, Util}
   require Logger
-  @type parsed_dsn :: {String.t, String.t, Integer.t}
+  @type get_dsn :: {String.t, String.t, Integer.t}
   @sentry_version 5
   @max_attempts 4
 
@@ -21,7 +21,7 @@ defmodule Sentry.Client do
   """
   @spec send_event(%Event{}) :: {:ok, String.t} | :error
   def send_event(%Event{} = event) do
-    {endpoint, public_key, secret_key} = dsn_env |> parse_dsn!
+    {endpoint, public_key, secret_key} = get_dsn!
 
     auth_headers = authorization_headers(public_key, secret_key)
     body = Poison.encode!(event)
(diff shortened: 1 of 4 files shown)


lib/sentry/client.ex
```diff
@@ -21,7 +21,7 @@ defmodule Sentry.Client do
   """
   @spec send_event(%Event{}) :: {:ok, String.t} | :error
   def send_event(%Event{} = event) do
```
rafaeljesus: Basically we should be injecting dependences in function signature so we can easily test and move private functions to its own module which we can also test independently so it could be like:

```elixir
def send_event(%Event{} = event, dsn_parser \\ DsnParser) do
{endpoint, public_key, secret_key} = dsn_parser.parse!
# ....
end
```

Unfortunately I don't ave time to implement it yet, but we can refactor it soon

lib/sentry/client.ex
```diff
@@ -31,6 +31,13 @@ defmodule Sentry.Client do
     end)
   end
 
+  def dsn_env do
```
rafaeljesus: Unfortunately it is public for testing, this should be moved to another module `DsnParser` and be injected on `send_event` public function

mitchellhenke:
I can start reviewing this at some point today, but if you need it ASAP, you can reference your fork in mix with: `{:sentry, git: "git://github.com/rafaeljesus/sentry-elixir.git", branch: "read-dsn-from-system"}`

rafaeljesus:
@mitchellhenke Thanks!

lib/sentry/client.ex
mitchellhenke: I'm not sure that that would be necessary.  This kind of pattern would usually come along with specifying a Behaviour, which might be overkill at this point 🙂

lib/sentry/client.ex
```diff
@@ -31,6 +31,13 @@ defmodule Sentry.Client do
...
+    case Application.get_env(:sentry, :dsn) do
```
mitchellhenke: I think this should stay as `Application.fetch_env!`?

rafaeljesus: Yes you are right!

rafaeljesus:
@mitchellhenke I also just realized that I need to do the same on `Mix.Tasks.Sentry.SendTestEvent`

mitchellhenke:
@rafaeljesus It might be easier to have something like `Sentry.Client.get_dsn!` that fetches from sys/env and parses.  Then it can just be called in `Sentry.Tasks.Sentry.SendTestEvent` and `Sentry.Client.send_event/1`

rafaeljesus:
@mitchellhenke mate I did the changes you requested may you take a 👀  again?

rafaeljesus:
Also execute the send_test_event manually with my sentry account and works like a charm

mitchellhenke:
Thanks for the contribution!

PR #98, opened 2016-12-08
Filter Events

mitchellhenke:
closes #63  

@jeregrine 

going to write some docs if this API works

PR #110, opened 2017-02-01
Ignore non-existent route exceptions

colinrymer:
404s aren't exceptional, so there's no need for us to use up bandwidth on something that we
don't need to record in Sentry if we can easily avoid it.

Two concerns I have that should probably be addressed before this is considered ready to merge:

1) This doesn't test the Phoenix logic. I'm hesitant to include Phoenix (even as a dev dependency) solely for a single test, but I also don't feel comfortable leaving this without test coverage.

2) This is a breaking change. I'm not sure if I should wrap this in a config flag that keeps it from breaking existing behavior by default or if you want to advocate people use this (my preference since this reduces resource usage for all parties) and bump the major version to reflect the change in behavior.

I'm happy to address both those concerns, but I want to address them as you all see fit.


mitchellhenke:
Filtering exceptions is something we added recently in #98.  The documentation is here: https://hexdocs.pm/sentry/Sentry.html#module-filtering-exceptions

The above functionality should cover this use case, though it could be easier to set up.  It may be worthwhile to include a default filter for Phoenix within the library, so it's easier to set up a better "default" for Phoenix applications.

Sentry could include something like: 
```elixir
defmodule Sentry.DefaultPhoenixFilter do
  @behaviour Sentry.EventFilter

  if Code.ensure_loaded?(Phoenix) do
    def exclude_exception?(:plug, %Elixir.Phoenix.Router.NoRouteError{}), do: true
  end

  def exclude_exception?(_, _), do: false
end
```

and users wanting to use it would then only have to alter their config as such:

```elixir
config :sentry, filter: Sentry.DefaultPhoenixFilter
# etc.
```

My naming of modules isn't super great, but do you think something like that would suffice?  It's not as easy as dropping in Sentry to a Phoenix app and having it exclude NoRouteError automatically, but it's potentially better than what we have currently?  Would love to hear your thoughts 🙂

colinrymer:
I'm not opposed to using filters to handle this scenario, but the current filter implementation leaves me with some reservations.

* As you mentioned, it's not a "drop it in and use it" solution.
* The filtering implementation only supports a single filter module, meaning users would have to reimplement the logic around filtering 404s if they wanted to filter additional exceptions as well.
* I'm slightly concerned about performance - using pattern matching on method dispatch to determine whether to exclude something should be much faster than the several function a filter solution requires.

I certainly think making the dropping of 404s optional is good thing (in the name of backwards compatibility), but I do think it should be something that's easily used out of the box, hence this approach.

mitchellhenke:
>As you mentioned, it's not a "drop it in and use it" solution.

There is some difficulty in balancing drop-in components like `Sentry.Plug` and `Sentry.Logger` that live within the Sentry library while also trying to offer some configurability from outside the library.  It is certainly possible to include features like those in the PR within `Sentry.Plug`, though a line does have to be drawn _somewhere_.  I am inclined to include these changes, as it's likely the most popular use case for this library currently, but I'll ask around a bit first.

>The filtering implementation only supports a single filter module, meaning users would have to reimplement the logic around filtering 404s if they wanted to filter additional exceptions as well.

You are correct, and I'd agree that my initial suggestion is not ideal.  For those wishing to have the greatest degree of control over how/when to report exceptions, it would probably be best to build one's own modules (perhaps modeled after the default implementations) within the context of their application and use `Sentry.capture_exception/2` directly.

>I'm slightly concerned about performance - using pattern matching on method dispatch to determine whether to exclude something should be much faster than the several function a filter solution requires.

Some improvements could be made by doing more work at compile-time to reduce some of the layers of functions, but I am not sure that it would be significant.  If you see something that suggests otherwise, I would be glad to look into it further.

All that said, I would be okay just including this and bumping a major version.

colinrymer:
I'm definitely in favor of merging this and bumping the major version. I'm happy to include the version bump in a commit in this PR if you'd like.

mitchellhenke:
Sorry for the delay on this, thanks for the time and effort 🙂

PR #111, opened 2017-02-01
fix documentation around exception filtering

Ebtoulson:
Documentation and `Sentry.EventFilter` callback didn't seem to match what was actually being used.

Failing test `test/event_test.exs:13`. The line number check doesn't match for me.


mitchellhenke:
oh gosh, thank you so much for fixing this 💖

PR #116, opened 2017-02-07
Handles String ip address

### LINKED ISSUE #115: protocol String.Chars not implemented for {:error, :einval}
As per doc, we put these inside our Phoenix router:

```
use Plug.ErrorHandler
use Sentry.Plug
```

However, this overrides Phoenix's error handler and effectively making it impossible to trace the actual error, while spitting this error:

```
protocol String.Chars not implemented for {:error, :einval}
```

Any ideas?

syamilmj:
Fixes #115

### DIFF
--- a/lib/sentry/plug.ex
+++ b/lib/sentry/plug.ex
@@ -97,9 +97,10 @@ defmodule Sentry.Plug do
     header_scrubber = Keyword.get(opts, :header_scrubber)
     request_id = Keyword.get(opts, :request_id_header) || @default_plug_request_id_header
 
-    conn = conn
-            |> Plug.Conn.fetch_cookies
-            |> Plug.Conn.fetch_query_params
+    conn =
+      conn
+      |> Plug.Conn.fetch_cookies
+      |> Plug.Conn.fetch_query_params
 
     %{
       url: "#{conn.scheme}://#{conn.host}:#{conn.port}#{conn.request_path}",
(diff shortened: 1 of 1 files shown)


mitchellhenke:
after some investigation, it looks like `remote_ip` should always be in the form of a tuple, based on the typespecs [here](http://erlang.org/doc/man/inet.html#type-ip_address) and [here](https://hexdocs.pm/plug/Plug.Conn.html#t:t/0).  What is the value in `remote_ip` that you're seeing that causes this to break?

syamilmj:
On my local, it returns `"127.0.0.1"` thus causing `:inet.ntoa/1` to fail with `{error, einval}`

Elixir: 1.4.0
Phoenix: 1.2.1
Plug: 1.30

Sentry integrated in web.ex:

```
def router do
    quote do
      use Phoenix.Router
      use Plug.ErrorHandler
      use Sentry.Plug
    end
end
```

jeregrine:
What about in your Endpoint? Also can you show us the entire Plug.Conn?
Because remote_ip 'should not' return that. The "host" can return that.

mitchellhenke:
Do you potentially have a Plug somewhere that is modifying `remote_ip`?

syamilmj:
Thanks for the replies. I'll investigate the issue as this seems to be unique for my setup. Closing this PR for now. Thanks!

Issue #90, opened 2016-11-17
ErrorLogger will not report all exceptions


DavidAntaramian:
Discussed this with @mitchellhenke last night. I'll detail the issue here and let you decide what, if anything, needs to be done.

Currently, the Sentry's [`:error_logger` handler](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/logger.ex), which captures exceptions reported by `:error_logger`, is only set to capture errors reported with the `:error_report` tag, which reports a process crash. It will _not_ capture exceptions reported with the `:error` tag which is sent as an exception report.

Consider the following two problems (note, the error here is that we are trying to enumerate over a binary):

```elixir
Task.async(fn -> Enum.each("hello", &IO.puts/1) end)
```

This causes the process running the task to crash and sends an `:error_report` via `:error_logger`. Therefore, this error will be seen by Sentry and reported.

```elixir
Process.spawn(fn -> Enum.each("hello", &IO.puts/1) end, [])
```

This causes the process running the task to report an exception and then exit. It reports an error as `:error` via `:error_logger`, therefore this error will not be seen by Sentry.

An important thing to note is that `:error` reports do not contain process information (_i.e._, the process dictionary is not included in the report, therefore the report will not contain any Sentry contextual data).

DavidAntaramian:
Just to follow up, `:error_report` seems to be sent whenever an OTP-compliant / supervised process exits abnormally.

I'm still trying to figure out under what circumstances `:error` is sent.

To give more context, I was running an app that included Phoenix and for various reasons was not using `Plug.ErrorLogger`. I was expecting that errors that occurred would be reported to Sentry regardless via the `:error_logger` system. They weren't, even though everything seems to suggest they should (Plug terminates the request using `exit/1` which is reported via `:error_report` as far as I can tell). There are a bunch of rabbit holes that this could probably be followed down, but I'd be interested to know if anyone else is having issues with exceptions not being reported.

DavidAntaramian:
After some digging, I figured out that the `:error` from Phoenix is actually being triggered by Ranch: https://github.com/ninenines/ranch/blob/master/src/ranch_conns_sup.erl#L316. 

Essentially, if you don't have the `Plug.ErrorLogger` being used (which I don't because it screws up CORS when there's an error), any error will cause a termination that is reported by Ranch using `:error_logger.error_msg/2`. This in turn causes an `:error` notice to be sent to the `:error_logger` app.

[...]

PR #120, opened 2017-02-14
Bring Sentry.Plug in line with Plug.ErrorLogger expectations

DavidAntaramian:
This goes back to #90. I finally got to circle back to it in our application, and I had a bit of a V8 moment when I realized what was going on.

As background: I was trying to figure out why our application kept disposing of the CORS headers when we encountered an error. I _thought_ that the issue was our utilization of `Plug.ErrorLogger` (and ancillary to that `Sentry.Plug` for reporting), since the definition of `Plug.ErrorLogger` captures the original `conn` and passes that to the error handler function. After reading elixir-lang/plug#486, I _thought_ that perhaps the answer was to utilize multiple plug pipeline as Jose suggested. But this last piece from Jose was confusing:

> That's how Phoenix solves the issue

So if Phoenix already implements the multiple plug pipeline strategy, I _shouldn't_ have to.

Queue more digging last week when the CORS errors brought our front-end team to a grinding halt, I did some digging and realized that Plug/Cowboy was having a fit because there was never a response sent when an exception occurred. And thus my V8 moment when I realized that adding `use Sentry.Plug` defines a function that doesn't send a response. I deleted the `use` call, and replaced it with the following:

```elixir
  def handle_errors(conn, opts = %{kind: kind, reason: reason, stack: stack}) do
    body_scrubber = {Sentry.Plug, :default_body_scrubber}
    header_scrubber = {Sentry.Plug, :default_header_scrubber}
    request_id_header = nil

    opts = [
      body_scrubber: body_scrubber,
      header_scrubber: header_scrubber,
      request_id_header: request_id_header
    ]

    request = Sentry.Plug.build_request_interface_data(conn, opts)
    exception = Exception.normalize(kind, reason, stack)
    Sentry.capture_exception(exception, [stacktrace: stack, request: request, event_source: :plug])
    super(conn, opts)
  end
```

This is essentially the code the `Sentry.Plug` macro will generate, with two important distinctions:

1. It defines `handle_error/2` as a public function, which is the same visibility as [the function it overrides](https://github.com/elixir-lang/plug/blob/v1.3.0/lib/plug/error_handler.ex#L49).
2. It calls `super/2` to fall back to the function it overrides, and that function _sends a response_.

I'm not sure that the public visibility is a big issue, but sending the response is. I've opened this PR to address both.

But I am surprised: am I the only person encountering this issue?

### DIFF
--- a/lib/sentry/plug.ex
+++ b/lib/sentry/plug.ex
@@ -82,12 +82,13 @@ defmodule Sentry.Plug do
     request_id_header = Keyword.get(env, :request_id_header, nil)
 
     quote do
-      defp handle_errors(conn, %{kind: kind, reason: reason, stack: stack}) do
+      def handle_errors(conn, opts = %{kind: kind, reason: reason, stack: stack}) do
         opts = [body_scrubber: unquote(body_scrubber), header_scrubber: unquote(header_scrubber),
                  request_id_header: unquote(request_id_header)]
         request = Sentry.Plug.build_request_interface_data(conn, opts)
         exception = Exception.normalize(kind, reason, stack)
         Sentry.capture_exception(exception, [stacktrace: stack, request: request, event_source: :plug])
+        super(conn, opts)
       end
     end
   end


jeregrine:
Huh. Good catch David. That is pretty wild, that nobody has seen this yet but you.

DavidAntaramian:
Maybe I'm just missing something?

mitchellhenke:
Thanks for this David ❤️ ❤️ 

Is yours a Phoenix app?  I think Phoenix relies on a response not being sent yet for use in [Phoenix.Endpoint.RenderErrors](https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/render_errors.ex), and this may cause issues with that?

An application currently using `Sentry.Plug` without Phoenix's error handling would seem to not send a response at all though if an error occurs.

DavidAntaramian:
Yes, we're running Phoenix 1.2.1

DavidAntaramian:
@mitchellhenke Just read through the Phoenix error handling you linked to. It seems that's what we encountered _after_ Sentry failed to send a response. Sentry's `handle_errors/2` is returning the result of `Sentry.capture_exception/2` which specs to ` {:ok, String.t} |  :error |  :excluded`; by my reading, `handle_errors/2` should have a return spec of `Plug.Conn.t`.

In otherwords, perhaps instead of calling `super/2`, it would be more advantageous to simply return `conn`? That way previously defined/higher level error handlers such as the Phoenix module you linked to can be relied upon to actually send the error.

mitchellhenke:
I did a quick test, and this branch does interrupt Phoenix error rendering.  Returning the response from [Plug](https://github.com/elixir-lang/plug/blob/v1.3.0/lib/plug/error_handler.ex#L50) instead of the one from [Phoenix](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/new/web/views/error_view.ex#L9).  I'm not sure what then is causing your app to not respond though.

I do think it should return `conn`, though we do maybe have to think about whether to return a response for non-Phoenix apps.

DavidAntaramian:
Assuming you have a Phoenix application using the master branch of this library, if you add headers to the  `conn` in `MyApp.Endpoint` before calling `plug MyApp.Router`, does the response returned to the client have the modified headers? My experience is that it doesn't.

DavidAntaramian:
@mitchellhenke Put together a small example app to reproduce. The key difference is if you comment out the `use Sentry.Plug` line here: https://github.com/DavidAntaramian/sentest/blob/master/web/router.ex#L4

On `GET http://localhost:4000/` you can see the difference in headers between when the Sentry plug is active and when it's not:

Sentry Example

The top example in the image is with the Sentry plug commented out. The second one is with the Sentry plug in use. You can see that the CORS headers aren't returned when Sentry is used.

mitchellhenke:
@DavidAntaramian have there been any developments on this?

DavidAntaramian:
@mitchellhenke Not that I know of. As we've discussed, this seems to have some issues with higher level integrations. Since this works for us, we'll be keeping it in our codebase, but I don't know what to do for other users.

PR #123, opened 2017-02-15
Add X-Sentry-Event-Id on error

hauleth:
This would allow to use Sentry's Users Feedback feature in API only
Elixir applications. Rest of changes will also allow to catch event
content in API after it was sent, so for example we could display
error ID on staging for QA.


lib/sentry/plug.ex
```diff
@@ -82,16 +82,26 @@ defmodule Sentry.Plug do
...
-                 request_id_header: unquote(request_id_header)]
-        request = Sentry.Plug.build_request_interface_data(conn, opts)
-        exception = Exception.normalize(kind, reason, stack)
-        Sentry.capture_exception(exception, [stacktrace: stack, request: request, event_source: :plug])
+      defp handle_errors(conn, error) do
+        opts = [body_scrubber: unquote(body_scrubber),
+                header_scrubber: unquote(header_scrubber),
+                request_id_header: unquote(request_id_header)]
+
+        {_, entry} = Sentry.Plug.capture(conn, error, opts)
```
hauleth: I exposed it as it would be nice to have possibility to write `handle_errors` by myself without digging too much into how `Sentry.Plug` works.

lib/sentry/client.ex
```diff
@@ -29,6 +29,8 @@ defmodule Sentry.Client do
     Task.Supervisor.async_nolink(Sentry.TaskSupervisor, fn ->
       try_request(:post, endpoint, auth_headers, body)
     end)
+
+    {:ok, event}
```
hauleth: Already done as I needed it to fix rest of app after changes ;)

lib/sentry/plug.ex
```diff
@@ -82,16 +82,26 @@ defmodule Sentry.Plug do
...
+
+        conn
+        |> Plug.assign(:sentry_entry, entry)
```
hauleth: I will try to write some tests that would cover this case.

hauleth:
The problem is that everything is connected. It is kind of holistic PR that need to change dialyzer specs (this also helped me with finding some bugs in app that was introduced by my change). So with exception to some whitespaces it would be really hard to separate all changes.

mitchellhenke:
I understand, but I would prefer that the primary feature being added be in good shape with tests covering it before other changes are added and reviewed.

The `Plug.Conn` being modified in `Sentry.Plug` is not used to send a response.  Unless a response is sent within `Sentry.Plug` after the header is added (which would break things like Phoenix's error handling views), I don't think there is a way for this to work.  Because of that, I think tests need to be added before other changes are made so we can verify whether it's possible to accomplish this feature 🙂

mitchellhenke:
@hauleth I appreciate the effort for this PR, but will unfortunately be closing it if we cannot find a solution to the above problems 😕

hauleth:
@mitchellhenke I will split it up into separate PRs and I will try to find a way to provide such feature.

PR #124, opened 2017-02-16
Default configs

aptinio:
**Set default configs in `Sentry.Mixfile.application[:env]`**

From the `config/config.exs` generated by `mix new`:

> This configuration is loaded before any dependency and is restricted
> to this project. If another project depends on this project, this
> file won't be loaded nor affect the parent project. For this reason,
> if you want to provide default values for your application for
> 3rd-party users, it should be done in your "mix.exs" file.

From `mix help compile.app`:

> `:env` - default values for the application environment. The application
> environment is one of the most common ways to configure applications.

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -93,6 +93,7 @@ defmodule Sentry do
 
   @client Application.get_env(:sentry, :client, Sentry.Client)
   @use_error_logger Application.get_env(:sentry, :use_error_logger, false)
+  @default_environment_name Mix.env
 
   def start(_type, _opts) do
     children = [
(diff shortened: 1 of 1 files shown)


aptinio:
You're welcome! 😃

The problem I see with setting the defaults using `Application.get_env/3` is that `Mix.env` is not available at runtime in an Erlang release. A couple of workarounds:

* Add `:mix` in the `:applications` list of `mix.exs`, although I've seen others advise against this, citing `Mix` is intended to set application env at compile time, and to get config from application env at runtime.
* Use `:dev` instead of `Mix.env` as the default `:environment_name`. In this case, the users will be required to configure `:environment_name` for their other environments.

So if you can reconsider, I think using Mix's `application[:env]` is still the best way to go to be able to use `Mix.env` as the default `:environment_name`.

mitchellhenke:
Could it be saved at compile time in a module attribute?

aptinio:
It could! Good idea! 💡

mitchellhenke:
Thank you for this 🙂

PR #132, opened 2017-03-02
Fix Credo warnings

hauleth:


hauleth:
It is quite big change, but I have no idea how to split it into separate PRs. Should cleanup codebase a little.

hauleth:
It would be worth try to integrate https://github.com/integration/ebert into this project.

mitchellhenke:
Thank you for this 🙂

lib/sentry/plug.ex
```diff
@@ -136,14 +144,20 @@ defmodule Sentry.Plug do
...
   def default_header_scrubber(conn) do
-    Enum.into(conn.req_headers, %{})
+    conn.req_headers
+    |> Enum.into(%{})
     |> Map.drop(@default_scrubbed_header_keys)
   end
 
+  @spec default_body_scrubber(Plug.Conn.t) :: map()
   def default_body_scrubber(conn) do
+    keys = Map.keys(conn.query_params)
```
mitchellhenke: what was the purpose of adding this?

hauleth: In next line we reject all keys that are in `conn.query_params` which mean that we ignore query params in body. Otherwise some tests fails. We cannot use `Enum.concat([conn.body_params, conn.path_params])` as first isn't populated through `Plug.Test.comm/3`. In future we could check a way to use `Plug.Parsers` together with `Plug.Conn.read_body/2` to skip body params on too big requests (or streaming queries).

PR #135, opened 2017-03-07
Bump hackney

Mazyod:
I was getting warnings from mix/hex that hackney 1.6.6 is retired. Looking at my lock file, it seems sentry is holding it back from updating to 1.7+.

Of course, I could lock it back to 1.6.5, but figured fixing it for everyone would be more beneficial.

Thanks,
Maz


mitchellhenke:
Thank you!

Issue #117, opened 2017-02-07
Better error handling for throws, exits, errors


mitchellhenke:
For reference:

```
[12:51:54]  <fishcakez>	Logger is the lib like error_logger for elixir isnt it?
[12:54:38]  <fishcakez>	jeregrine: the first step is that not everything is an exception
[12:54:50]  <fishcakez>	or rather an Exception.t
[12:59:05]  <jeregrine>	yea I suppose we mostly did this to avoid https://github.com/elixir-addicts/rollbax/blob/master/lib/rollbax/item.ex#L99
[12:59:41]  <jeregrine>	(ignore line number)
[12:59:49]  <jeregrine>	but like going through the message and formatting it
[13:02:54]  <jeregrine>	at the time, almost everyone used error_logger and just went with the flow but clearly we need to take a second look
[13:04:39]  <jeregrine>	the original sentry kind of did this, but it wasn't tested/was kind of a mess
[13:07:46]  <fishcakez>	jeregrine: maybe thus: https://gist.github.com/fishcakez/fb10fdafe6c79f37ca5db4cc98bdd663
[13:07:54]  <fishcakez>	but i wrote it directly in the gist so may have errors
[13:16:13]  <fishcakez>	jeregrine: updated as missed an :exit clause
[13:16:21]  <jeregrine>	and thats an error_logger handler?
[13:16:40]  <fishcakez>	no thats just a function to split the information you seem to want
[13:17:12]  <fishcakez>	there might be a better way but it was the best i could do in a couple of minutes
[13:17:51]  <fishcakez>	but it is quite improved on what you currently have
[13:25:33]  <fishcakez>	jeregrine: https://github.com/getsentry/sentry-elixir/issues/101 <- it would fix this issue for example
[13:25:50]  <jeregrine>	not sure I follow
[13:26:02]  <fishcakez>	the code above
[13:26:15]  <fishcakez>	because sentry-elixir;s error formatting is wrong it shows something very misleading there
[13:27:36]  <jeregrine>	fishcakez you mean to replace https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/event.ex#L107-L121
[13:27:46]  <jeregrine>	or the error_logger handler
[13:28:20]  <fishcakez>	i am confused
[13:28:50]  <fishcakez>	its just a function to split the error info of a crash report
[13:28:58]  <fishcakez>	i am confused
[13:29:00]  <fishcakez>	its just a function to split the error info of a crash report
[...]

Issue #101, opened 2016-12-09
Genserver/Task.Supervisor crash error/stacktrace is too vague in sentry


sikanhe:
I had a bug in the code that caused genservers being crashed, and the only print out I got in the sentry app was 

```
Elixir.ArgumentError: argument error
  File "gen_server.erl", line 812, in :gen_server.terminate/7
  File "proc_lib.erl", line 247, in :proc_lib.init_p_do_apply/3
``` 

I had to ssh into to the server to get the full error stack trace and crash report: 

```** (ArgumentError) argument error
    (kernel) gen_tcp.erl:149: :gen_tcp.connect/4
    (hackney) src/hackney_connect.erl:246: :hackney_connect.do_connect/5
    (hackney) src/hackney_connect.erl:37: :hackney_connect.connect/5
    (hackney) src/hackney.erl:328: :hackney.request/5
    (quesly) lib/quesly/utils/file.ex:20: App.Utils.File.download/3
    (quesly) web/models/audio_file.ex:51: App.AudioFile.import_from_url/1
    (quesly) lib/quesly/harbor/series_resync.ex:52: anonymous fn/2 in App.Harbor.Resync.resync/1
    (elixir) lib/enum.ex:651: Enum."-each/2-lists^foreach/1-0-"/2
```

mitchellhenke:
Thank you for creating this issue, this is something I'd like to improve.  Due to the way processes are, sometimes errors reported are less than helpful.

Could you gist a simple reproduction?

mitchellhenke:
I'm not positive, but it appears to partially be related to an error occurring when a GenServer tries to call the [`terminate/2`](http://elixir-lang.org/docs/v1.3/elixir/GenServer.html#c:terminate/2) callback.  Is `terminate/2` defined for the GenServer?

sikanhe:
@mitchellhenke no terminate/2 is not defined, i will try to recreate a gist this weekend

walter:
@mitchellhenke can you talk more about the error with GenServer, please? How does it present itself? We've started experiencing some mysterious timeouts from a package's GenServer that calls `Task.await` not long after switching to Sentry. Shot in the dark that it is related, but I thought I would check.

mitchellhenke:
@walter sorry, my sentence above may have been unclear.  To be more specific, I meant things like context and stacktrace will get lost when communicating over multiple processes.

The library executes the HTTP calls asynchronously, so it shouldn't hold up normal execution.

[...]

PR #136, opened 2017-03-07
Better reports from logger

mitchellhenke:
This should address #101, #117, and make it somewhat easier to make improvements going forward.  It should also help disambiguate errors from Processes.

@jeregrine

### DIFF
--- a/lib/sentry/event.ex
+++ b/lib/sentry/event.ex
@@ -66,6 +66,7 @@ defmodule Sentry.Event do
     breadcrumbs = Keyword.get(opts, :breadcrumbs, []) ++ breadcrumbs_context
 
     level = Keyword.get(opts, :level, "error")
+    culprit = Keyword.get(opts, :culprit) || culprit_from_stacktrace(stacktrace)
 
     release = Application.get_env(:sentry, :release)
 
@@ -74,7 +75,7 @@ defmodule Sentry.Event do
     env = Application.get_env(:sentry, :environment_name)
 
     %Event{
-      culprit: culprit_from_stacktrace(stacktrace),
+      culprit: culprit,
       message: message,
       level: level,
       platform: "elixir",
--- a/lib/sentry/event.ex
+++ b/lib/sentry/event.ex
@@ -108,14 +108,28 @@ defmodule Sentry.Event do
   """
   @spec transform_exception(Exception.t, keyword()) :: Event.t
   def transform_exception(exception, opts) do
+    error_type = Keyword.get(opts, :error_type) || :error
     normalized = Exception.normalize(:error, exception)
 
+    type = case error_type do
+      :error ->
+        normalized.__struct__
+      _ ->
+        error_type
+    end
+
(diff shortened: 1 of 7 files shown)


lib/sentry/event.ex
```diff
@@ -107,15 +107,30 @@ defmodule Sentry.Event do
   """
   @spec transform_exception(Exception.t, keyword()) :: Event.t
   def transform_exception(exception, opts) do
+    error_type = Keyword.get(opts, :error_type) || :error
     normalized = Exception.normalize(:error, exception)
 
+    type = case error_type do
```
jeregrine: could also do an if/else here

lib/sentry/event.ex
```diff
@@ -107,15 +107,30 @@ defmodule Sentry.Event do
...
+      :error ->
+        normalized.__struct__
+      _ ->
+        error_type
+    end
+
+    value = case error_type do
```
jeregrine: same

lib/sentry/event.ex
```diff
@@ -107,15 +107,30 @@ defmodule Sentry.Event do
...
+      _ ->
+        Exception.format_banner(error_type, exception)
+    end
+
+    module = Keyword.get(opts, :module)
+    exception = [%{type: type, value: value, module: module}]
     message = :error
               |> Exception.format_banner(normalized)
               |> String.trim("*")
               |> String.trim
```
jeregrine: parens()

lib/sentry/logger.ex
```diff
@@ -21,40 +21,45 @@ defmodule Sentry.Logger do
...
+
+  defp get_exception_and_stacktrace({kind, {exception, sub_stack}, _stack}) when is_list(sub_stack) do
+    {kind, exception, sub_stack}
+  end
+  defp get_exception_and_stacktrace({kind, exception, stacktrace}) do
+    {kind, exception, stacktrace}
+  end
+
+  defp get_initial_call_and_module({kind, exception, stacktrace}, error_info) do
+    case Keyword.get(error_info, :initial_call) do
```
jeregrine: whats going on here?

mitchellhenke: I need to add some documentation around this.  Sentry primarily groups by stacktrace, and error reports will include an `:initial_call` that's generally the MFA of how the Process was started, e.g. `{MyModule.GenServer, :init, [:arg]`.  For bad exits or throws, the stacktrace is the exact same regardless of which module it came from, so they ended up being grouped together.  Something like:

```elixir
[{:gen_server, :terminate, 7, [file: 'gen_server.erl', line: 812]},
 {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 247]}]
```

jeregrine: Ahh I see

mitchellhenke:
@jeregrine updated

jeregrine:
cc @fishcakez

PR #138, opened 2017-03-20
Source code context

mitchellhenke:
A feature to allow displaying the source code that caused the exception.

Example:

image


README.md
```diff
@@ -67,9 +67,14 @@ config :sentry,
...
 | `server_name` | False  | None | |
 | `use_error_logger` | False  | False | |
-| hackney_opts | False  | `[pool: :sentry_pool]` | |
-| hackney_pool_max_connections | False  | 50 | |
-| hackney_pool_timeout | False  | 5000 | |
+| `hackney_opts` | False  | `[pool: :sentry_pool]` | |
+| `hackney_pool_max_connections` | False  | 50 | |
+| `hackney_pool_timeout` | False  | 5000 | |
+| `enable_source_code_context` | True | | |
+| `root_source_code_path` | Required if `enable_source_code_context` is enabled | | Should generally be set to `File.cwd!`|
```
jeregrine: Maybe the default should be File.cwd!

README.md
```diff
@@ -67,9 +67,14 @@ config :sentry,
...
+| `context_lines` | False  | 3 | |
+| `exclude_patterns` | False  | `[~r"/_build/", ~r"/deps/", ~r"/priv/"]` | |
```
jeregrine: `source_code_exclude_patterns`

lib/sentry/sources.ex
```diff
@@ -0,0 +1,109 @@
...
+        root_source_code_path: File.cwd!,
+        context_lines: 5
+
+  ### Source code storage
+
+  The file contents are saved when Sentry is compiled, which can cause some
+  complications.  If a file is changed, and Sentry is not recompiled,
+  it will still report old source code.
+
+  The best way to ensure source code is up to date is to recompile Sentry
```
jeregrine: This probably needs to also be apart of main docs, "on deployment it is advised to recompile sentry to make sure you are always reporting the latest code."

README.md
mitchellhenke: good call.  to boot, this actually found a bug in that the option wasn't being used at all.

README.md
mitchellhenke: Discussed offline, but for posterity, the issue is that it will get called from within the Sentry source code, not the project, causing it to only read source code within Sentry.

lib/sentry/sources.ex
mitchellhenke: added

lib/sentry/sources.ex
```diff
@@ -0,0 +1,110 @@
...
+      Map.put(acc, key, value)
+    end)
+  end
+
+  @spec get_source_context(source_map, String.t, pos_integer()) :: {[String.t], String.t | nil, [String.t]}
+  def get_source_context(files, file_name, line_number) do
+    context_lines = Application.get_env(:sentry, :context_lines, @default_context_lines)
+    file = Map.get(files, file_name)
+    if file do
+      Enum.reduce(0..(2 * context_lines), {[], nil, []}, fn(i, {pre_context, context, post_context}) ->
```
jeregrine: I'd possibly refactor this to use more pattern matching this into a function something like

```
def get_context_lines(nil, line_number), do: {[], nil, []}
def get_context_lines(file, line_number) do 
  each_context_line(fn(i, {pre_context, context, post_context}) -> 
        context_line_number = line_number - context_lines + i
        source = Map.get(file, context_line_number)
        cond do
          context_line_number == line_number && source ->
            {pre_context, source, post_context}
          context_line_number < line_number && source ->
            {pre_context ++ [source], context, post_context}
          context_line_number > line_number && source ->
            {pre_context, context, post_context ++ [source]}
          true ->
            {pre_context, context, post_context}
        end
  end)
end

def each_context_line(fn) do
   context_lines = Application.get_env(:sentry, :context_lines, @default_context_lines)
   total_context_lines = 2 * context_lines
   Enum.reduce(0..total_context_lines, {[], nil, []}, fn)
end

```

jeregrine: Mostly to self document it as well as break it up a bit.

jeregrine: Could even break the cond into multiple function heads but thats prolly overkill.

mitchellhenke:
@jeregrine a relatively simple test is there, and I broke up the long function that gets the source code a bit

lib/sentry/sources.ex
```diff
@@ -0,0 +1,112 @@
...
+  def get_source_context(files, file_name, line_number) do
+    context_lines = Application.get_env(:sentry, :context_lines, @default_context_lines)
+    file = Map.get(files, file_name)
+
+    do_get_source_context(file, line_number, context_lines)
+  end
+
+  defp do_get_source_context(nil, _, _), do: {[], nil, []}
+  defp do_get_source_context(file, line_number, context_lines) do
+    Enum.reduce(0..(2 * context_lines), {[], nil, []}, fn(i, {pre_context, context, post_context}) ->
```
jeregrine: I don't think this makes it any clearer what is going on, the main issue IMO is that its not clear how you are accomplishing this. Breaking it into two function only add's 3 more lines :-/ 

Need to either add docs explaining the reduce, the 2* context lines, and the structure you are building and returning or break it into smaller functions with more descriptive names. 

Thats what I was trying to accomplish here https://github.com/getsentry/sentry-elixir/pull/138#discussion_r107304853, I wasn't clear before.

PR #141, opened 2017-03-23
Push in front of list instead of appending to end

### LINKED ISSUE #140: Optimize breadcrumb list creation
Currently this is done by appending to the end of the list [source](https://github.com/getsentry/sentry-elixir/blob/9077b52c5b25f257d53f58126d671343471b4142/lib/sentry/context.ex#L64) which is quite inefficient due to list construction. Instead we should append to the beginning and reverse list only when needed. This should provide some performance gain, allow simple removal of breadcrumbs and would make code more idiomatic overall.

hauleth:
Close #140


lib/sentry/context.ex
```diff
@@ -54,16 +54,15 @@ defmodule Sentry.Context do
   end
 
   defp get_context do
-    Process.get(@process_dictionary_key) || %{}
+    Process.get(@process_dictionary_key, %{})
```
mitchellhenke: These are unfortunately not the same, and should stay as is.

jeregrine:
```elixir
iex(1)> Process.get(:hello, [])    
[]
iex(2)> Process.get(:hello)    
nil
iex(3)> Process.get(:hello) || []
[]
iex(4)> Process.put(:hello)
** (UndefinedFunctionError) function Process.put/1 is undefined or private. Did you mean one of:

      * put/2

    (elixir) Process.put(:hello)
iex(4)> Process.put(:hello, nil)
nil
iex(5)> Process.get(:hello) || []
[]
iex(6)> Process.get(:hello, [])  
nil
iex(7)> 
```

hauleth: Reverted

mitchellhenke:
Thank you! 💖

PR #142, opened 2017-03-23
Allow modifying Event before it is sent

### LINKED ISSUE #137: Consider including logger metadata as tags in reported errors
Elixir's `Logger` supports the concept of `metadata`, which are arbitrary key-value pairs that, if configured, will be included in all logged messages in a standardized format.  In our application we have some standard things that define the context for every process (such as `user`) which we include in `Logger.metadata` -- that way, every logged message includes them in an easy-to-parse, standard format.

Unfortunately, since we don't also include those bits of data in our exception messages, we don't have that context when looking at errors reported to Sentry.  That makes it hard to troubleshoot errors since the context in `Logger.metadata` is so important to understanding the error (we need to know, for example, which user's data triggered the error).

It would be a huge help if `Logger.metadata` was included in errors submitted to sentry.  It seems like it would fit naturally in `tags`.  Is that something you would consider?

mitchellhenke:
closes #137 

cc @myronmarston

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ config :sentry,
 | `hackney_opts` | False  | `[pool: :sentry_pool]` | |
 | `hackney_pool_max_connections` | False  | 50 | |
 | `hackney_pool_timeout` | False  | 5000 | |
-| `pre_event_send_function` | False | | |
+| `before_send_event` | False | | |
 | `enable_source_code_context` | True | | |
 | `root_source_code_path` | Required if `enable_source_code_context` is enabled | | Should generally be set to `File.cwd!`|
 | `context_lines` | False  | 3 | |
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ config :sentry,
 | `hackney_opts` | False  | `[pool: :sentry_pool]` | |
 | `hackney_pool_max_connections` | False  | 50 | |
 | `hackney_pool_timeout` | False  | 5000 | |
+| `pre_event_send_function` | False | | |
 | `enable_source_code_context` | True | | |
 | `root_source_code_path` | Required if `enable_source_code_context` is enabled | | Should generally be set to `File.cwd!`|
 | `context_lines` | False  | 3 | |
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -27,6 +27,7 @@ defmodule Sentry.Client do
     {endpoint, public_key, secret_key} = get_dsn!()
 
     auth_headers = authorization_headers(public_key, secret_key)
+    event = maybe_call_pre_send_event_function(event)
     case Poison.encode(event) do
       {:ok, body} ->
         Task.Supervisor.async_nolink(Sentry.TaskSupervisor, fn ->
@@ -138,5 +139,18 @@ defmodule Sentry.Client do
     |> Kernel.*(1000)
     |> Kernel.round()
     |> :timer.sleep()
+  end
+
+  defp maybe_call_pre_send_event_function(event) do
+    case Application.get_env(:sentry, :pre_event_send_function) do
+      function when is_function(function) ->
+        function.(event)
+      {module, function} ->
+        apply(module, function, [event])
+      nil ->
+        event
+      _ ->
+        raise ArgumentError, message: ":pre_event_send_function must be an anonymous function or a {Module, Function} tuple"
+    end
   end
 end
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -160,7 +160,7 @@ defmodule Sentry.Client do
 
   defp maybe_call_before_send_event(event) do
     case Application.get_env(:sentry, :before_send_event) do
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -70,6 +70,7 @@ config :sentry,
 | `hackney_opts` | False  | `[pool: :sentry_pool]` | |
 | `hackney_pool_max_connections` | False  | 50 | |
 | `hackney_pool_timeout` | False  | 5000 | |
+| `pre_event_send_function` | False | | |
```
myronmarston: From our conversation in #137, it sounded like you agreed with my suggestion of `before_send_event` for this instead of this name.  Any reason you decided to use this name after all?

lib/sentry/client.ex
```diff
@@ -139,4 +157,17 @@ defmodule Sentry.Client do
     |> Kernel.round()
     |> :timer.sleep()
   end
+
+  defp maybe_call_pre_send_event_function(event) do
+    case Application.get_env(:sentry, :pre_event_send_function) do
+      function when is_function(function) ->
```
myronmarston: To ensure `function.(event)` doesn't get a `BadArityError`, you may want to use `is_function(function, 1)`.

test/client_test.exs
```diff
@@ -39,18 +39,74 @@ defmodule Sentry.ClientTest do
...
+      assert request_map["extra"] == %{"key" => "value"}
+      assert request_map["user"]["id"] == 1
+      Plug.Conn.resp(conn, 200, ~s<{"id": "340"}>)
+    end
+
+    Application.put_env(:sentry, :dsn, "http://public:secret@localhost:#{bypass.port}/1")
+    Logger.metadata([key: "value", user_id: 1])
+
+    Application.put_env(:sentry, :pre_event_send_function, fn(e) ->
+      metadata = Enum.into(Logger.metadata, %{})
```
myronmarston: This can be simplified to `Map.new(Logger.metadata)` (which is also faster than `Enum.into`).

test/client_test.exs
```diff
@@ -39,18 +39,74 @@ defmodule Sentry.ClientTest do
...
+    try do
+      Event.not_a_function
+    rescue
+      e ->
+        assert capture_log(fn ->
+          Sentry.capture_exception(e)
+        end)
+    end
+
+    Application.delete_env(:sentry, :pre_event_send_function)
```
myronmarston: This line won't get reached if any thing above fails, which means that if this test fails, it "leaks" its mutations to the app env.  It's better to use an `on_exit` to define cleanup that will be guaranteed to run.  On my projects, we've tended to wrap this up in a helper function:

``` elixir
    def modify_env(app, overrides) do
      original_env = Application.get_all_env(app)
      Enum.each(overrides, fn {key, value} -> Application.put_env(app, key, value) end)

      ExUnit.Callbacks.on_exit(fn ->
        Enum.each overrides, fn {key, _} ->
          if Keyword.has_key?(original_env, key) do
            Application.put_env(app, key, Keyword.fetch!(original_env, key))
          else
            Application.delete_env(app, key)
          end
        end
      end)
    end
```

Then any test can use `modify_env` and trust it will get cleaned up correctly.

test/support/test_client.ex
```diff
@@ -1,11 +1,25 @@
...
+      function when is_function(function) ->
+        function.(event)
+      {module, function} ->
+        apply(module, function, [event])
+      nil ->
+        event
+      _ ->
+        raise ArgumentError, message: ":pre_event_send_function must be an anonymous function or a {Module, Function} tuple"
+    end
+  end
```
myronmarston: This seems completely identical to the helper function in the `Client` module.  Seems like it might be worth pulling into a helper module?

test/support/test_pre_event_send_function.ex
```diff
@@ -0,0 +1,6 @@
+defmodule Sentry.PreEventSendFunctionTest do
+  def pre_event_send_function(event) do
+      metadata = Enum.into(Logger.metadata, %{})
```
myronmarston: Do you have a preference for 2 space indentation instead of 4 spaces?

mitchellhenke: nope, this was my bad

test/support/test_client.ex
mitchellhenke: Originally was from it being private, but it can be made public as people implementing their own Client behaviour could also make use of it.

test/client_test.exs
mitchellhenke: This has been an issue I wasn't sure how to handle, but I like this solution.  Thank you for sharing it.

README.md
mitchellhenke: mistake on my part, your name is still much better

mitchellhenke:
@myronmarston thank you so much for the feedback, I really appreciate it.  I've made a pass through to hopefully fix all of the issues.  I'll go through other tests modifying env to use the helper later.

PR #146, opened 2017-04-02
Fix invalid `@spec` on `Sentry` methods

hauleth:


lib/sentry.ex
```diff
@@ -97,40 +97,41 @@ defmodule Sentry do
...
       supervisor(Task.Supervisor, [[name: Sentry.TaskSupervisor]]),
       :hackney_pool.child_spec(Sentry.Client.hackney_pool_name(),  [timeout: @hackney_timeout, max_connections: @max_hackney_connections])
     ]
     opts = [strategy: :one_for_one, name: Sentry.Supervisor]
 
-
-    if @use_error_logger do
-      :error_logger.add_report_handler(Sentry.Logger)
-    end
+    if @use_error_logger, do: :error_logger.add_report_handler(Sentry.Logger)
```
jeregrine: We've discussed this before, but please don't make cosmetic changes in specific PRs.

hauleth:
@jeregrine I haven't found any place in files that you mentioned that describes return value of any method that I have fixed. What is more in most cases `@spec` was just invalid as returned values was different than specified (i.e. `Sentry.send_event/1` was returning `Task.t` while `@spec` stated that this will be either `{:ok, String.t}` or `:error`).

mitchellhenke:
Thank you! 💖 💖

PR #149, opened 2017-04-20
Always close the hackney client

waj:
When the request to the server returns a status code != 200 the client is ignored and the body is not read. This makes the client to stay in the connection pool forever. Once the max number of connections is reached (50 by default) no more requests can be sent to the server.


mitchellhenke:
Thanks!

PR #153, opened 2017-04-24
Bump hackney version

edevil:
Hackney 1.8.0 fixes a nasty bug: https://github.com/benoitc/hackney/issues/400

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -24,7 +24,7 @@ defmodule Sentry.Mixfile do
 
   defp deps do
     [
-      {:hackney, "~> 1.7.0"},
+      {:hackney, "~> 1.8.0"},
       {:uuid, "~> 1.0"},
       {:poison, "~> 1.5 or ~> 2.0 or ~> 3.0"},
       {:plug, "~> 1.0", optional: true},


mitchellhenke:
thanks!

myronmarston:
This makes it hard to upgrade my project that uses ExAws, because of its hackney requirements:

https://github.com/CargoSense/ex_aws/commit/93b48628e222af53efcaacc11c836a26db86ec86#commitcomment-21922987

Any chance you could relax this a bit to make it possible for me to upgrade?

mitchellhenke:
@myronmarston `4.0.1` only contained the hackney version change, so `4.0.0` should still work with hackney 1.7.x.  Is that sufficent?  Can relax the versions if it isn't.

myronmarston:
I saw that, and tried to get on hackney 1.7.1 (the only 1.7.x version supported by ExAws) and sentry 4.0.0 but ran into pinned version issues with other libraries.  We are currently on hackney 1.6.5 so if sentry doesn't need newer 1.7 or 1.8 features it would be nice to be able to use it with 1.6.5.

mitchellhenke:
@myronmarston added https://github.com/getsentry/sentry-elixir/commit/f21cb87c17b82396d7e5c602defdd856b968e1a8 and released in `4.0.2` 🙂

myronmarston:
❤️ ❤️ ❤️

PR #155, opened 2017-05-13
Extend Travis build matrix

samhamilton:
Hi, 

A small update to the OTP version and also to always test against the latest version of Elixir v1.4 so that you don't need to remember to add the specific patch release to Travis.

Thanks
Sam


mitchellhenke:
Awesome thank you!

PR #157, opened 2017-05-15
Avoid potential infinite loop when using a Sentry Logger backend

SimonWoolf:
if an exception occurs encoding a message

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -160,7 +160,7 @@ defmodule Sentry.Client do
   defp log_api_error(body) do
     Logger.warn(fn ->
       ["Failed to send Sentry event.", ?\n, body]
-    end)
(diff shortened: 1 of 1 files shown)


mitchellhenke:
Hey, thanks for opening this.  I don't think this is something I'll be accepting unfortunately as it's probably better to make use of a custom HTTP client.   It's not a well-documented feature, and I need to set up a behaviour for it to make it clear what the interface is, but the library allows you to define your own HTTP client and how it handles errors, retries, etc.  I've opened #158 to improve that area of it.

If you have any questions/issues around it, I'd be happy to help 🙂

mattheworiordan:
@SimonWoolf worth posting in the issue about the infinite loop we hit today in production, or not necessarily relevant to this PR?

PR #158, opened 2017-05-16
define and document http client behaviour

mitchellhenke:


jeregrine:
Whats the motivation for this change?

mitchellhenke:
@jeregrine it's a configurable option we have allowed for awhile, but was never documented.  It was prompted by #157

PR #160, opened 2017-05-23
Adding Fingerprint option to the event.

jeregrine:
To allow for custom grouping https://docs.sentry.io/learn/rollups/\#custom-grouping

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -81,7 +81,7 @@ defmodule Sentry do
   Simply calling `capture_exception/2` will send the event.
 
       Sentry.capture_exception(my_exception)
-      Sentry.capture_exception(other_exception, [source_name: :my_source])
+      Sentry.capture_exception(other_exception, [event_source: :my_source])
 
   ### Options
     * `:event_source` - The source passed as the first argument to `Sentry.EventFilter.exclude_exception?/2`
--- a/lib/sentry/event.ex
+++ b/lib/sentry/event.ex
@@ -21,7 +21,8 @@ defmodule Sentry.Event do
             request: %{},
             extra: %{},
             user: %{},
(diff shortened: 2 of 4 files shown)


lib/sentry.ex
```diff
@@ -81,7 +81,7 @@ defmodule Sentry do
   Simply calling `capture_exception/2` will send the event.
 
       Sentry.capture_exception(my_exception)
-      Sentry.capture_exception(other_exception, [source_name: :my_source])
+      Sentry.capture_exception(other_exception, [event_source: :my_source])
```
jeregrine: I also noticed this was mispelled I THINK so I just fixed it.

lib/sentry/event.ex
```diff
@@ -56,6 +58,8 @@ defmodule Sentry.Event do
 
     stacktrace = Keyword.get(opts, :stacktrace, [])
 
+    fingerprint = Keyword.get(opts, :fingerprint, ["{{ default }}"])
```
jeregrine: I'm not sure if we should ALWAYS send this up, or if sentry will ignore it if its `fingerprint: nil` or `fingerprint: []`

lib/sentry.ex
mitchellhenke: yep, good catch

lib/sentry/event.ex
mitchellhenke: Discussed in-person, but for posterity it seems other Sentry libraries do not send up a key/value for fingerprint if it's nil, so it would probably be best for us to do that too.

Currently we're directly encoding the Event struct to JSON, which is not ideal, and we'll need to not do that to fix the above issue.  Created an issue for it here: #161

jeregrine:
@mitchellhenke ping

PR #162, opened 2017-05-23
Fix the backoff to really be exponential

lewapkon:
Even though `send_event/1` docs told that the backoff is supposed to be
exponential it was quadratic


mitchellhenke:
haha oops, thanks for catching this! 🙂

PR #170, opened 2017-06-14
Fix arity stacktrace

### LINKED ISSUE #169: ** (exit) :undef consolidation
I'd like to understand if there's a better way to consolidate errors with the Sentry library. Early this morning we had an issue in our staging system where a function being called was undefined.

The Sentry error header looked like this:

```
exit MyApp.Storage.Cache.remove(5446, #<DateTime(2017-03-16T14:10:00Z Etc/UTC)>)
** (exit) :undef
```

The function was being called based on elements in a queue, resulting in over 1000 errors. Since Sentry identifies the error based on its header, and the parameters for each function call were different, each of these errors was treated as a unique Sentry issue. I would have preferred that Sentry be able to aggregate the error so all 1000 errors were treated as part of the same Sentry issue.

Is there a way to do this?

mitchellhenke:
closes #169

### DIFF
--- a/lib/sentry/event.ex
+++ b/lib/sentry/event.ex
@@ -150,6 +150,7 @@ defmodule Sentry.Event do
     stacktrace
     |> Enum.map(fn(line) ->
         {mod, function, arity, location} = line
+        arity = arity_to_integer(arity)
         file = Keyword.get(location, :file)
         file = if(file, do: String.Chars.to_string(file), else: file)
         line_number = Keyword.get(location, :line)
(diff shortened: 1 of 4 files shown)


lib/sentry/event.ex
```diff
@@ -180,4 +181,7 @@ defmodule Sentry.Event do
   @spec culprit_from_stacktrace(Exception.stacktrace) :: String.t | nil
   def culprit_from_stacktrace([]), do: nil
   def culprit_from_stacktrace([{m, f, a, _} | _]), do: Exception.format_mfa(m, f, a)
+
+  defp arity_to_integer(arity) when is_list(arity), do: Enum.count(arity)
+  defp arity_to_integer(arity), do: arity
```
jeregrine: Should you add a guard to check that its an integer?

mitchellhenke: I had that but decided against it since we weren't checking for it before and so I wasn't sure how safe it would be.  Erlang's typespecs say it should be an integer 0..255 or a list, so we could, but I erred on the side of caution.

I will switch it back though.

PR #175, opened 2017-06-27
post event hook

### LINKED ISSUE #166: Post-send Hook

mitchellhenke:
Closes #166

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -12,6 +12,9 @@ defmodule Sentry.Client do
     it is sent.  Accepts an anonymous function or a {module, function} tuple, and
     the event will be passed as the only argument.
 
+  * `:after_send_event` - callback that is called after an event is successfully sent.
+    Accepts an anonymous function or a {module, function} tuple, and the event will be passed as the only argument.
+
   Example configuration of putting Logger metadata in the extra context:
 
       config :sentry,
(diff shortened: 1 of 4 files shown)


lib/sentry/client.ex
```diff
@@ -157,6 +163,22 @@ defmodule Sentry.Client do
     {endpoint, public_key, secret_key}
   end
 
+  def maybe_call_after_send_event({:ok, _} = result, event) do
+    case Application.get_env(:sentry, :after_send_event) do
```
jeregrine: Can probably break this case into a function and use it for both scenarios instead of copying it.

lib/sentry/client.ex
```diff
@@ -157,6 +163,22 @@ defmodule Sentry.Client do
     {endpoint, public_key, secret_key}
   end
 
+  def maybe_call_after_send_event({:ok, _} = result, event) do
```
mitchellhenke: @jeregrine Right now we only try to call the hook if the sending event was successful, but I was thinking since we're passing the result, we could call `function.(result, event)` and let the user decide so we can skip the logic on our side?

lib/sentry/client.ex
mitchellhenke: they each call the configured function differently because of different arities, so this makes less sense now right?

PR #177, opened 2017-06-27
App frames

### LINKED ISSUE #165: Support In-App Frames
To distinguish which frames are in the application vs. libraries or Elixir core

mitchellhenke:
closes #165 

Gives you the option to only view stacktrace frames from your app.

[Module.split/1](https://hexdocs.pm/elixir/Module.html#split/1) unfortunately does not work for Erlang modules, as it's only supposed to be called on Elixir modules (https://github.com/elixir-lang/elixir/pull/5909).  As we sometimes have Erlang modules appear in stacktraces, we need to handle it ourselves 😕 

Example of viewing App Only when configuring `List` as one of the whitelisted modules:
image

image


lib/sentry/event.ex
```diff
@@ -191,4 +199,23 @@ defmodule Sentry.Event do
 
   defp arity_to_integer(arity) when is_list(arity), do: Enum.count(arity)
   defp arity_to_integer(arity) when is_integer(arity), do: arity
+
+  defp is_in_app?(nil, _in_app_whitelist), do: false
+  defp is_in_app?(_, []), do: false
```
jeregrine: Is this right? Shouldn't it default to true because we want to send everything?

mitchellhenke: This is the part that is difficult to explain/document for me.

It sends up the entire stacktrace regardless, but certain lines in it are marked as being part of the app or not.  If no modules are configured as being "part of the app", then we default to false.

jeregrine: I'm also wondering if instead of doing all that work we could rely on filepath like ruby.

jeregrine: Right but would it hide all of the lines? Should we default to nil in that case?

mitchellhenke: In the case that no lines are marked as "in_app", it just shows everything with no option to filter by "in_app".

File path is probably a no go, as it doesn't give us full path. Example:

```elixir
{Poison.Encoder.Any, :encode, 2, [file: 'lib/poison/encoder.ex', line: 383]}
{Poison, :encode!, 2, [file: 'lib/poison.ex', line: 41]}
```

mitchellhenke: @jeregrine ping

Issue #179, opened 2017-06-29
Config via System Environment


mitchellhenke:

mitchellhenke:
Closed in #180

PR #180, opened 2017-06-29
Configuration [WIP]

### LINKED ISSUE #178: Config via DSN Query String

mitchellhenke:
Centralized configuration, and will be adding something like https://github.com/getsentry/sentry-java/blob/master/sentry/src/main/java/io/sentry/config/Lookup.java#L86-L152 where we'll check env, then DSN, else use default value.

Will eventually close #178 and #179

### DIFF
--- a/lib/mix/tasks/sentry.send_test_event.ex
+++ b/lib/mix/tasks/sentry.send_test_event.ex
@@ -1,5 +1,6 @@
 defmodule Mix.Tasks.Sentry.SendTestEvent do
   use Mix.Task
+  alias Sentry.Config
 
   @shortdoc "Attempts to send a test event to check Sentry configuration"
   @moduledoc """
@@ -21,8 +22,8 @@ defmodule Mix.Tasks.Sentry.SendTestEvent do
     Mix.shell.info "public_key: #{public_key}"
     Mix.shell.info "secret_key: #{secret_key}"
     Mix.shell.info "included_environments: #{inspect included_environments()}"
-    Mix.shell.info "current environment_name: #{inspect environment_name()}"
-    Mix.shell.info "hackney_opts: #{inspect hackney_opts()}\n"
+    Mix.shell.info "current environment_name: #{inspect Config.environment_name()}"
+    Mix.shell.info "hackney_opts: #{inspect Config.hackney_opts()}\n"
   end
 
   defp included_environments do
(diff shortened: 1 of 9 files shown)


lib/sentry/config.ex
```diff
@@ -0,0 +1,89 @@
...
+  @default_hackney_timeout 5000
+  @default_exclude_patterns [~r"/_build/", ~r"/deps/", ~r"/priv/"]
+  @default_path_pattern "**/*.ex"
+  @default_context_lines 3
+  @default_sample_rate 1.0
+
+  def validate_config! do
+  end
+
+  def get_included_environments do
```
jeregrine: Wondering if we need the `get_` prefix.

lib/sentry/config.ex
```diff
@@ -0,0 +1,89 @@
+defmodule Sentry.Config do
```
jeregrine: I'm wondering if we load the config once at boot|compile and then return a struct with all the values set? Or do we need to hit the Application env ets table every time?

lib/sentry/config.ex
mitchellhenke: yeah, wasn't sure about that.  will drop

mitchellhenke: I can't really see the module doing any updating, etc., so it should be safe to assume we're only ever getting.

lib/sentry/config.ex
mitchellhenke: Have thought about it, but would prevent people updating config at runtime.  I'm not sure on the popularity of doing that.

lib/mix/tasks/sentry.send_test_event.ex
```diff
@@ -21,8 +22,8 @@ defmodule Mix.Tasks.Sentry.SendTestEvent do
     Mix.shell.info "public_key: #{public_key}"
     Mix.shell.info "secret_key: #{secret_key}"
     Mix.shell.info "included_environments: #{inspect included_environments()}"
-    Mix.shell.info "current environment_name: #{inspect environment_name()}"
-    Mix.shell.info "hackney_opts: #{inspect hackney_opts()}\n"
+    Mix.shell.info "current environment_name: #{inspect Config.environment_name()}"
```
jeregrine: Should this function be moved into Config? Config.print_info()

lib/sentry/config.ex
samphilipd: Please continue to allow us to set configuration at runtime :)

mitchellhenke: @samphilipd thanks for the feedback.  this should still be the case.  was there a regression that broke runtime configuration changes?

samphilipd: @mitchellhenke no, its working fine. I meant, please don't change the current, working behaviour :)

mitchellhenke: @samphilipd ah got it! thanks again 🙂

PR #182, opened 2017-07-05
Modules

### LINKED ISSUE #181: Support modules
https://docs.sentry.io/clientdev/attributes/

mitchellhenke:
Need to figure something out for handling the case that Travis is running into

Closes #181

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -97,6 +97,10 @@ Optional settings
 
   Expects a list of modules that is used to distinguish among stacktrace frames that belong to your app and ones that are part of libraries or core Elixir.  This is used to better display the significant part of stacktraces.  The logic is greedy, so if your app's root module is ``MyApp`` and your setting is ``[MyApp]``, that module as well as any submodules like ``MyApp.Submodule`` would be considered part of your app.  Defaults to ``[]``.
 
+.. describe:: report_deps
+
+  Will attempt to load Mix dependencies at runtime to report alongside events.  Defaults to `true`.
+
 .. describe:: context_lines
 
   The number of lines of source code before and after the line that caused the exception to be included.  Defaults to ``3``.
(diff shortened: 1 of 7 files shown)


test/event_test.exs
```diff
@@ -124,4 +124,16 @@ defmodule Sentry.EventTest do
         },
       ]} == event.stacktrace
   end
+
+  test "transforms mix deps to map of modules" do
+    modify_env(:sentry, modules: Mix.Dep.cached())
```
jeregrine: Ahh I don't know how I like this because it will fail whenever the deps change.

docs/config.rst
```diff
@@ -97,6 +97,10 @@ Optional settings
 
   Expects a list of modules that is used to distinguish among stacktrace frames that belong to your app and ones that are part of libraries or core Elixir.  This is used to better display the significant part of stacktraces.  The logic is greedy, so if your app's root module is ``MyApp`` and your setting is ``[MyApp]``, that module as well as any submodules like ``MyApp.Submodule`` would be considered part of your app.  Defaults to ``[]``.
 
+.. describe:: modules
+
+  Expects a map or a list of `%Mix.Dep{}` to report the loaded dependencies.  To include all Mix dependencies, it can be set to `Mix.Dep.cached()`.
```
jeregrine: We probably need to say that it expects the modules to be in the same format as Mix.Dep.cached. I would almost prefer for us to call Mix.Dep.cached() at runtime to this.

mitchellhenke: Unfortunately cannot call `Mix.Dep.cached` at runtime as a deployed application may be running without Mix

jeregrine: Could we do a macro instead? 

Also we could match on the type that Mix.Dep.cached returns and transform it, if it doesn't match just pass through?

Issue #183, opened 2017-07-14
mix deps.compile fails


KamilLelonek:
I cannot compile my code with `sentry`:

```
== Compilation error on file lib/sentry/event.ex ==
** (ArgumentError) application :sentry is not loaded, or the configuration parameter :enable_source_code_context is not set
    (elixir) lib/application.ex:261: Application.fetch_env!/2
    lib/sentry/event.ex:38: (module)
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6

could not compile dependency :sentry, "mix compile" failed. You can recompile this dependency with "mix deps.compile sentry", update it with "mix deps.update sentry" or clean it with "mix deps.clean sentry"
```

```elixir
  defp deps do
    [
      {:phoenix, "~> 1.3.0-rc"},
      {:phoenix_ecto, "~> 3.2"},
      {:cowboy, "~> 1.1"},
      {:postgrex, "~> 0.13"},
      {:ecto, "~> 2.1"},
      {:gettext, "~> 0.11"},
      {:sentry, "~> 5.0.0"},
      # ...
    ]
  end
```

```elixir
config :sentry,
  dsn:                        "...",
  enable_source_code_context: true,
  use_error_logger:           true,
  root_source_code_path:      File.cwd!,
  tags:                       %{env: "production"},
  environment_name:           Mix.env(),
  included_environments:      [:prod]
```

mitchellhenke:
@KamilLelonek thanks for opening this!  Unfortunately, I'm not able to reproduce the error with that configuration.  Did this configuration work in the past and break when you updated to `5.0.0`?

KamilLelonek:
I started with `5.0.0` and it broke immediately. I haven't tried any other versions.

KamilLelonek:
`4.0.3` fails as well:

```
== Compilation error on file lib/sentry/event.ex ==
** (ArgumentError) application :sentry is not loaded, or the configuration parameter :enable_source_code_context is not set
    (elixir) lib/application.ex:261: Application.fetch_env!/2
    lib/sentry/event.ex:29: (module)
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6

could not compile dependency :sentry, "mix compile" failed. You can recompile this dependency with "mix deps.compile sentry", update it with "mix deps.update sentry" or clean it with "mix deps.clean sentry"
```

KamilLelonek:
[...]

PR #186, opened 2017-07-19
Update README.md

KamilLelonek:
Add a note about Dockerfiles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,58 @@ current environment_name: :prod
 Sending test event!
 ```
 
+## Docker
+
+There are a couple of important things to keep in mind when building a `Docker` image.
+
+### config files
+
+If you follow [_Best practices for writing Dockerfiles_](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/), you probably copy just `mix.exs` and `mix.lock` files firstly to leverage caching. Then, you install and compile all dependencies.
+
+```dockerfile
+COPY mix.* ./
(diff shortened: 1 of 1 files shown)


mitchellhenke:
Thanks for opening this :)

I've opened #187 in relation to #183 to investigate if we can make configuration smoother, but I don't think this documentation in its current form adds enough to be included in the README.

Trying to compile without necessary configuration files is not an issue specific to Sentry, though documentation explaining some of the complexity around source code reading and compilation could be improved.

It could be prudent to add a warning [here](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/sources.ex#L3-L49) about the recursive file reading and a `Logger.info` call [here](https://github.com/getsentry/sentry-elixir/blob/08e14e4fb8db035980de3215d4633d88c2526359/lib/sentry/sources.ex#L53) that announces the recursive search is starting.  What do you think?

KamilLelonek:
It's completely up to you how you will proceed with this PR and the documentation.

You now have some setup that is relevant in terms of `Dockerfile` preparation.

There are already some people who started using `sentry` again with your library after they were informed about this solution.

mitchellhenke:
@KamilLelonek I'd be happy to link to a gist or blog post of yours that describes getting Docker and sentry-elixir set up, but maintaining docs for Docker builds within this repo is not something I am currently comfortable doing 🙂

KamilLelonek:
OK, I'm closing it then.

PR #191, opened 2017-08-01
adjust client to not send events when the DSN is set to an empty string

ryan-mahoney:
I'm not sure if this is the best way to do this... but this will inhibit the attempt to transmit error reports back to Sentry when the DSN is set to an empty string.

In our application, we would like to have the Sentry client included in both local, dev, staging and live environments. However, we don't want to transmit errors from the local environment. As we have set our DSN as an environment variable, the variable is not set in local environments and the read is set to an empty string.

Please let me know what you think... and if there is a much easier way to achieve this without changing the source ;)

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -58,15 +58,19 @@ defmodule Sentry.Client do
   """
   @spec send_event(Event.t) :: {:ok, Task.t | String.t} | :error | :unsampled
   def send_event(%Event{} = event, opts \\ []) do
-    result = Keyword.get(opts, :result, :async)
-    sample_rate = Keyword.get(opts, :sample_rate) || Config.sample_rate()
-
-    event = maybe_call_before_send_event(event)
-
-    if sample_event?(sample_rate) do
-      encode_and_send(event, result)
-    else
+    if Config.dsn() == "" do
       :unsampled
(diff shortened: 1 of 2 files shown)


mitchellhenke:
The best option is probably to use the `included_environments` and `environment_name`.

```
config :sentry,
  included_environments: [:dev, :staging, :prod],
  environment_name: :local
```

Would that work for your use case?

ryan-mahoney:
I had considered this. Unfortunately, our "local" and "dev" are both currently `:dev`

ryan-mahoney:
Even outside of our use case... I think it's a bit of a defect for the client library to throw an exception itself because it assumes the DSN would always be properly set in the config

mitchellhenke:
@ryan-mahoney The DSN has a specific use, and I'm not sure I want to expand it to also be an on/off switch for the library.  I am not familiar with your environments or deployment, but is there something preventing changing `:environment_name` or `:included_environments` on your local or dev environments?

I am open to changing error-handling on misconfigured DSNs.

ryan-mahoney:
I completely understand your objection. I'm going to close the PR. We'll probably just utilize the fork for the time being until we are able to better differentiate our local environment from our remote dev.

PR #197, opened 2017-08-07
Change `capture_exception` arity to 2 instead of 3

kamilogorek:
This function has only `1` and `2` arities – https://github.com/getsentry/sentry-elixir/blob/ec57fdb9f4ca0842512a1ef7991f1308968c9977/lib/sentry.ex#L117-L128

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ end
 
 ### Capture Exceptions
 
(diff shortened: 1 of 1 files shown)


mitchellhenke:
Thanks! 💖

PR #199, opened 2017-08-18
Allow Hackney 1.9 and later

halostatue:
`~> 1.8.0 or 1.6.5` is too strict; it should be `~> 1.8 or 1.6.5` to permit later compatible versions of Hackney.


mitchellhenke:
@halostatue thank you so much for the contribution! 💖

PR #202, opened 2017-09-04
fix incorrect default value included_environments

CatTail:


mitchellhenke:
While these evaluate to essentially the same, I think your way is much clearer!  Thank you 🙂

Issue #194, opened 2017-08-03
Avoid raising on malformed DSN


mitchellhenke:
Should just log error

From #191

PR #203, opened 2017-09-05
Ignore dsn if it's a tuple

brandonjoyce:
We have a config tool that uses tuples in mix config that get replaced at runtime.  The tuple makes the URI parse blow up during compile.  This would probably fix our issue.

Considering this WIP, but curious to get input from maintainers.  

**EDIT**
The reason this happens is due to to some attributes in the event module that use config values.

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -16,7 +16,8 @@ defmodule Sentry.Config do
   end
 
   def dsn do
-    get_config(:dsn, check_dsn: false)
+    dsn = get_config(:dsn, check_dsn: false)
+    if is_tuple(dsn), do: nil, else: dsn
   end
 
   def included_environments do
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -16,8 +16,8 @@ defmodule Sentry.Config do
   end
 
   def dsn do
-    dsn = get_config(:dsn, check_dsn: false)
-    if is_tuple(dsn), do: nil, else: dsn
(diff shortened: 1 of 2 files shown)


mitchellhenke:
Hi Brandon!

Thanks for opening this.  I'm curious about the root cause, and have a few questions.  Do you have an example of how you normally load config?  Is the issue at all related to #194?  I've started work on it, and so can prioritize it if it's related.

We do support loading via `{:system, "RUNTIME_SYSTEM_CONFIG_KEY"}` to allow reading from System at runtime, if that is helpful.

brandonjoyce:
Depending on how you approach solving #194, it might be a fix.

### How our config stuff works
We're doing Distillery releases for our production deployments, but we had to devise a way to make runtime config work in our current environment.  So, we wrote a package that uses something very similar to the `{:system, "KEY"}` convention that you mention.

Basically, our config will look like this:
```
config :sentry, dsn: {:enviro, "sentry_dsn"}
```

Our tool knows where to get config values and replaces these tuples appropriately.  Since that happens at runtime, we will have that tuple where the DSN should be.  It's worth mentioning that our build servers do not have access to those configuration values.  

### Alternative Possible Fix
We're kinda wondering if the DSN needs to be checked at all for this particular configuration value? `Config.source_code_path_pattern()`  If that could be set to `check_dsn: false` when calling get_config, then I think that might be another way to resolve our issue.

Apologies for not providing more context earlier.  I was in a rush.

PR #214, opened 2017-09-22
Fix Culprit ambiguity

### LINKED ISSUE #210: Sentry client truncates "culprit" field in JSON
Here is an example "culprit" send from our app:

`"CloseIO.HTTPClient.handle_response({:ok, %HTTPoison.Response{body: %{\"field-errors\" => %{\"phones\" => %{\"0\" => %{\"phone\" => \"Phone number is not valid. Please use the international format like +1650..."`

As a developer I would like to see the full, untruncated culprit message in the JSON to allow for easier debugging in case logs are not available, and to make it easier to find the error in the logs if they are available.

### LINKED ISSUE #213: should args act the role in culprit info to aggregate report ?
Hello,

When I using sentry dashboard, I met trouble about aggregating error report. In short, when I capture the exception (or message), different args will not aggregate together, just like:

```
> List.length 1
> List.length 2
> List.length 3
> List.length 4
```

these exception will not aggregate together but just like:

```
transaction List.length(1)
transaction List.length(2)
transaction List.length(3)
transaction List.length(4)
```

After scanned the source code, I noticed that

https://github.com/getsentry/sentry-elixir/blob/9cdfaac2465494aab31195d0705aa59fc7fbc866/lib/sentry/event.ex#L205

Culprit is made up of module, function and args. BUT, I need not to pay attention to the args here, and I will send the stacktrace information to the sentry dashboard.

So, any idea or advice ?

Thanks very much.

mitchellhenke:
Closes #210 and closes #213 

Culprit will now be formatted as `List.first/1` instead of `List.first(%{})`, which should help disambiguate errors.

To avoid losing the args, which can be helpful, they will now appear under `"vars"` for the stacktrace itself.

Results in something like this:

image


lib/sentry/event.ex
```diff
@@ -202,7 +204,24 @@ defmodule Sentry.Event do
 
   @spec culprit_from_stacktrace(Exception.stacktrace) :: String.t | nil
   def culprit_from_stacktrace([]), do: nil
-  def culprit_from_stacktrace([{m, f, a, _} | _]), do: Exception.format_mfa(m, f, a)
+  def culprit_from_stacktrace([{m, f, a, _} | _]) do
+    Exception.format_mfa(m, f, arity_to_integer(a))
```
redink: Cool, it will convert to `"m.f/n"` rather than `"m.f(a...l...l...a...r...g...s)"`. Thanks again.

mitchellhenke: @redink yep, and the `a...l...l...a...r...g...s` will still show up in the stacktrace on Sentry's website if they are present in the Elixir stacktrace 🙂

redink: of course, but it will not appear in `culprit` and will help me for aggregate similar reports.

PR #219, opened 2017-10-16
allow Sentry.Plug.handle_errors to be overridden

mpiercy827:
I have a use case where I want to use the `Sentry.Plug` module in my router, but also define some custom behavior in the `Plug.ErrorHandler.handle_errors/2` callback, such as sending metrics to our StatsD server, logging, etc.

Currently, the `handle_errors` method in the `Sentry.Plug` can't be overridden or extended, so I have to rescue exceptions and use `Sentry.capture_exception/2` in order to perform this extra behavior.

### DIFF
--- a/lib/sentry/plug.ex
+++ b/lib/sentry/plug.ex
@@ -113,6 +113,8 @@ defmodule Sentry.Plug do
         exception = Exception.normalize(kind, reason, stack)
         Sentry.capture_exception(exception, [stacktrace: stack, request: request, event_source: :plug])
       end
+
+      defoverridable [handle_errors: 2]
     end
   end
 


mitchellhenke:
hey @mpiercy827, thanks for opening this!

This is an area where I'm not 100% sure when it to allow further configuration, or if more complex use cases should be using their own Plug.  Currently, the `handle_errors/2` function only does call `Sentry.capture_exception/2`.

Has creating a separate Plug and working with Sentry in it been manageable?  Should we add documentation around to make it clearer, or do you think we should make the Plug more configurable (as done here) to avoid that if possible?

mpiercy827:
The things we do before/after sending the event to Sentry require the context of the Conn. This includes sending custom metrics, and sending the request body to Sentry as well.

Currently, we are just making a call to `Sentry.capture_exception/2` in our controller, because the `Sentry.Plug` doesn't send our POST body. We are interested in using the Plug because it sends a lot more information, although we still have to find a way to send the request params as well.

I think a custom Plug would work here, but I also think it would be a lot of duplicate code, since we only need a little bit of extra functionality on top of the Plug.

I suppose the ability to override the handle errors function of the Plug isn't essential, but it would be a nice-to-have.

mitchellhenke:
@mpiercy827 Apologies for the long comment.

In overriding`Sentry.Plug`'s `handle_errors/2`, wouldn't one have to duplicate anyway?  Once the function is overridden, the original `Sentry.Plug.handle_errors/2` won't be called, so you wouldn't be sending an error unless you copied over `Sentry.capture_exception`.  From my understanding, I think one should implement their own `error_handler`, and use Sentry functions where necessary.

`Sentry.Plug` should be sending the POST body in the default configuration.  If there is an issue with it, let me know 🙂 

For ways to handle more complex logging that can decouple the retrieval/storage and the sending of logs, I would suggest looking into `Logger.metadata`: [Logger docs](https://hexdocs.pm/logger/Logger.html#module-console-backend) / [good blog post by Timber](https://webcache.googleusercontent.com/search?q=cache:Ir1pnctyz7gJ:https://timber.io/blog/elixir-logger-and-the-power-of-metadata+&cd=4&hl=en&ct=clnk&gl=us)

Sentry also supports sending up metadata with errors using `Sentry.Context`, ex: `Sentry.Context.set_extra_context(%{my_metadata: "value"})` or `Sentry.capture_exception(exception, [extra: %{my_metadata: "value"}]`

mpiercy827:
When overriding `Sentry.Plug`'s `handle_errors/2` function, we actually don't have to duplicate any code if we call [`super`](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#super/1):

```elixir
  def handle_errors(conn, params) do
    send_logs(conn)
    send_metrics(conn)

    super(conn, params)

    send_response(conn)
  end
```

This way, we can just add some additional functionality on top 👍 

And as for the POST body, it doesn't seem to get sent with our errors. When we've tried it, the plugs in our router are ordered like this:

```elixir
  plug Plug.Router
  plug Plug.ErrorHandler
  plug Sentry.Plug
  ...
  plug :put_resp_content_type, "application/json"
  plug Plug.Parsers, parsers: [:urlencoded, :multipart, :json], json_decoder: Poison
  ...
  plug :match
  plug :dispatch
```

The issue is that when in the handle errors function, the `Plug.Conn` has unfetched `body_params`. If we were able to send along the POST body that would clean up our current code considerably!

mitchellhenke:
Ah, I see, I thought there was stuff you wanted to pass along while capturing the exception.

Are you able to get the application to compile calling:`plug Plug.ErrorHandler` and `plug Plug.Sentry`?  Neither implement the Plug Behaviour since they don't define the callbacks, and so including them in the pipeline should error.  They should be included

mitchellhenke:
Ah, I see, I thought there was stuff you wanted to pass along while capturing the exception.

Are you able to get the application to compile by calling:
```
plug Plug.Router
plug Plug.ErrorHandler
plug Plug.Sentry
```

Those do not implement the Plug Behaviour since they don't define the callbacks, and so including them in the pipeline as regular Plugs should error.  If you can share a relatively minimal test or application repo that reproduces the error with Plug bodies, I can take a look.

As far as making the function overridable, I'm open to it, but would rather see this can be solved outside of that first.

mitchellhenke:
The error could also be happening before the body parser Plug gets called, in which case the body will not have been read, and won't be available in the `handle_errors` callback, and Sentry won't be able to send it up.

mpiercy827:
Sorry for the delay. I also noticed that I made a mistake when listing the order of the plugs above:

```elixir
  use Plug.Router
  use Plug.ErrorHandler
  use Sentry.Plug
  ...
  plug :put_resp_content_type, "application/json"
  plug Plug.Parsers, parsers: [:urlencoded, :multipart, :json], json_decoder: Poison
  ...
  plug :match
  plug :dispatch
```

I wrote `plug Sentry.Plug` instead of `use Sentry.Plug`. Does this look like the correct usage? When an error occurs in one of our Plug routes, the Sentry Plug doesn't capture the body. I'll work on getting a simple test up.

mitchellhenke:
That does look correct, yep.  A minimal test would be great.  https://github.com/getsentry/sentry-elixir/blob/master/test/plug_test.exs has some examples of using a similar Plug stack (https://github.com/getsentry/sentry-elixir/blob/master/test/support/test_plug.ex).

If you run into any issues or have questions, please feel free to ask, I'm happy to help 🙂

mitchellhenke:
I'm going to close this PR for now, but I'm happy to re-open or assist if there's a way for me to help with the issue.

Issue #220, opened 2017-10-18
Can't Run Tests Locally


jalcine:
Running Elixir 1.5.2 on Erlang 20 gives me this off a master build:

```shell
✔ ~/src/sentry-elixir [supporting-confex L|⚑ 1]
02:06 $ mix test


  1) test parses error exception (Sentry.EventTest)
     test/event_test.exs:14
     Assertion with == failed
     code:  assert event.stacktrace() == %{frames: Enum.reverse([%{filename: nil, function: "Sentry.Event.not_a_function/3", lineno: nil, module: Sentry.Event, context_line: nil, post_context: [], pre_context: [], in_app: false, vars: %{"arg0" => "1", "arg1" => "2", "arg2" => "3"}}, %{filename: "test/event_test.exs", function: "Sentry.EventTest.event_generated_by_exception/1", lineno: 8, module: Sentry.EventTest, context_line: nil, post_context: [], pre_context: [], in_app: false, vars: %{}}, %{filename: "test/event_test.exs", function: "Sentry.EventTest.\"test parses error exception\"/1", lineno: 15, module: Sentry.EventTest, context_line: nil, post_context: [], pre_context: [], in_app: false, vars: %{}}, %{filename: "lib/ex_unit/runner.ex", function: "ExUnit.Runner.exec_test/1", lineno: 302, module: ExUnit.Runner, context_line: nil, post_context: [], pre_context: [], in_app: false, vars: %{}}, %{filename: "timer.erl", function: ":timer.tc/1", lineno: 166, module: :timer, context_line: nil, post_context: [], pre_context: [], in_app: false, vars: %{}}, %{filename: "lib/ex_unit/runner.ex", function: "anonymous fn/3 in ExUnit.Runner.spawn_test/3", lineno: 250, module: ExUnit.Runner, context_line: nil, post_context: [], pre_context: [], in_app: false, vars: %{}}])}
[...]

PR #221, opened 2017-10-18
Fix text support for Elixir 1.5 and Erlang/OTP 20.0.

jalcine:
> Relates to https://github.com/getsentry/sentry-elixir/issues/220.

The aim of this PR is to allow for support of Elixir 1.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,11 @@ language: elixir
 elixir:
   - 1.3.4
   - 1.4
+  - 1.5
 otp_release:
   - 18.3
   - 19.3
+  - 20.0
 
 matrix:
   allow_failures:
--- a/test/event_test.exs
+++ b/test/event_test.exs
@@ -11,6 +11,39 @@ defmodule Sentry.EventTest do
     end
   end
 
+  def get_stacktrace_frames_for_elixir() do
+    cond do
+      Version.match?(System.version, "< 1.4.0") ->
+        [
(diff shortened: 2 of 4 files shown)


jalcine:
So initially I just bumped the version use of Erlang and Elixir to see what dependencies resulted in failures. This helped out noticing these failures:

1. Erlang/OTP 20.0 fails.
2. Elixir 1.5 just _doesn't_ work with this.

mitchellhenke:
Can you be more specific with how the library fails or doesn't work with Erlang 20 and Elixir 1.5?

The library should be functioning on Elixir 1.5 and Erlang 20 despite surprising test failures.  The failures on Travis suggest this as well, as the failures are around a couple tests that have specifics related to Erlang/Elixir versions.

jalcine:
Apologies for lack of clarity.

The library _does_ work for 1.5 and OTP 20; it's the act of running tests with them.

jalcine:
This gets trickier as I go - some line numbers shift even in versions of Erlang.

Would it make sense to _drop_ line numbers as something to match against?

mitchellhenke:
@jalcine Given the extent of it, I think dropping line numbers for core Elixir/Erlang functions is fine for the match.  I would like to at least```assert is_integer(json["stacktrace"]["frames"]["lineno"])` in it's place though

jalcine:
That line shift was due to me adding a method - didn't notice that. No need to drop the line numbers anymore. I understand why it's done now more in depth; these values are definitive and shouldn't shift.

mitchellhenke:
I think the `cond` works well for situations like `NaiveDateTime.from_erl/3`, as that should be way less common, but line numbers are too much of a hassle.

jalcine:
Going to squash these commits once I get Elixir 1.3.4 + OTP < 20.0 working again.

mitchellhenke:
Thanks @jalcine! ❤️ 💖

PR #236, opened 2017-11-26
Update README.md

dustinfarris:
The application.ex example is confusing because of the Sentry namespace.  It looks like it should all be copy/pasted, verbatim.  Use "MyApp" instead to signify what is not required for Sentry.


mitchellhenke:
hi @dustinfarris! I absolutely agree, thanks for the contribution 🙂

PR #237, opened 2017-11-27
fix warning for GenEvent

redink:
GenEvent module is deprecated.

> If you have to implement an event handler to integrate with an existing system, such as Elixir’s Logger, please use :gen_event instead.

So, I replaced `GenEvent` with `:gen_event`, just like: https://github.com/elixir-lang/elixir/blob/v1.5.2/lib/logger/lib/logger/error_handler.ex


mitchellhenke:
Thank you for the contribution! ❤️ 💖 💛 💚

redink:
Could you make a new release please ? Thanks !

mitchellhenke:
Yep, I can tomorrow!

mitchellhenke:
@redink I've released `6.0.5`. If you have any issues, please let me know 🙂

redink:
Thanks very much.

PR #238, opened 2017-11-29
include error_type in Plug.Sentry

mitchellhenke:
We weren't including the kind of error when reporting errors from `Sentry.Plug` (`:error`, `:exit`, etc.) and this PR fixes that.

Issue originally mentioned in https://github.com/elixir-ecto/db_connection/issues/107#issuecomment-347787579

### DIFF
--- a/lib/sentry/plug.ex
+++ b/lib/sentry/plug.ex
@@ -111,7 +111,8 @@ defmodule Sentry.Plug do
                  request_id_header: unquote(request_id_header)]
         request = Sentry.Plug.build_request_interface_data(conn, opts)
         exception = Exception.normalize(kind, reason, stack)
-        Sentry.capture_exception(exception, [stacktrace: stack, request: request, event_source: :plug])
+        Sentry.capture_exception(exception, [stacktrace: stack, request: request,
+                                             event_source: :plug, error_type: kind])
       end
     end
   end


novaugust:
I'm still seeing these generic erlang errors from the original comment with Sentry 6.1.0. 
I'm guessing this is because I'm using Sentry.Logger in addition to Sentry.Plug. Does a similar update need to [happen for the Logger](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/logger.ex#L56)? Is it a mistake to be using both of these hooks at the same time?

mitchellhenke:
@novaugust thanks for the heads up! I am looking into this now.

mitchellhenke:
@novaugust I wasn't able to find anything in checking just now. You should be able to use both Sentry.Logger and Sentry.Plug together.  Sentry.Logger gets the error_type where possible (it isn't always provided).

Are you seeing the same issue with DBConnection errors or is the error from something else?

novaugust:
Here's the trace that should be coming in as DBConnectionErrors
screen shot 2018-03-21 at 8 54 30 am

novaugust:
I do have _one_ instance of the right error coming in
screen shot 2018-03-21 at 8 54 30 am

novaugust:
Here's every connection checkout failure. You'll see that only one is caught as DBConnectionError, the rest are erlang
screen shot 2018-03-21 at 9 01 37 am

PR #248, opened 2018-01-24
fix compile warning

mitchellhenke:
Current implementation causes this warning because we're checking a compile-time module attribute at runtime, so the result is always the same. Looks like this:

```
  warning: this check/guard will always yield the same result
    lib/sentry/event.ex:195
```

Not super familiar with metaprogramming stuff, but this defines a macro that will either find and insert the source code context or return the frame as-is and remove the warning.

### DIFF
--- a/lib/sentry/event.ex
+++ b/lib/sentry/event.ex
@@ -188,7 +188,9 @@ defmodule Sentry.Event do
         do_put_source_context(unquote(frame), unquote(file), unquote(line_number))
       end
     else
-      frame
+      quote do
+        unquote(frame)
+      end
     end
   end
 
@@ -218,7 +220,8 @@ defmodule Sentry.Event do
     |> Enum.reverse()
   end
 
-  defp do_put_source_context(frame, file, line_number) do
+  @spec do_put_source_context(map(), String.t(), integer()) :: map()
+  def do_put_source_context(frame, file, line_number) do
     {pre_context, context, post_context} =
       Sentry.Sources.get_source_context(@source_files, file, line_number)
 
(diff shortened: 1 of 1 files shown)


lib/sentry/event.ex
```diff
@@ -208,18 +218,14 @@ defmodule Sentry.Event do
     |> Enum.reverse()
   end
 
-  @spec put_source_context(map(), String.t(), integer()) :: map()
-  def put_source_context(frame, file, line_number) do
-    if @source_code_context_enabled do
-      {pre_context, context, post_context} =
-        Sentry.Sources.get_source_context(@source_files, file, line_number)
+  @spec macro_put_source_context(map(), String.t(), integer()) :: map()
+  def macro_put_source_context(frame, file, line_number) do
```
jeregrine: make this private and rename it to "do_put_source_context" instead of macro. Its pretty common.

lib/sentry/event.ex
```diff
@@ -208,18 +218,14 @@ defmodule Sentry.Event do
...
+    {pre_context, context, post_context} =
+      Sentry.Sources.get_source_context(@source_files, file, line_number)
 
-      Map.put(frame, :context_line, context)
-      |> Map.put(:pre_context, pre_context)
-      |> Map.put(:post_context, post_context)
-    else
-      frame
-    end
+    Map.put(frame, :context_line, context)
```
jeregrine: Start this with frame |>

mitchellhenke:
@jeregrine updated 🙂

lib/sentry/event.ex
```diff
@@ -182,6 +182,16 @@ defmodule Sentry.Event do
     end)
   end
 
+  defmacrop put_source_context(frame, file, line_number) do
+    if Config.enable_source_code_context() do
+      quote do
+        do_put_source_context(unquote(frame), unquote(file), unquote(line_number))
+      end
+    else
+      frame
```
jeregrine: Do you need to quote this?

mitchellhenke:
@jeregrine updated 🙂

PR #254, opened 2018-02-15
Scrub Cookies

### LINKED ISSUE #253: Is there a way to scrub cookies?
I store the auth token in a cookie, would be nice to scrub those

mitchellhenke:
Fixes a bug in that we should not be sending cookies, as there's no good default to prevent sending sensitive data.

By default, we won't send any, but will allow for users to override it with the `cookie_scrubber` option on `Sentry.Plug`

Closes #253

### DIFF
--- a/lib/sentry/plug.ex
+++ b/lib/sentry/plug.ex
@@ -1,7 +1,7 @@
 if Code.ensure_loaded?(Plug) do
   defmodule Sentry.Plug do
     @default_scrubbed_param_keys ["password", "passwd", "secret"]
-    @default_scrubbed_header_keys ["authorization", "authentication"]
+    @default_scrubbed_header_keys ["authorization", "authentication", "cookie"]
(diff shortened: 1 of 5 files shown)


lib/sentry/plug.ex
```diff
@@ -171,6 +180,13 @@ if Code.ensure_loaded?(Plug) do
       fun.(conn)
     end
 
+    @spec default_cookie_scrubber(Plug.Conn.t()) :: map()
+    def default_cookie_scrubber(_conn) do
+      # Enum.into(conn.req_headers, %{})
```
jeregrine: remove commented out code

test/plug_test.exs
```diff
@@ -169,4 +162,23 @@ defmodule Sentry.PlugTest do
       |> Sentry.ExampleApp.call([])
     end)
   end
+
+  test "scrubs cookies" do
+    bypass = Bypass.open()
```
jeregrine: add a test with a custom scrubber

PR #261, opened 2018-03-26
allow overriding handle_errors after using Sentry.Plug

azhi:
mention this in docs, as well as a fact that using Sentry.Plug does not retain Plug.ErrorHandler default behaviour of sending "Something went wrong" as a response (because Sentry.Plug does not call `super()` in it's handle_errors/2 implementation)


mitchellhenke:
Thanks for opening this, I’ll take a look soon!

mitchellhenke:
Thanks! 💖

PR #263, opened 2018-04-05
feat: Allow public DSNs

jan-auer:
The private DSNs (including secret keys) will be officially deprecated soon. This PR allows users to specify a public DSN and makes sure the `sentry_secret` is not included in the header anymore. For backwards compatibility, a private DSN can still be used, though. 

Also, this PR updates docs to use interactive DSN code snippets and fixes an invalid `try do ... rescue` statement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ An example production config might look like this:
 
 ```elixir
 config :sentry,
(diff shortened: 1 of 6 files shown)


mitchellhenke:
@jan-auer Thank you! 💖 ❤️ 💜 💚 💖

PR #273, opened 2018-06-17
Use `get_peer_data` function

tt:
In elixir-plug/plug@d24ba4f7c859b47866b7031a1ea68469f1808183, the `peer` field got replaced with a `get_peer_data` function.

This makes the package require Plug 1.6.0 and uses this new function to resolve the remote port.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -27,7 +27,7 @@ defmodule Sentry.Mixfile do
       {:hackney, "~> 1.8 or 1.6.5"},
       {:uuid, "~> 1.0"},
       {:poison, "~> 1.5 or ~> 2.0 or ~> 3.0"},
-      {:plug, "~> 1.0", optional: true},
+      {:plug, "~> 1.6", optional: true},
       {:phoenix, "~> 1.3", optional: true},
       {:dialyxir, "> 0.0.0", only: [:dev], runtime: false},
       {:ex_doc, "~> 0.18.0", only: :dev},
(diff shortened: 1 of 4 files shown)


mix.exs
```diff
@@ -27,7 +27,7 @@ defmodule Sentry.Mixfile do
       {:hackney, "~> 1.8 or 1.6.5"},
       {:uuid, "~> 1.0"},
       {:poison, "~> 1.5 or ~> 2.0 or ~> 3.0"},
-      {:plug, "~> 1.0", optional: true},
+      {:plug, ">= 1.6.0 and < 2.0.0", optional: true},
```
mitchellhenke: Could you change this to `~> 1.6`? I think that would be equivalent to the above?

mitchellhenke: All good, thanks for addressing git!

If it's helpful, https://hexdocs.pm/elixir/Version.html has an explanation of the version locking rules

mitchellhenke:
@tt Thanks again for opening this PR! 💖 💖 💖 💖

stevedomin:
@mitchellhenke I don't think the library should use the peer's port as the remote port. With today's architectures this is often going to be wrong. Might be best to skip it.

mitchellhenke:
@stevedomin This is an area I'm not overly familiar with, so if you don't mind I have a few questions. Is there something that would be more correct? Or it's not feasible to do that, and the current port information isn't super valuable, and should be removed?

Any links to information on the topic would also be greatly appreciated if that exists. 🙂

stevedomin:
@mitchellhenke sure.

If your application is behind a load-balancer/proxy the `X-Forwarded-Port` header will sometimes be set (AWS does this I believe). The `Forwarded` header can also contain port information of the nodes (but it's not guaranteed).

Despite this I still think it might be best to remove it completely. It's not going to be set often enough that it is useful and having the peer port is not super valuable.

PR #292, opened 2018-06-28
add note about testing against Sentry in a user application

### LINKED ISSUE #231: Add documentation around testing with Sentry
Users include Sentry as part of their tests, to ensure an HTTP call is made certain scenarios.  Our tests use Bypass, so even just directing to those would be helpful.

mitchellhenke:
closes #231

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,26 @@ You can then view the docs in your browser:
 $ open docs/_build/html/index.html
 ```
 
+## Testing with Sentry
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -268,6 +268,26 @@ You can then view the docs in your browser:
 $ open docs/_build/html/index.html
 ```
 
+## Testing with Sentry
+
+In some cases, users may want to test that certain actions in their application cause a report to be sent to Sentry.  Sentry itself does this by using [Bypass](https://github.com/PSPDFKit-labs/bypass).  It is important to note that when modifying the environment configuration the test case should not be run asynchronously.  Not returning the environment configuration to its original state could also affect other tests depending on how the Sentry configuration interacts with them.
```
jeregrine: Might also be worth mentioning mox https://github.com/plataformatec/mox or possibly just redoing our tests to use mox or similar ;-)

PR #293, opened 2018-06-28
limit size of message

### LINKED ISSUE #38: Reject Params too big
https://docs.sentry.io/hosted/clientdev/advanced/#variable-size

mitchellhenke:
Closes #38


lib/sentry/client.ex
```diff
@@ -286,7 +288,7 @@ defmodule Sentry.Client do
       event_id: event.event_id,
       culprit: event.culprit,
       timestamp: event.timestamp,
-      message: event.message,
+      message: String.slice(event.message, 0, @max_message_length),
```
jeregrine: should we show ... or [TRUNCATED] at the end? or do we not have guidance?

mitchellhenke: We could? The Ruby library does not though https://github.com/getsentry/raven-ruby/blob/120a416cf0df0a08803992a7c643741862f000ad/lib/raven/event.rb#L88

PR #302, opened 2018-08-06
[WIP] Elixir 1.7/OTP 21

mitchellhenke:
This PR will bundle significant changes, including:

- Switching from Poison to Jason for JSON operations
- Removal of the  `Sentry.Logger` `:error_logger` backend in favor of `Sentry.LoggerBackend` which leverages the new changes in Elixir 1.7 and OTP 21


lib/sentry/client.ex
```diff
@@ -75,7 +75,7 @@ defmodule Sentry.Client do
 
   defp encode_and_send(event, result) do
     render_event(event)
-    |> Poison.encode()
+    |> Jason.encode()
```
jeregrine: lets make this configurable a-la ecto/phoenix. "json_library"

mitchellhenke: This would take some more thought as it complicates the installation. If an app adds Sentry, and `:jason` is the default (but still an optional dep), and they don't have Jason as a dependency or another library configured, it'll break Sentry. Phoenix solves this with the generation of a project and including a default dependency in `mix.exs`.

We'd have to do something like: https://github.com/phoenixframework/phoenix/blob/18871c898fee3a91034cd23a4b86444449daa991/lib/phoenix.ex#L79-L90 as well

It also seems semi-scary that we can't enforce that the JSON library implements a protocol or interface we expect?  I don't see how Ecto/Phoenix catch that a bad configuration like `:sentry, json_library: :ranch` wouldn't work, and you don't find out until runtime?

I dunno how I feel for cost/benefit on it.

PR #320, opened 2018-12-19
Convenience function for capturing any exception produced by an anonymous function

shiroyasha:
A common pattern we found useful in our services is to wrap relevant code segments
in a `capture` function that logs errors and reraises them for further processing.

Example usage:

``` elixir
Sentry.capture_exception(fn ->
  1 / 0
end)
```

The above example would send the exception to Sentry.

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -110,6 +110,33 @@ defmodule Sentry do
   end
 
   @doc """
+  Captures and retrows exception raised in the callback function.
+  """
+  @spec capture_exception((() -> any())) :: send_result
+  def capture_exception(fun) do
+    fun.()
+  rescue
+    e ->
+      stacktrace = __STACKTRACE__
+
+      Sentry.capture_exception(e, stacktrace: stacktrace)
+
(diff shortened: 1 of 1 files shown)


shiroyasha:
Related exception and an example use case with GPRC Elixir https://github.com/tony612/grpc-elixir/issues/84.

mitchellhenke:
@shiroyasha thanks for opening this!  I'll have to do some thinking on including functionality like this.  It does look like the tests are failing, and I can take a closer look once that's taken care of 🙂

shiroyasha:
@mitchellhenke thanks for reviewing.

The PR probably needs a bit more love to be complete, like for example tests.
Also the name of the method might be a bit confusing.

Let me know if I can help and improve the code in any way possible.

mitchellhenke:
@shiroyasha I appreciate the effort and time you've put into the PR, but I will not be accepting it as I don't think the benefits outweigh having to support it within the library.

I'm happy to revisit if I've underestimated how many users would make use of the feature.

PR #326, opened 2019-01-31
Avoid an error when formatting remote data

schneiderderek:
We're running into 2 issues:

The first where `:inet.ntoa()` is returning `{:error, :einval}` causing the `|> to_string` call to blow up. This change checks the return of `:inet.ntoa/1` before passing it to `to_string/1` to avoid the above situation.

The second where `Plug.Conn.get_peer_data` is returning an error tuple.


mitchellhenke:
Thank you for contribution! 💖

PR #329, opened 2019-02-13
Update Sentry.Config moduledoc

LysanderGG:
I would suggest this edit because I think it can be an easy mistake to copy paste the tuple, change the env var and forget about the quotes (it happened to me :sweat_smile: )


mitchellhenke:
Thanks! 💖

PR #334, opened 2019-04-16
Allow configuration of Sentry log level

halostatue:
Implements the feature of #304

Because `Logger` has explicit permitted macro levels, I’ve put in some validation.

We’re currently experiencing periodic spike protection and the Sentry logging is also causing some spiking on our log service, too (and because of the `?\n` in the log, each Sentry send failure is taking up three log lines). I’d like to handle such logging through `after_send_event`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ erl_crash.dump
 *.ez
 /doc
 .DS_Store
+.elixir_ls/
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,3 @@
 erl_crash.dump
 *.ez
 /doc
-.DS_Store
-.elixir_ls/
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ For optional settings check the [docs](https://hexdocs.pm/sentry/readme.html).
 | `source_code_path_pattern` | False  | `"**/*.ex"` | |
 | `filter` | False | | Module where the filter rules are defined (see [Filtering Exceptions](https://hexdocs.pm/sentry/Sentry.html#module-filtering-exceptions)) |
 | `json_library` | False | `Jason` | |
-| `log_level` | False | `:warn` | |
+| `log_level` | False | `:warn` | This sets the log level used when Sentry fails to send an event due to an invalid event or API error. |
 
 An example production config might look like this:
 
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ For optional settings check the [docs](https://hexdocs.pm/sentry/readme.html).
 | `source_code_path_pattern` | False  | `"**/*.ex"` | |
 | `filter` | False | | Module where the filter rules are defined (see [Filtering Exceptions](https://hexdocs.pm/sentry/Sentry.html#module-filtering-exceptions)) |
 | `json_library` | False | `Jason` | |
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ erl_crash.dump
 *.ez
 /doc
 .DS_Store
+.elixir_ls/
```
mitchellhenke: This doesn't apply specifically to the repository and probably should not be included.

It's an unrelated request, but if you wanted to remove `.DS_Store`, that'd be cool too, but not necessary 🙂

Platform/tooling specific stuff typically ends up using git global ignore configuration: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

README.md
```diff
@@ -118,6 +118,7 @@ For optional settings check the [docs](https://hexdocs.pm/sentry/readme.html).
 | `source_code_path_pattern` | False  | `"**/*.ex"` | |
 | `filter` | False | | Module where the filter rules are defined (see [Filtering Exceptions](https://hexdocs.pm/sentry/Sentry.html#module-filtering-exceptions)) |
 | `json_library` | False | `Jason` | |
+| `log_level` | False | `:warn` | |
```
mitchellhenke: If you could add a short note about what the config is used for, that would be super.  Something along the lines of `This sets what level Sentry will log at when it fails to send due to an invalid event or API error`

lib/sentry/config.ex
```diff
@@ -14,6 +14,11 @@ defmodule Sentry.Config do
   @default_context_lines 3
   @default_sample_rate 1.0
 
+  @permitted_log_level_atom_values ~w(debug info warn error)a
+  @permitted_log_level_string_values ~w(debug info warn error)
```
mitchellhenke: The string values can be removed, and the atom set can be renamed to `permitted_log_level_values`.  Though Sentry does a lot of it currently, my goal is to limit how much the library transforms things from the original config.  Especially in cases like this where we can check the validity of the config early, I think it is okay to be stricter.

.gitignore
halostatue: Done.

README.md
halostatue: Done.

lib/sentry/config.ex
halostatue: Done.

lib/sentry/config.ex
```diff
@@ -127,6 +129,15 @@ defmodule Sentry.Config do
     get_config(:json_library, default: Jason, check_dsn: false)
   end
 
+  def log_level do
+    case get_config(:log_level, default: :warn, check_dsn: false) do
```
mitchellhenke: Both cases return `value` so I think we can drop the `case` and the function can return directly

```elixir
get_config(:log_level, default: :warn, check_dsn: false)
```

halostatue: Done.

halostatue:
Hopefully we can get a release soon that includes this so that I can start silencing some of our noisier errors on a spike and reporting the errors we care about (we’re getting a number of JSON encoding errors because of tuples in our data that we can’t quite localize).

mitchellhenke:
@halostatue just published `7.0.6` to Hex 🙂

PR #338, opened 2019-05-01
Logger metadata

### LINKED ISSUE #336: Logger metadata unavailable when using LoggerBackend
Currently the recommended way to attach Logger.metadata is via a presend hook that pulls out relevant metadata and attaches it to the event.  This works great under normal situations because the presend hook runs in the same process as where the exception occurred, making Logger.metadata accessible.

However when using the LoggerBackend the message is first sent to a gen event process before being turned into a Sentry exception.  The presend hook loses access to the original Logger.context.

This is fixable and there's a few ways to do but wanted to get it on the radar as I work on the issue

mitchellhenke:
Closes #336

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,16 @@ elixir:
   - 1.7
   - 1.8
 otp_release:
-  - 20.2
-  - 21.0
+  - 20.3
+  - 21.3
 env:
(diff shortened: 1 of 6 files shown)


thdxr:
Looks great, thanks for the fast turnaround

PR #341, opened 2019-05-08
default result type to :none

### LINKED ISSUE #360: Sentry not reporting errors because process gets killed
### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 22 [erts-10.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]
Elixir 1.9.0 (compiled with Erlang/OTP 20)
```
* Erlang/OTP version (erl):
```
Erlang/OTP 22 [erts-10.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]
```
* Sentry version (mix deps): 7.1.0
* Operating system: macOS

### Description

Errors are not reported to Sentry in some cases. This happens when the process which reports the error to Sentry is killed.
When the code [here](https://github.com/getsentry/sentry-elixir/blob/dfcbf2349cd839f29b55cf231ba63e7cdef08655/lib/sentry/phoenix_endpoint.ex#L34) executes, the response has already been sent to the client. This means that the client may close the connection. If client closes the connection, the process which handles the request (and reports errors to Sentry) is shut down.

### Expected Behavior
 
Errors are always reported to Sentry, regardless of how much time it takes to report them

### Actual Behavior

Errors are not sent to Sentry if they are not processed quickly enough

### Steps to Reproduce

I have created a simple PoC.

First, I modified Sentry library to make it slower. This change can be found here https://github.com/aronglia/sentry-elixir/commit/290596596cb05e4d6fe739d2a0d6d9cdf266fa4c

PoC which uses Sentry with that change can be found here https://github.com/aronglia/sentry-bug

1. Run PoC with `mix phx.server`
2. Make a request with
```curl http://localhost:4000```
Observe that the loop execution stops at random time and the code which should report error to Sentry is not reached.
3. Make a request with
```
curl -H 'Connection: keep-alive' http://localhost:4000 http://localhost:4000/x
```
Observe that the execution of the loop does not stop since the client does not close the connection.

Similar issue https://github.com/ninenines/cowboy/issues/1397

mitchellhenke:
https://github.com/elixir-lang/elixir/issues/9028 shows a Sentry bug from #174 in which the current behavior causes messages to build up in the calling process.  This is likely primarily affecting processes calling `Sentry.capture_exception/1`, as `Sentry.LoggerBackend` will handle the messages on `handle_info/2` and `Sentry.Plug` will generally live in a process that's about to exit.

The default should have been `:none` all along, so this PR fixes that, and the documentation already discusses the potential issues with `:async` and `Task.Supervisor.async_nolink/2`

This closes #360

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -56,12 +56,12 @@ defmodule Sentry.Client do
   The event is dropped if it all retries fail.
 
   ### Options
-  * `:result` - Allows specifying how the result should be returned. Options include `:sync`, `:none`, and `:async`.  `:sync` will make the API call synchronously, and return `{:ok, event_id}` if successful.  `:none` sends the event from an unlinked child process under `Sentry.TaskSupervisor` and will return `{:ok, ""}` regardless of the result.  `:async` will start an unlinked task and return a tuple of `{:ok, Task.t}` on success where the Task can be awaited upon to receive the result asynchronously.  When used in an OTP behaviour like GenServer, the task will send a message that needs to be matched with `GenServer.handle_info/2`.  See `Task.Supervisor.async_nolink/2` for more information.  `:async` is the default.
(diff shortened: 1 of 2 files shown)


mitchellhenke:
Note to myself, we should consider returning the generated ID of the event regardless of the result type to make stuff like #365 easier

jeroenvisser101:
@mitchellhenke would love to be able to configure defaults like these. Was looking for a way to change this, as we tried reporting unknown messages to Sentry (which ended up being an infinite loop). Being able to configure this allows it to not require a major release (yet), as the default could stay the same. Either way, we would also prefer to use `:none`, so configurable or not, we would love to see this in a future version :)

mitchellhenke:
@jeroenvisser101 a configuration for that is possible, but the default needs to change either way due to the inherent issues mentioned in the issue 🙂

PR #344, opened 2019-05-14
Fixes dialyzer warning when using the Sentry.Phoenix.Endpoint macro

heymackey:
When using the Sentry.Phoenix.Endpoint on a Phoenix project with Dialyzer, set with the `:unmatched_returns` flag, Dialyzer will complain with a `../endpoint.ex:1:umatched_return`, with the expectation of the 'send_result()' type signature.

By assigning the return of `Sentry.capture_exception/2` to `_`, projects using this macro won't see the warning.

### DIFF
--- a/lib/sentry/plug.ex
+++ b/lib/sentry/plug.ex
@@ -171,22 +171,12 @@ if Code.ensure_loaded?(Plug) do
         headers: handle_data(conn, header_scrubber),
         env: %{
           "REMOTE_ADDR" => remote_address(conn.remote_ip),
-          "REMOTE_PORT" => remote_port(conn),
+          "REMOTE_PORT" => Plug.Conn.get_peer_data(conn).port,
           "SERVER_NAME" => conn.host,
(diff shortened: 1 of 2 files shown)


lib/sentry/plug.ex
```diff
@@ -171,24 +171,14 @@ if Code.ensure_loaded?(Plug) do
         headers: handle_data(conn, header_scrubber),
         env: %{
           "REMOTE_ADDR" => remote_address(conn.remote_ip),
-          "REMOTE_PORT" => remote_port(conn),
+          "REMOTE_PORT" => Plug.Conn.get_peer_data(conn).port,
```
mitchellhenke: this function was added in https://github.com/getsentry/sentry-elixir/pull/326 as `Plug.Conn.get_peer_data` was returning an `:error` tuple, but I'm not sure if that's possible looking at Plug docs as it is now, and the typespecs seem to confirm that.

I'm going to assume it's good now though 🙂

PR #346, opened 2019-05-21
Fix documentation error relating to File.cwd()!

nthock:
`File.cwd()!` is not a valid elixir syntax. I have amended it to `File.cwd!()` instead.


mitchellhenke:
Thank you! 💖💖💖💖

PR #347, opened 2019-05-27
Docs: Add parens to File.cwd! to prevent confusion

mathiasose:
There seems to be a bug in the documentation engine that results in the docs coming out wrong. It is inserting parens into `File.cwd!` at the wrong position, creating invalid syntax.

image
image

I didn't find an issue about this in `ex_doc`, but I didn't look very closely. It might be that this has been fixed in a release of `ex_doc` and you might not need this PR at all if you update. Or maybe you guys should open an issue yourself :man_shrugging:


mitchellhenke:
Thanks for the fix! 💖

PR #348, opened 2019-05-27
Check that DSN is binary before trying URI.parse

mathiasose:
I was setting up Sentry with DSN read from env, and got some crashes when the required variable was not set, because `URI.parse` was being called with `nil`, which doesn't match.

To fix this and get a nicer warning message instead of an exception (like #218), I added a `is_binary` guard in `Sentry.Client.get_dsn`. Plus a couple of tests to verify that it works as intended.


mitchellhenke:
Thank you for the contribution (and tests)! 💖

PR #351, opened 2019-06-20
Remove newline from log_api_error

### LINKED ISSUE #343: Remove newline from API error logging
### Environment

* Elixir version (elixir -v): 1.7.4
* Erlang/OTP version (erl): 21
* Sentry version (mix deps): 7.0.6
* Operating system: MacOS

### Description

The private function `Sentry.Client.log_api_error/1` ([lib/sentry/client.ex:338-345](https://github.com/getsentry/sentry-elixir/blob/ace9faf4f518a0764a81f40ac6ac580f62fe41e6/lib/sentry/client.ex#L338-L345)) currently includes a newline in its log body, which is problematic when capturing Erlang logs to something like an ELK stack, because `Failed to send Sentry event` and the body of the failure are on two different lines.

I would recommend that this change to something like:

```elixir
  defp log_api_error(body) do
    Logger.log(
      Config.log_level(),
      fn ->
        ["Failed to send Sentry event: ", inspect(body)]
      end
    )
  end
```

halostatue:
Resolves #343

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -340,7 +340,7 @@ defmodule Sentry.Client do
     Logger.log(
       Config.log_level(),
       fn ->
-        ["Failed to send Sentry event.", ?\n, body]
+        ["Failed to send Sentry event.", inspect(body)]
       end
     )
   end


mitchellhenke:
Thanks! 💖

PR #354, opened 2019-07-17
return error information in after_send_event

### LINKED ISSUE #342: Details for after_send_event callback are incorrect
### Environment

* Elixir version (elixir -v): 1.7.4
* Erlang/OTP version (erl): 21
* Sentry version (mix deps): 7.0.6
* Operating system: MacOS

### Description

The description of the `:after_send_event` callback appears to be incorrect:

> * `:after_send_event` - callback that is called after attempting to send an event.
> Accepts an anonymous function or a {module, function} tuple. The result of the HTTP call as well as the event will be passed as arguments. The return value of the callback is not returned.

The spec for `maybe_call_after_send_event/2` is `@spec maybe_call_after_send_event(send_event_result, Event.t()) :: Event.t()`, and `@type send_event_result :: {:ok, Task.t() | String.t() | pid()} | :error | :unsampled`. So an event result of `:error` includes no information related to _why_ the event failed.

Part of the reason I implemented #334 was so that I could switch Sentry logging to `:debug` and specifically suppress logging for Sentry rate-limit rejections but explicitly log other errors, such as that seen in #240 (we an alert in our ELK system for these errors, but these are currently swamped by spike protection 429s). I can’t do that with an `:error` result, but would be better served by an `{:error, error}` tuple result.

mitchellhenke:
closes #342

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -4,7 +4,8 @@ config :sentry,
   environment_name: :test,
   included_environments: [:test],
   client: Sentry.TestClient,
-  hackney_opts: [recv_timeout: 50]
+  hackney_opts: [recv_timeout: 50],
+  max_client_attempts: 1
 
 config :ex_unit,
   assert_receive_timeout: 500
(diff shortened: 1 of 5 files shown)


mitchellhenke:
this is handled as part of the serious bug fixed in #372

PR #364, opened 2019-10-16
before_send_event filter

### LINKED ISSUE #363: Enhancement: Use `before_send_event` for filtering
In other Sentry SDK's, the equivalent of `before_send_event` (`beforeSend` in JS, `before_send` in Python, etc) serve two purposes: they let you modify the event before it's sent, but they also let you examine it and decide whether or not you even want to send it in the first place. (In those SDK's, if `beforeSend`/`before_send` returns something falsy, the event is dropped.) Docs on this [here](https://docs.sentry.io/error-reporting/configuration/?platform=browser#before-send).

A common use case for this is filtering events for various reasons, because it's more flexible than any built-in filter could be, as it lets you filter on literally any piece of data in the event. Opening this tracking issue to document possible support for this behavior in `sentry-elixir`.

mitchellhenke:
closes #363

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -43,6 +43,7 @@ defmodule Sentry.Client do
   @type send_event_result ::
           {:ok, Task.t() | String.t() | pid()} | :error | :unsampled | :excluded
   @type dsn :: {String.t(), String.t(), String.t()} | :error
+  @type result :: :sync | :none | :async
   @sentry_version 5
   @max_attempts 4
   @hackney_pool_name :sentry_pool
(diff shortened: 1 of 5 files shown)


lib/sentry/client.ex
```diff
@@ -257,14 +276,14 @@ defmodule Sentry.Client do
     result
   end
 
-  @spec maybe_call_before_send_event(Event.t()) :: Event.t()
+  @spec maybe_call_before_send_event(Event.t()) :: Event.t() | false
```
jeregrine: Why false instead of :error or something more descriptive.

mitchellhenke: this kind of wraps the user-implemented callback, so there shouldn't be an error case. The user's callback should return an event or something falsey. If it's an event, it means Sentry will send as normal, but if the callback returns something false-y, this function will return `false` (and we won't send).

PR #368, opened 2019-11-01
Additional example to handle sentry in specific environments in phoenix (README.md)

cdvv7788:
Following the instructions, I was receiving errors in the test environment. I added a small example of how I fixed it to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,15 @@ use Plug.ErrorHandler
 use Sentry.Plug
 ```
 
+In phoenix, this will add Sentry to all of your environments. If you want to enable Sentry in an specific environment, you can do it this way:
+
+```elixir
+if Mix.env() == :prod do
+  use Sentry.Plug
+end
+```
+
+
 If you are using Phoenix, you can also include [Sentry.Phoenix.Endpoint](https://hexdocs.pm/sentry/Sentry.Phoenix.Endpoint.html) in your Endpoint. This module captures errors occurring in the Phoenix pipeline before the request reaches the Router:
 
 ```elixir


mitchellhenke:
@cdvv7788 can you elaborate on the errors you are receiving?

cdvv7788:
@mitchellhenke Just warnings when running the controller tests. Situations where normally sentry sends the message. It complains about an invalid DSN.
Obviously, setting the DSN for the test settings is not an option.
Is there a better approach to handle this?

mitchellhenke:
Controlling the situations in which Sentry runs should be done via the `:environment_name` and `:included_environments` configuration keys. An example is [here](https://github.com/getsentry/sentry-elixir#configuration):

```elixir
config :sentry,
  dsn: "https://public_key@app.getsentry.com/1",
  # ...
  environment_name: Mix.env(),
  included_environments: [:prod]
```

cdvv7788:
@mitchellhenke Well, my setup differs a little from what is in https://github.com/getsentry/sentry-elixir#configuration
I have the configuration in `prod.secret.exs`, to load the `dsn` from an environment variable.
If you just use the `Sentry.Plug` in the router, it will try to load sentry when it has not been configured.

I think that loading the dsn from an environment variable is a pretty common pattern. Is it still recommended to add it in the `config.exs` file, even when it is only wanted for sites that use `prod.exs` settings? (staging and prod use that one, for example).

mitchellhenke:
@cdvv7788 The method for loading the DSN shouldn't cause any issues around this, it's an issue with Sentry's defaults, which is hard to make fit for the different use cases (though yours is a common pattern).

Sentry is trying to send events while in the `:test` MIX_ENV because the configuration for `:environment_name` is matching whatever is in `:included_environments`.

Could you try adding the following to `config/config.exs`?

```elixir
# config/config.exs

config :sentry,
  included_environments: []
```

cdvv7788:
@mitchellhenke that worked. Does this deserve something in the readme or should I just close the PR?

mitchellhenke:
@cdvv7788 Closing I think is fine. I've rewritten the README a little bit to hopefully make it clearer, and I'm working on some configuration-related things to make the initial setup better.

PR #372, opened 2019-11-19
No logger logging

### LINKED ISSUE #371: Unreachable/misconfigured Sentry hangs the Elixir app
### Environment

* Elixir version (elixir -v): 1.9.1
* Erlang/OTP version (erl): 22
* Sentry version (mix deps): 7.2.0
* Operating system: macOS 10.13

### Description

If Sentry is misconfigured or otherwise can't reach the API, and is configured to use the Logger backend, any Logger usage in the app will hang indefinitely after a certain number of attempted logged messages.

### Expected Behavior

The app continues to work and logger messages are dropped or otherwise properly handled when Sentry is misconfigured/unreachable.

### Actual Behavior

The app hangs indefinitely on any Logger call after a number of failures.

### Steps to Reproduce

See https://github.com/dkulchenko/sentrytestcase. This is a basic Phoenix app configured with `mix phx.new`, but I've added a task supervisor to the supervision tree, added sentry, and set up the logger backend in application.ex.

From within this app, if you run `mix run test.exs`, the output will be:

```
processing item 1
processing item 2
processing item 3
processing item 4
processing item 5
processing item 6
processing item 7
processing item 8
processing item 9
processing item 10
processing item 11
processing item 12
processing item 13
processing item 14
processing item 15
processing item 16
processing item 17
processing item 18
processing item 19
processing item 20
processing item 21
```

and then hangs indefinitely. If the Sentry logger backend is removed, the test case processes all 1000 items as expected.

Note that I haven't configured Sentry at all in config.exs, but I'm guessing this would happen as well if it was properly configured but the API was unreachable/down.

mitchellhenke:
The cause of #371 is that logging from within the Sentry.LoggerBackend can deadlock Logger itself. When Logger receives a bunch of messages and can't process them fast enough, it switches into synchronous mode (default is queue length of 20). This means when Sentry.LoggerBackend tries to log while Logger is in synchronous mode, it will deadlock because Sentry.LoggerBackend is waiting for Logger to return a response while Logger is waiting for Sentry.LoggerBackend to process messages.

I started a little bit of a refactor in how Client handles errors to separate the Logger pieces from the Client to potentially make changes going forward, but the fix is relatively straightforward for now in that if the event_source on the event is the LoggerBackend, we don't log.

I'm not expecting many users to see this, especially in production, but this is dangerous behavior that should not be possible 🙂

closes #371

### DIFF
--- a/test/mix/sentry.send_test_event_test.exs
+++ b/test/mix/sentry.send_test_event_test.exs
@@ -55,7 +55,7 @@ defmodule Mix.Tasks.Sentry.SendTestEventTest do
            """
   end
 
-  test "handles :error when Sentry server is failing" do
+  test "handles error when Sentry server is failing" do
     bypass = Bypass.open()
 
     Bypass.expect(bypass, fn conn ->
@@ -78,7 +78,7 @@ defmodule Mix.Tasks.Sentry.SendTestEventTest do
                     hackney_opts: [recv_timeout: 50]
 
                     Sending test event...
-                    Error sending event!
+                    Error sending event: "Received 500 from Sentry server: "
                     """
            end) =~ "Failed to send Sentry event"
   end
--- a/lib/mix/tasks/sentry.send_test_event.ex
+++ b/lib/mix/tasks/sentry.send_test_event.ex
@@ -58,8 +58,8 @@ defmodule Mix.Tasks.Sentry.SendTestEvent do
         {:ok, id} ->
           Mix.shell().info("Test event sent!  Event ID: #{id}")
 
-        :error ->
-          Mix.shell().info("Error sending event!")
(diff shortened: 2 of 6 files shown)


test/mix/sentry.send_test_event_test.exs
```diff
@@ -78,7 +78,7 @@ defmodule Mix.Tasks.Sentry.SendTestEventTest do
                     hackney_opts: [recv_timeout: 50]
 
                     Sending test event...
-                    Error sending event!
+                    Error sending event: "Received 500 from Sentry server: "
```
jeregrine: prolly dont need the the last colon after server/

PR #387, opened 2020-03-02
allow configuring feedback form

### LINKED ISSUE #386: Customizing user feedback form
### Description
With the new feature to enable user feedback on errors with code 500, the form shows up (thanks for adding it btw). Now I would like to be able to customize it according to the options available at https://docs.sentry.io/enriching-error-data/user-feedback/?platform=django#customizing-the-widget. Specifically, the lang, the title, subtitle, etc. The information to modify the dsn and the eventId is not priority for now as the default implementation works well.

### Expected Behavior
The SDK allows to configure the feedback form to send information to sentry

### Actual Behavior
The SDK supports the user feedback feature, but provides no way to customize it

mitchellhenke:
closes #386

### DIFF
--- a/lib/sentry/plug.ex
+++ b/lib/sentry/plug.ex
@@ -100,13 +100,17 @@ if Code.ensure_loaded?(Plug) do
 
     ### Collect User Feedback after Error
 
-    Sentry allows collecting user feedback after they hit an error.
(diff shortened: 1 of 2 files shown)


cdvv7788:
@mitchellhenke This works! I tested it with several options and they are being passed correctly.

mitchellhenke:
@cdvv7788 thanks for trying it out! going to do a release today with the changes

PR #391, opened 2020-04-24
fix request url port in payloads for HTTPS requests

smetana:
We have noticed invalid request urls for HTTPS requests (like https://mysite.com:80)
This is because request URLs are built from conn properties like
"#{conn.scheme}://#{conn.host}:#{conn.port}#{conn.request_path}"

Plug.Conn does not allow to configure port for HTTPS requests (conn.port is always 80 and ignored) when using Plug.SSL. https://hexdocs.pm/plug/Plug.SSL.html

It is better to use Plug.Conn.request_url() instead of building request URL manually

### DIFF
--- a/lib/sentry/plug.ex
+++ b/lib/sentry/plug.ex
@@ -230,7 +230,7 @@ if Code.ensure_loaded?(Plug) do
         |> Plug.Conn.fetch_query_params()
 
       %{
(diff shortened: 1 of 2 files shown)


mitchellhenke:
Awesome, thank you!

PR #393, opened 2020-04-29
Cache environment config in application config

devonestes:
This will speed up subsequent calls to `Config` when using environment
config. Now, once we get the environment config once, we cache that in
the application config. This also makes it so other OTP applications
don't need to edit the `sentry` application config (which is best
avoided if possible).


mitchellhenke:
@devonestes thank you! I have plans to redo how config is handled at some point, but greatly appreciate this.

PR #398, opened 2020-06-09
async_nolink -> start_child

josevalim:
Since we don't await, it is not necessary to use an async function.


josevalim:
Ignore me. The away is elsewhere. :)

mitchellhenke:
@josevalim yep I made a mistake on this initially, but plan to have a major release with a handful of breaking changes that includes fixing this.

PR #400, opened 2020-06-09
Make hackney an optional dependency

josevalim:
This PR makes it easier to replace HTTP clients by
changing the old Sentry.HTTPClient behaviour to focus
exclusively on the HTTP requests bits.

This is a breaking change for developers implementing
custom clients - although doing so was unlikely, as the
client API was complex to implement. Other than that,
the Sentry.Client remains the same.

Two new configs have also been added, `send_max_attempts`
and `send_result`. They are necessary for Sentry's own
test suite, since `Sentry.TestClient` has now been fully
removed.


mix.exs
```diff
@@ -9,11 +9,10 @@ defmodule Sentry.Mixfile do
...
       package: package(),
       deps: deps(),
-      elixirc_paths: elixirc_paths(Mix.env()),
       dialyzer: [
         plt_file: {:no_warn, "priv/plts/dialyzer.plt"},
         plt_add_deps: :app_tree,
-        plt_add_apps: [:mix, :plug]
+        plt_add_apps: [:mix, :plug, :hackney]
       ],
       docs: [extras: ["README.md"], main: "readme"]
```
josevalim, line 17:
```suggestion
      docs: [extras: ["README.md"], main: "readme"],
      xref: [exclude: [:hackney]]
```

mitchellhenke:
thank you!! 💖

PR #402, opened 2020-06-17
PlugCapture and PlugContext

mitchellhenke:
@josevalim if you don't mind taking a peek 🙂


README.md
```diff
@@ -30,26 +22,67 @@ defp deps do
 end
 ```
 
-### Setup with Plug or Phoenix
+### Setup with Plug and Phoenix
+Capturing errors in Plug applications is done with `Sentry.PlugContext` and `Sentry.PlugCapture`. `Sentry.PlugContext` adds contextual metadata from the current request which is then included in errors that are captured and reported by `Sentry.PlugCapture`.
 
-In your Plug.Router or Phoenix.Router, add the following lines:
+If you are using Phoenix, first add `Sentry.PlugCapture` above the `use Phoenix.Endpoint` line in your endpoint file. `Sentry.PlugContext` should be added below `Plug.Parsers`.
```
josevalim, line 28: Nitpick but it may be clearer?

```suggestion
If you are using Phoenix, first add `Sentry.PlugCapture` before the `use Phoenix.Endpoint` line in your endpoint file. `Sentry.PlugContext` should be added below `Plug.Parsers`.
```

lib/sentry/client.ex
```diff
@@ -119,6 +119,11 @@ defmodule Sentry.Client do
           maybe_log_result(result)
         end
 
+        if match?({:ok, _}, result) do
+          {event.event_id, event.event_source}
```
josevalim, line 123:
```suggestion
```

josevalim:
Looks great! I have dropped just one minor comment. Two other suggestions are:

1. You are testing `last_event_id_and_source` in PlugCapture but you may also want to test it directly when calling `Sentry.capture_exception` directly

2. I would remove phoenix as a dependency. Potentially bump the plug_cowboy dependency to "~> 2.3" since we will need it for excluding domains (or do it in another PR)

README.md
josevalim, line 28: IIRC, we had to add it above if we were going to render something, but we are not going to render anything, so the order most likely does not matter. :)

mitchellhenke: I thought it was necessary because PlugContext may try to manipulate/scrub the body of the request, and if it wasn't parsed before then it could cause issues?

josevalim: Sorry, I was not clear. :D I wanted to say `Sentry.PlugCapture` does not need to be above/before "use Phoenix.Endpoint"/"use Plug.Router" anymore. :) But the parsers one is recommended, yes. :)

mitchellhenke: ohh, my fault, thank you, yeah

mitchellhenke: I've written a few more tests, and it seems like PlugCapture does have to go first in the Endpoint. If you flip the lines [here](https://github.com/getsentry/sentry-elixir/pull/402/files#diff-24ca02f3e0bdcfc2284daa120a1c5889R19-R20) and run the tests it fails because the ErrorView doesn't get anything from last_event_id.

josevalim: Ah, good call!

PR #403, opened 2020-06-19
Logger Backend improvements

### LINKED ISSUE #383: Exceptions with breadcrumbs using Sentry.LoggerBackend
### Environment

* Elixir version (elixir -v): All 
* Erlang/OTP version (erl): all 
* Sentry version (mix deps): 7.x
* Operating system: All

### Description
When a process crashes or an exception is raised sentry-elixir will capture the exception correctly. However, if you use breadcrumbs or other context sentry-elixir will not pass this correctly.  

As I understand the code we are using the Process dictionary to hold all context.  The problem comes up when you are utilizing `Sentry.LoggerBackend` as your logger backend.  The logger and the process the exception occurred in are not the same resulting in no context being sent.  

### Expected Behavior
Context will be passed correctly when errors and exceptions happen.  

### Actual Behavior
Context is lost because the logger is a different process.

### Steps to Reproduce
```elixir
defmodule Test.MimicGenserver do
  alias GenServer
  use GenServer

  def start_link(_) do
    GenServer.start_link(__MODULE__, %{}, name: __MODULE__)
  end

  def init(_) do
    {:ok, %{}}
  end

  def error, do: GenServer.call(__MODULE__, :error)

  def terminate(reason, state) do
    :normal
  end

  def handle_call(:error, _from, _state) do
    Sentry.Context.add_breadcrumb(%{category: "will not show"})
    {:stop, "crash", :error, nil}
  end
end

defmodule ClientTest do
  use ExUnit.Case

  alias Test.MimicGenserver
  import ExUnit.CaptureLog

  test "if breadcrumb exists" do
    # assert {:ok, %{"test" => "test"}} = Client.make_request(request())
    bypass = Bypass.open()
    pid = self()
    start_supervised!(MimicGenserver)

    Bypass.expect(bypass, fn conn ->
      {:ok, body, conn} = Plug.Conn.read_body(conn)
      decoded_body = Jason.decode!(body)
      assert body =~ "will not show"
      send(pid, "API called")
      Plug.Conn.resp(conn, 200, ~s<{"id": "340"}>)
    end)

    Application.put_env(:sentry, :dsn, "http://public:secret@localhost:#{bypass.port}/1")

    capture_log(fn ->
      MimicGenserver.error()
      assert_receive "API called"
    end)
  end

end

```

### Solution?
I threw together a solution last night that I would be willing to clean up a bit more and submit if you are open to PR's and adding another dependency.  

My idea is to move away from the process dictionary and to utilize something like `cachex` as the context store.  When the context is set we could obtain the PID of the process and store that as the key in cachex.  This would allow us to utilize 98% of the context module that's in place.  When the exception is logged we can grab the PID from the metadata allowing us to obtain the context for the specific process.  Allowing us to pass the context with the error.  

The only limitation I can think of with this solution is how we would monitor the process and delete keys when the process is killed normally.  

Other than that I think this could be a major version bump as you would need to also incorporate a supervisor tree and start sentry as an application within your application.  If you didnt want to do a major version bump we could possibly break this out in to another module and switch between the process dictionary context and cachex context depending on if a user mounted the application. 

### Final thoughts...
You  may already have plans to add this or have another idea.  Just wanted to throw some thoughts your way in case you need a helping hand to implement this.  It's something I feel we are going to need to do for multiple apps we use. That or we may implement some custom functionality that mimics the above.  Utilizing a new logger, custom context module and a custom before_send event. 

Thanks for the work!  The code was extremely easy to follow and took minimal effort to understand.

### LINKED ISSUE #397: Multiple reports per exception
### Environment

* Elixir version (elixir -v): 1.10.3
* Erlang/OTP version (erl): 23.0.2
* Sentry version (mix deps): 7.2.4
* Operating system: Linux
* Is this an umbrella application?: No
* Release type (Distillery, mix release, Mix, etc.): mix release

### Description

Every exception in a plug or controllers gets reported three times.

### Expected Behavior

A single event per exception

### Actual Behavior

There's three Sentry events per exception, with varying amounts of metadata:

* One report has the usual metadata, respecting header scrubber setting
* Another one looks similar, but includes sensitive headers (it's not scrubbed)
* The final one is missing almost all metadata

I believe these come from Sentry.Plug, Sentry.Phoenix.Endpoint, and Sentry.LoggerBackend respectively.

### Steps to Reproduce

Setup a new Phoenix app with:

* Sentry.Plug
* Sentry.Phoenix.Endpoint
* Sentry.LoggerBackend

then add `raise "boom"` inside a controller and try to load the page.

I suspect the Phoenix and Cowboy versions are important. This is with Phoenix 1.5.3 and Cowboy 2.8.0. The LoggerHandler seems to check for Cowboy2Handler in the stack trace, which isn't present. Here'a sample `:crash_reason` as received by the logger backend:

```
[
   {MyAppWeb.TheController, :index, 2,
    [file: 'lib/myapp_web/controllers/the_controller.ex', line: 22]},
   {MyAppWeb.TheController, :action, 2,
    [file: 'lib/myapp_web/controllers/the_controller.ex', line: 1]},
   {MyAppWeb.TheController, :phoenix_controller_pipeline, 2,
    [file: 'lib/myapp_web/controllers/the_controller.ex', line: 1]},
   {Phoenix.Router, :__call__, 2, [file: 'lib/phoenix/router.ex', line: 352]},
   {MyAppWeb.Router, :call, 2, [file: 'lib/plug/error_handler.ex', line: 65]},
   {MyAppWeb.Endpoint, :plug_builder_call, 2,
    [file: 'lib/myapp_web/endpoint.ex', line: 1]},
   {MyAppWeb.Endpoint, :"call (overridable 3)", 2,
    [file: 'lib/plug/debugger.ex', line: 132]},
   {MyAppWeb.Endpoint, :call, 2,
    [file: 'lib/myapp_web/endpoint.ex', line: 1]}
 ]
```

This is similar to https://github.com/getsentry/sentry-elixir/issues/374.

I'm not sure what causes the duplication between Sentry.Phoenix.Endpoint and Sentry.Plug.

mitchellhenke:
With significant initial work from @josevalim 

- Ignores Plug/Cowboy process crashes by excluding by Logger `:domain`
- Allows sending all messages received by Logger
- Non-Sentry metadata to be included in reports is now specified by a list of keys

closes #383, closes #397

### DIFF
--- a/lib/sentry/logger_backend.ex
+++ b/lib/sentry/logger_backend.ex
@@ -127,7 +127,7 @@ defmodule Sentry.LoggerBackend do
     end
   end
 
-  defp get_sentry_from_callers([head | tail]) do
+  defp get_sentry_from_callers([head | tail]) when is_pid(head) do
     with {:dictionary, [_ | _] = dictionary} <- :erlang.process_info(head, :dictionary),
          %{sentry: sentry} <- dictionary[:"$logger_metadata$"] do
       sentry
--- a/lib/sentry/logger_backend.ex
+++ b/lib/sentry/logger_backend.ex
@@ -112,7 +112,10 @@ defmodule Sentry.LoggerBackend do
         Sentry.capture_exception(exception, [stacktrace: stacktrace] ++ opts)
 
       {other, stacktrace} when is_list(stacktrace) ->
-        Sentry.capture_exception(Sentry.CrashError.exception(other), [stacktrace: stacktrace] ++ opts)
+        Sentry.capture_exception(
+          Sentry.CrashError.exception(other),
+          [stacktrace: stacktrace] ++ opts
+        )
 
       _ ->
         if state.capture_log_messages do
--- a/lib/sentry/logger_backend.ex
+++ b/lib/sentry/logger_backend.ex
@@ -1,54 +1,50 @@
 defmodule Sentry.LoggerBackend do
   @moduledoc """
-  Report logger events.
+  Report Logger events like crashed processes to Sentry. To include in your
+  application, add this module to your Logger backends:
 
       config :logger,
-        backends: [:console, BytepackWeb.SentryLoggerBackend]
+        backends: [:console, Sentry.LoggerBackend]
 
-  It sends all messages to Sentry, so it is recommended
-  to disable Sentry for non-prod environments and set the
-  log level to at least `:warn` in prod:
+  Sentry context will be included in metadata in reported events. Example:
 
-      config :logger, BytepackWeb.SentryLoggerBackend,
-        level: :warn,
-        metadata: [:foo_bar]
-
-  You can add request metadata to the reports like this:
-
-      SentryLoggerBackend.context(:request, %{ 
-        url: Plug.Conn.request_url(conn),
-        method: conn.method,
-        query_string: conn.query_string,
-        env: %{
-          "SERVER_NAME" => conn.host,
-          "SERVER_PORT" => conn.port,
-          "REQUEST_ID" => Plug.Conn.get_resp_header(conn, request_id) |> List.first()
-        }
-      })
-
-  You can add user metadata to the reports like this:
-
-      SentryLoggerBackend.context(:user, %{
+      Sentry.Context.set_user_context(%{
         user_id: current_user.id
       })
 
+  ## Configuration
+
+  * `:excluded_domains` - Any messages with a domain in the configured
+  list will not be sent. Defaults to `[:cowboy]` to avoid double reporting
+  events from `Sentry.PlugCapture`.
+
+  * `:metadata` - To include non-Sentry Logger metadata in reports, the
+  `:metadata` key can be set to a list of keys. Metadata under those keys will
+  be added in the `:extra` context under the `:logger_metadata` key  Defaults
+  to `[]`
+
+  * `:level` - The minimum [Logger level](https://hexdocs.pm/logger/Logger.html#module-levels) to send events for.
+  Defaults to `:error`.
+
+  * `:send_all_messages` - When `true`, this module will send all Logger
+  messages. Defaults to `false`, which will only send messages with metadata
+  that has the shape of an exception and stacktrace.
+
+  Example:
+
+      config :logger, Sentry.LoggerBackend,
+        # Also send warn messages
+        level: :warn,
+        # Send messages from Plug/Cowboy
+        excluded_domains: [],
+        # Include metadata added with `Logger.metadata([foo_bar: "value"])`
+        metadata: [:foo_bar],
+        # Send messages like `Logger.error("error")` to Sentry
+        send_all_messages: true
   """
   @behaviour :gen_event
 
-  defstruct level: :warn, metadata: [], excluded_domains: [:cowboy]
-
-  def context(key, value) when is_atom(key) and is_map(value) do
-    {sentry, metadata} =
-      case :logger.get_process_metadata() do
-        %{sentry: sentry} = metadata -> {sentry, metadata}
-        %{} = metadata -> {%{}, metadata}
-        :undefined -> {%{}, %{}}
-      end
-
-    sentry = Map.update(sentry, key, value, &Map.merge(&1, value))
-    :logger.set_process_metadata(Map.put(metadata, :sentry, sentry))
-    :ok
-  end
+  defstruct level: :error, metadata: [], excluded_domains: [:cowboy], send_all_messages: false
 
   def init(__MODULE__) do
     config = Application.get_env(:logger, __MODULE__, [])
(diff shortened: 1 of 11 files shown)


lib/sentry/logger_backend.ex
```diff
@@ -128,27 +97,54 @@ defmodule Sentry.LoggerBackend do
...
+        extra: %{logger_metadata: logger_metadata(meta, state), logger_level: level},
+        result: :none
+      ] ++ Map.to_list(sentry)
+
+    case meta[:crash_reason] do
+      {exception, stacktrace} when is_list(stacktrace) ->
+        Sentry.capture_exception(exception, [stacktrace: stacktrace] ++ opts)
+
+      reason when is_atom(reason) and not is_nil(reason) ->
+        Sentry.capture_exception(reason, opts)
```
josevalim, line 115: Do you know when this would ever be the case?

lib/sentry/logger_backend.ex
```diff
@@ -128,27 +97,54 @@ defmodule Sentry.LoggerBackend do
...
+    end
+  end
 
-    %{
-      state
-      | level: level,
-        include_logger_metadata: include_logger_metadata,
-        ignore_plug: ignore_plug
-    }
+  defp get_sentry_from_callers([head | tail]) do
```
josevalim, line 130:
```suggestion
  defp get_sentry_from_callers([head | tail]) when is_pid(head) do
```

Otherwise a list of `callers: [1]` would make it fail.

test/logger_backend_test.exs
```diff
@@ -205,68 +211,165 @@ defmodule Sentry.LoggerBackendTest do
...
+      Sentry.TestGenServer.add_logger_metadata(pid, :atom, :atom)
+      Sentry.TestGenServer.add_logger_metadata(pid, :number, 43)
+      Sentry.TestGenServer.add_logger_metadata(pid, :list, [])
+      Sentry.TestGenServer.invalid_function(pid)
+      assert_receive "terminating"
+      assert_receive "API called"
+    end)
+  end
+
+  test "sends all messages if capture_log_messages is true" do
```
josevalim, line 302: Let's add a test that the level is respected. For example, Logger.debug is never sent. :)

lib/sentry/logger_backend.ex
josevalim: We should probably add a test for malformed callers. :)

lib/sentry/logger_backend.ex
mitchellhenke: yeah, I meant to put some details in the PR when I made these changes, but forgot.

A few of the cases don't match `%{__exception__: true}`. Specifically:

https://github.com/getsentry/sentry-elixir/pull/403/files#diff-c895426ce24017d1803054d94e4d71c3L41 crash_reason is:
`crash_reason: {{:bad_return_value, "I am throwing"}, []}`

https://github.com/getsentry/sentry-elixir/pull/403/files#diff-c895426ce24017d1803054d94e4d71c3L69 is:
`crash_reason: {:bad_exit, []}`

https://github.com/getsentry/sentry-elixir/pull/403/files#diff-c895426ce24017d1803054d94e4d71c3L155:
`crash_reason: {{:timeout,
    {GenServer, :call, [#PID<0.738.0>, {:sleep, 20}, 1]}},
   [
     {GenServer, :call, 3, [file: 'lib/gen_server.ex', line: 1023]},
     {Task.Supervised, :invoke_mfa, 2,
      [file: 'lib/task/supervised.ex', line: 90]},
     {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 226]}
   ]}`

They eventually get passed into `Exception.normalize`.

josevalim: @mitchellhenke none of those are exceptions though. They are all exits. And I don't think we should treat them as exceptions. If you really want those reported, then I would recommend enabling `capture_log_message` by default and then maybe set the default level to `:error`?

test/logger_backend_test.exs
mitchellhenke: Done!

lib/sentry/logger_backend.ex
josevalim, line 115: You can try this instead:

```elixir
defmodule Sentry.CrashError do
  defstruct [:message]

  def exception({:nocatch, reason}) do
    %Sentry.CrashError{message: Exception.format_banner(:throw, reason, [])}
  end

  def exception(reason) do
    %Sentry.CrashError{message: Exception.format_banner(:exit, reason, [])}
  end
end
```

and then here you do:

```suggestion
      {%_{__exception__: true} = exception, stacktrace} when is_list(stacktrace) ->
        Sentry.capture_exception(exception, [stacktrace: stacktrace] ++ opts)

      {other, stacktrace} when is_list(stacktrace) ->
        Sentry.capture_exception(Sentry.CrashError.exception(other), [stacktrace: stacktrace] ++ opts)
```

PR #404, opened 2020-06-22
Do not rely on conditional module definition

josevalim:
The downside of conditional module definition is that
it may not work as expected in umbrellas. Given the
Sentry Logger configuration in an umbrella project is
shared across all apps, if we do:

    config :logger, :backends, [Sentry.LoggerBackend]

then `:sentry` needs to be defined on all dependencies.
However, if a dependency is not a web application,
Sentry may be compiled before Plug! And if it is compiled
before Plug, it wouldn't define Sentry.PlugCapture and
friends, which means the web app in the umbrella wouldn't
have those modules available.


mitchellhenke:
Thank you!!

Issue #413, opened 2020-07-14
Sentry getting error reports for non-500 errors from endpoint.ex


aselder:
### Environment

* Elixir version (elixir -v): 1.10.3
* Erlang/OTP version (erl): Erlang/OTP 23 [erts-11.0.2]
* Sentry version (mix deps): 7.2.4
* Operating system: Alpine Linux
* Is this an umbrella application?: yes
* Release type (Distillery, mix release, Mix, etc.): mix release

### Description
If an error is raised by a Plug in the endpoint file (outside the router), a report will be generated to Sentry even if that error does not result in a 500 Internal Server Error. For example, is Plug.Parser blows up due to garbage JSON, the response is properly sent back down the wire at 400 Bad Request

### Expected Behavior

No Sentry report will be generated because this is a handled error

### Actual Behavior

A Sentry report is generated

### Steps to Reproduce

1. Set up a Phoenix Endpoint with Sentry and a JSON endpoint.
2. Use curl to make a request with a garbage body.
3. Observe the request comes back with the proper 400 status code.
4. Find the error in Sentry.

aselder:
Probably as simple as adding 

      `Plug.Parsers.RequestTooLarge`,
      `Plug.Parsers.BadEncodingError`,
      `Plug.Parsers.ParseError`,
      `Plug.Parsers.UnsupportedMediaTypeError`

to the default list of excluded params. I'll whip up a PR.

mitchellhenke:
closed in #414

PR #414, opened 2020-07-14
Add plug parsing errors to list of default excluded params

aselder:
Phoenix handles these errors by default. In addition, these
errors are not really actionable by the application author, so
they just contribute noise to the Sentry exception list.

This addresses #413

### DIFF
--- a/lib/sentry/default_event_filter.ex
+++ b/lib/sentry/default_event_filter.ex
@@ -3,7 +3,15 @@ defmodule Sentry.DefaultEventFilter do
 
   @moduledoc false
 
-  def exclude_exception?(%x{}, :plug) when x in [Phoenix.Router.NoRouteError] do
+  @ignored_exceptions [
+    Phoenix.Router.NoRouteError,
+    Plug.Parsers.RequestTooLarge,
+    Plug.Parsers.BadEncodingError,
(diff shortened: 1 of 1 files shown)


lib/sentry/default_event_filter.ex
```diff
@@ -3,7 +3,15 @@ defmodule Sentry.DefaultEventFilter do
 
   @moduledoc false
 
-  def exclude_exception?(%x{}, :plug) when x in [Phoenix.Router.NoRouteError] do
+  @ignored_exceptions [
+    Phoenix.Router.NoRouteError,
+    Plug.Parsers.RequestTooLarge,
```
mitchellhenke, line 8: I think this should be `RequestTooLargeError`

aselder: Good catch!

aselder:
@mitchellhenke It's in https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/render_errors.ex. If a exception implements Plug.Exception behavior which these errors do (https://github.com/elixir-plug/plug/blob/master/lib/plug/parsers.ex), it will read the status of the status and send that as the HTTP status.

mitchellhenke:
Thanks! 💖💖💖

PR #418, opened 2020-07-20
Allow users to configure maximum number of breadcrumbs

### LINKED ISSUE #407: Allow configuration to set maximum number of breadcrumbs.
### Environment

* Sentry version (mix deps): `7.2.0`
* Operating system: macOS, Linux

### Description

Sometimes for long lived processes, too many breadcrumbs are added to a sentry event. When posting the event, the client fails if the body is too large. Similar to this issue: https://forum.sentry.io/t/http-error-413-request-entity-too-large/1942.

Is it possible to get configuration to limit the number of breadcrumbs that sentry will attempt to send? Something akin to `max_breadcrumbs` that is present in the python sentry library?

jfmyers9:
This configuration option allows a user to specify the maximum number of breadcrumbs when posting events to Sentry. The default value is 100.

This is helpful in situations in which long running processes can add a large number of breadcrumbs resulting in failures to create Sentry events when the message becomes too large.

Fixes #407


mitchellhenke:
Thank you!

PR #420, opened 2020-07-29
Kafai  allow setting all as metadata for backend config

bruteforcecat:
I branched off from 8.0.0 but it looks like some change from 8.0.0 is not merged to master HEAD yet so this PR also contains some of those change

This PR is to allow use set `metadata: :all` in config, same as the current behaviour for Logger.Backend.Console(https://hexdocs.pm/logger/Logger.html#module-console-backend).

Currently, I hard-coded list of key to be excluded when metadata set as all. I am not sure if you find this approach too dirty.
Another approach similar as Logger.console is to just exclude field that cannot be serialized in meaningful way like pid, func. But that will require changing more code in `Sentry.Client.encode_and_send`. So I decided to make a small PR first to see what you guys think.

Thank you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,19 @@
 # Changelog
 
+## 8.0.0 (2020-07-13)
+[https://gist.github.com/mitchellhenke/dce120a5515565076b13962ee0be749b](7.x -> 8.0 Upgrade Guide)
+
+* Bug Fixes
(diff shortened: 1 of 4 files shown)


mitchellhenke:
Please rebase off master 🙂

bruteforcecat:
just branched off from mater and  created new one https://github.com/getsentry/sentry-elixir/pull/421

PR #421, opened 2020-08-03
 allow setting all as metadata for Logger.Backend config

bruteforcecat:
I branched off from 8.0.0 but it looks like some change from 8.0.0 is not merged to master HEAD yet so this PR also contains some of those change

This PR is to allow use set metadata: :all in config, same as the current behaviour for Logger.Backend.Console(https://hexdocs.pm/logger/Logger.html#module-console-backend).

Currently, I hard-coded list of key to be excluded when metadata set as all. I am not sure if you find this approach too dirty.
Another approach similar as Logger.console is to just exclude field that cannot be serialized in meaningful way like pid, func. But that will require changing more code in Sentry.Client.encode_and_send. So I decided to make a small PR first to see what you guys think.

Thank you.

### DIFF
--- a/lib/sentry/logger_backend.ex
+++ b/lib/sentry/logger_backend.ex
@@ -21,7 +21,7 @@ defmodule Sentry.LoggerBackend do
   * `:metadata` - To include non-Sentry Logger metadata in reports, the
   `:metadata` key can be set to a list of keys. Metadata under those keys will
   be added in the `:extra` context under the `:logger_metadata` key. Defaults
-  to `[]`.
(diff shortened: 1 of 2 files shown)


mitchellhenke:
Thanks for rebasing!

I think this has a couple of issues that deserve further discussion. Using `:all` but having a hardcoded list that can't be included could be surprising. The values are unserializable with a default Jason encoder, but Sentry allows other encoders and users may have their own protocols implemented for encoding those types.

Exploring `:all` but with exclusions I think is more promising, but I'm not sure since it's not really `:all` then. Maybe there's a better name?

mitchellhenke:
Happy to continue discussion on this and reopen, but closing for now 🙂

PR #423, opened 2020-08-11
allow users to configure maximum number of breadcrumbs

mitchellhenke:
Backport of #418 to the 7.x release line


jeroenvisser101:
@mitchellhenke would it be better to discard/limit breadcrumbs in `add_breadcrumb`? That would also prevent memory exhaustion in long-running processes.

https://github.com/getsentry/sentry-elixir/blob/7b4295afb3b52cfaa2e7c7747a661040c3f4fde7/lib/sentry/context.ex#L253

mitchellhenke:
Yes, I think that would be much better!

PR #432, opened 2020-09-07
Update package & docs configuration

wojtekmach:
* Add CHANGELOG.md to the package so that it will be showed e.g.
    here: https://diff.hex.pm/diff/sentry/8.0.1..8.0.2

  * Add `@source_ref` & `@source_url` so that ExDoc will display links
    to the source code.


mitchellhenke:
Thank you!!

PR #437, opened 2020-10-06
Updates Sentry.Sources to support multiple source code root paths

jbernardo95:
Before the changes in this commit Sentry.Sources only looked for source
code files in one root path, the `:root_source_code_path`.

With the changes in this commit this module will now be able to look for
source code in multiple paths. The paths can be configured in the
`:root_source_code_paths` config key.

This change is specially important for umbrella applications [1], that
have their code nested in the `apps` path. If we don't specify the root
path for each individual application (which isn't possible without the
changes in this commit), and instead just use `File.cwd!()` as the root
path, the source code files of each umbrella app will be loaded with
`apps/<app name>` prefix in their name. This is a problem because these
names will never match the file names in the stacktrace, which means
that Sentry won't be able to attach source code context.

To avoid a configuration breaking change, the changes in this commit
still support the `:root_source_code_path` configuration key, but "soft
deprecate" it by removing it from documentation. This config key should
be fully deprecated in the next major Sentry release.

[1] https://elixir-lang.org/getting-started/mix-otp/dependencies-and-umbrella-projects.html

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -77,6 +77,17 @@ defmodule Sentry.Config do
     get_config(:enable_source_code_context, default: false, check_dsn: false)
   end
 
+  @deprecated "Use root_source_code_paths/0 instead"
+  def root_source_code_path do
+    path = get_config(:root_source_code_path)
+
+    if path do
+      path
+    else
+      raise ArgumentError.exception(":root_source_code_path must be configured")
+    end
+  end
+
   # :root_source_code_path (single path) was replaced by :root_source_code_paths (list of
   # paths).
   #
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -77,13 +77,27 @@ defmodule Sentry.Config do
     get_config(:enable_source_code_context, default: false, check_dsn: false)
   end
 
-  def root_source_code_path do
+  # :root_source_code_path (single path) was replaced by :root_source_code_paths (list of
+  # paths).
+  #
+  # In order for this to not be a breaking change we still accept the old
+  # :root_source_code_path as a fallback.
+  #
+  # We should deprecate this the :root_source_code_path completely in the next major
+  # release.
+  def root_source_code_paths do
+    paths = get_config(:root_source_code_paths)
     path = get_config(:root_source_code_path)
 
-    if path do
-      path
-    else
-      raise ArgumentError.exception(":root_source_code_path must be configured")
+    cond do
+      not is_nil(paths) ->
+        paths
+
+      not is_nil(path) ->
+        [path]
+
+      true ->
+        raise ArgumentError.exception(":root_source_code_paths must be configured")
     end
   end
 
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -90,6 +90,14 @@ defmodule Sentry.Config do
     path = get_config(:root_source_code_path)
 
     cond do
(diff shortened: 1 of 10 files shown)


jbernardo95:
@mitchellhenke would appreciate a review when you have the chance 😊

mitchellhenke:
Thanks for opening a PR!

The sticky issue that's prevented similar solutions is how to handle collisions. 

`apps/app_a/lib/module_a.ex` and `apps/app_b/lib/module_a.ex` would both resolve to `lib/module_a.ex`, and could lead to some surprising results in the Sentry interface. Raising an error is one option.

In terms of the option deprecation, what do you think of raising an error on configuring both, and potentially emitting a warning for the deprecated option?

Also as a reminder to myself that should this get merged, the documentation on the Sentry site will need updating as well 🙂

jbernardo95:
> The sticky issue that's prevented similar solutions is how to handle collisions.
>
> apps/app_a/lib/module_a.ex and apps/app_b/lib/module_a.ex would both resolve to lib/module_a.ex, and could lead to some surprising results in the Sentry interface. Raising an error is one option.

Very good point 🤔 I think that this should very rare, so raising an error if this happens makes sense to me. We can suggest users to rename files if this is the case.

> In terms of the option deprecation, what do you think of raising an error on configuring both, and potentially emitting a warning for the deprecated option?

100% agree!

jbernardo95:
@mitchellhenke pushed a new commit (that I'll squash before merging) addressing your suggestions. Let me know what you think!

jbernardo95:
@mitchellhenke quick reminder on this.

lib/sentry/config.ex
```diff
@@ -77,13 +77,35 @@ defmodule Sentry.Config do
     get_config(:enable_source_code_context, default: false, check_dsn: false)
   end
 
-  def root_source_code_path do
```
mitchellhenke, line 80: For the deprecation strategy, I think it would be preferable to keep the old method with the same behavior, but mark it as `@deprecated`?

I'm guessing its usage is rare, but I'd rather not remove it until the next major release.

mitchellhenke (commented):
Apologies for the delay, I've had some more time to think about this and left a couple comments.

I'm happy to merge and pick this up if you're unable to work on the comments I've added.

Thanks again for the work to improve Sentry 🙂

lib/sentry/config.ex
jbernardo95: @mitchellhenke do you mean keeping both the new and the old versions? See if the changes in https://github.com/getsentry/sentry-elixir/pull/437/commits/52d32cfb787d0a94c15228450f407917c57e1335 is what you had in mind.

mitchellhenke: Yep, exactly!

mitchellhenke:
@jbernardo95 thank you!! 💖 💖

jbernardo95:
Thank you @mitchellhenke! ❤️ 

(btw I hadn't rebased my commits, so a few `!fixup` commits were merged to master)

jbernardo95:
@mitchellhenke btw do you have any plans to release v8.0.3 soon?

jbernardo95:
Awesome! Thank you 🙌

jbernardo95:
@mitchellhenke I just realised that the README is now out of date [here](https://github.com/getsentry/sentry-elixir#umbrella-applications).

mitchellhenke:
@jbernardo95 whoops, thank you! Removed in dac9c3e

PR #441, opened 2020-11-16
Not reading dsn config at compile time

### LINKED ISSUE #440: Version 8.0.3 fails due to configuration evaluated at compile time
### Environment

* Elixir version (elixir -v): 1.11.1
* Erlang/OTP version (erl): 23
* Sentry version (mix deps): 8.0.3
* Operating system: alpine (Docker)
* Is this an umbrella application?: No
* Release type (Distillery, mix release, Mix, etc.): Distillery

### Description

After upgrading to 8.0.3, running `mix deps.compile` fails due to evaluating config values at compile time rather than runtime. We are using https://github.com/rockneurotiko/config_tuples to read configuration from the environment in order to ship the same docker containers in staging and prod (and only have different values inside the env variables). See example below on how this looks for the sentry dns url. 

### Expected Behavior

`deps.compile` succeeds even if values in config are not available or different than expected at compile time.

### Actual Behavior

Configuration is read at compile time and compile fails with 

```
(FunctionClauseError) no function clause matching in URI.parse/1
The following arguments were given to URI.parse/1:
        # 1
        {:system, "SENTRY_DNS", [required: true]}
```

### Steps to Reproduce

config.exs:

`config :sentry,
  dsn: {:system, "SENTRY_DNS", required: true}`

Run `deps.compile`

Sgoettschkes:
This PR attempts to fix #440.

As described in the bug report, the `:dsn` config value got read at compile time, failing if something else then a value that can be parsed by `URI.parse/1` is present.

I added a test as well but I'm not 100% happy as it does not confirm that `Sentry.config.dsn/0` is not called at compile time but rather that a call to `root_source_code_paths` does not evaluate the `dsn` config variable. Calling `get_config` without the `check_dsn: false` option at compile time somewhere else will result in the same error and is not catched by my test. I'm currently not sure on how to test this.

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -79,7 +79,7 @@ defmodule Sentry.Config do
 
   @deprecated "Use root_source_code_paths/0 instead"
   def root_source_code_path do
-    path = get_config(:root_source_code_path)
+    path = get_config(:root_source_code_path, check_dsn: false)
 
     if path do
       path
@@ -97,8 +97,8 @@ defmodule Sentry.Config do
   # We should deprecate this the :root_source_code_path completely in the next major
   # release.
   def root_source_code_paths do
-    paths = get_config(:root_source_code_paths)
-    path = get_config(:root_source_code_path)
+    paths = get_config(:root_source_code_paths, check_dsn: false)
+    path = get_config(:root_source_code_path, check_dsn: false)
 
     cond do
       not is_nil(path) and not is_nil(paths) ->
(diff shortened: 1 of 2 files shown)


mitchellhenke:
Thank you so much for catching this and sending a fix!

I will look further into the case you mentioned, but I think releasing this fix is more urgent 🙂

Sgoettschkes:
Thanks for the fast response! I'm glad I could contribute back!

PR #442, opened 2020-11-22
Adds support for lists in scrubbing logic

joladev:
The scrubbing logic in `Sentry.PlugContext` acts on keys, and so lists themselves are not a target for scrubbing. But lists can contain maps that need scrubbing. This adds support for lists in the scrubbing logic.

### DIFF
--- a/lib/sentry/plug_context.ex
+++ b/lib/sentry/plug_context.ex
@@ -171,7 +171,7 @@ defmodule Sentry.PlugContext do
   end
 
   @doc """
-  Recursively scrubs a map that may have nested maps
+  Recursively scrubs a map that may have nested maps or lists
 
   Accepts a list of keys to scrub, and a list of options to configure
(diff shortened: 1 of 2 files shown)


lib/sentry/plug_context.ex
```diff
@@ -206,11 +206,35 @@ defmodule Sentry.PlugContext do
...
+
           true ->
             value
         end
 
       {key, value}
     end)
   end
+
+  @spec scrub_map(list(), list(String.t()), keyword()) :: list()
```
mitchellhenke, line 220:
```suggestion
  @spec scrub_list(list(), list(String.t()), keyword()) :: list()
```

mitchellhenke (changes requested):
This is great, thank you!

mitchellhenke:
Thank you!

PR #447, opened 2020-12-14
Send Sentry reports on uncaught throws/exits

### LINKED ISSUE #446: Sentry not reporting on processes dying on uncaught `exit()` or `throw()`
### Environment

* Elixir version (elixir -v): 1.11.2
* Erlang/OTP version (erl): 23.0.4
* Sentry version (mix deps): 8.0.4
* Operating system: OS X
* Is this an umbrella application?: yes
* Release type (Distillery, mix release, Mix, etc.): release

### Description

Inside of `call`, Sentry.PlugCapture only rescues from it's try block, it does not `catch` in addition to `rescue`. This means is a request dies because of a uncaught `exit/1` or `throw/1`, this failure is not picked up by Sentry and the failure is never captured.

See https://github.com/phoenixframework/phoenix/blob/6f87ef24d02e6afc2634b4c42b20b1b5d89aebe5/lib/phoenix/endpoint.ex#L614-L631 for an example where Phoenix has very similar code in both a `rescue` block as well as a `catch` block.

### Expected Behavior

If an request dies because of an uncaught `exit/1` or `throw/1`, this error will be reported to Sentry.

### Actual Behavior

Sentry does not capture this event.

### Steps to Reproduce

1. Create a phoenix app with sentry logging.
2. Have an endpoint that triggers an exit, either call `exit/1` directly or do something like `GenServer.call(:some_non_existant_name, :foo, [])`
3. Phoenix will return a 500 for the request, but nothing will be captured in Sentry

See https://github.com/phoenixframework/phoenix/issues/4099 which is related.

aselder:
Fix #446

Previously PlugCapture would only catch exceptions. If there was an uncaught exit or throw, this would be ignored and Plug/Phoenix would end up send a 500 Internal Server Error to the clients with no notification to Sentry.

### DIFF
--- a/lib/sentry/plug_capture.ex
+++ b/lib/sentry/plug_capture.ex
@@ -45,6 +45,12 @@ defmodule Sentry.PlugCapture do
           e ->
             _ = Sentry.capture_exception(e, stacktrace: __STACKTRACE__, event_source: :plug)
             :erlang.raise(:error, e, __STACKTRACE__)
+        catch
+          kind, reason ->
+            message = "Uncaught #{kind} - #{inspect(reason)}"
+            stack = __STACKTRACE__
+            Sentry.capture_message(message, stacktrace: stack, event_source: :plug)
+            :erlang.raise(kind, reason, stack)
         end
       end
     end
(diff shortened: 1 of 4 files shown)


mitchellhenke:
Thank you so much, this is perfect! 💖

cheerfulstoic:
Can this be released?  I just noticed this issue and I see that the latest `sentry` package is 8.0.4 from Nov 16th

mitchellhenke:
@cheerfulstoic yep, thank you for the reminder and sorry for the delay. I've just pushed version 8.0.5 to hex.

cheerfulstoic:
🎉  ❤️  Thanks!

PR #448, opened 2021-01-04
Misc doc changes

kianmeng:
List of changes:
- Fix typo
- Add changelog to html doc
- Add Changelog link to hex info page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 8.0.4 (2020-11-16)
 
-* Do not read dsn config at compile time (#441)
+* Do not read DSN config at compile time (#441)
 
 ## 8.0.3 (2020-11-11)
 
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014 Stanislav Vishnevskiy
+Copyright (c) 2014-present, Stanislav Vishnevskiy
```
mitchellhenke, line 3: I'd have to double check on this, could you remove the change for now?

mitchellhenke (changes requested):
One small change, otherwise looks great

Thank you so much!

LICENSE
kianmeng: > I'd have to double check on this, could you remove the change for now?

Done.

kianmeng: @mitchellhenke It seems that the use of copyright year `2014-present` is not recommended, see https://github.com/riverrun/comeonin/pull/153#issuecomment-757405132.

Do let me know if you found something on this as well. Thanks. :smile:

mitchellhenke: Thanks for the reference, everything looks great now 🙂

mitchellhenke (approved):
Thank you so much for submitting these improvements!

PR #450, opened 2021-01-06
replace whitelist with allow list

### LINKED ISSUE #449: Review `in_app_whitelist`
Our developer guidelines (https://develop.sentry.dev/inclusion/#changing-code) identify terms we'd like to phase out to ensure our code uses inclusive language.

Specifically, for `in_app_whitelist`:
"**whitelist/blacklist**:
Generally should not be used. We try to avoid the word "list" now in general and try to be more descriptive in the UI. For instance, the use of "ip blacklist" has been changed to "ip filters" as part of the filter product. If you need a generic term "allowlist" and "denylist" or "blocklist" are good substitutions."

Can we phase this term out when we next make a breaking change?

mitchellhenke:
Closes #449

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## next release
+* Deprecate `Sentry.Config.in_app_module_whitelist` in favor of `Sentry.Config.in_app_module_allow_list` (#450)
+
 ## 8.0.4 (2020-11-16)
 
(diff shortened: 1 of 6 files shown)


PeloWriter:
Thank you!

Miradorn:
heads up: I think this change did not translate to your developer docs [here](https://docs.sentry.io/platforms/elixir/configuration/options/), which I jsut found out the hard way :D

PR #462, opened 2021-02-07
Default release to Elixir release "RELEASE_VSN"

### LINKED ISSUE #361: Enhancement: Grab release version from env
### Environment

* Elixir version (elixir -v): 1.9.1
* Erlang/OTP version (erl): 22.0
* Sentry version (mix deps): 7+
* Operating system: Linux

### Description

This is just an enhancement tip, in 1.9.0+ (the one with mix release) you can grab the application version via: `System.get_env("RELEASE_VSN")`

Would be cool if it default to this when the release is nil/not in the config.

mitchellhenke:
Closes #361 

Moves some of the config logic that was previously in the application module into the config module.

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -101,8 +101,7 @@ defmodule Sentry do
       Config.client().child_spec()
     ]
 
-    validate_json_config!()
-    validate_log_level_config!()
+    Sentry.Config.set_up_and_validate_config!()
 
     opts = [strategy: :one_for_one, name: Sentry.Supervisor]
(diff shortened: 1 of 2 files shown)


yordis (changes requested):
Kind of related to https://github.com/getsentry/sentry-elixir/issues/503

I don't think that keep adding more implicit environment variables is the way to go.

Adding example code, or whatnot could be a good addon, but such technicalities should be left alone to the Operations + App developers.

mitchellhenke:
I agree, thank you!

PR #467, opened 2021-03-19
Remove function that disabled non-group leader logging

### LINKED ISSUE #466: No logs resulting from mix release RPC
### Environment

How do you use Sentry?
Sentry SaaS

Which SDK and version?
Sentry 8.0

### Steps to Reproduce

1. Configure sentry logger on an elixir Release
2. Use [Elixir release RPC](https://hexdocs.pm/mix/master/Mix.Tasks.Release.html#module-one-off-commands-eval-and-rpc) to send an RPC request that'll result in an error.
3. Observe the Sentry logger will not have reported the error to sentry.

### Expected Result

An RPC to a running elixir node should result in a sentry error if applicable.

### Actual Result

Due to this line: https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/logger_backend.ex#L71 The logger is not the group leader and thus does not send a log when there's an error.

Same bug as in LoggerJSON: https://github.com/Nebo15/logger_json/issues/65

This bug comes from a bad implementation of Elixir's logger a very long time ago. When mix releases came out Elixir changed this so you'd actually get logs from RPCs but most libs that copied that code didn't change the offending function.

SophisticaSean:
fixes #466


mitchellhenke:
Thanks for fixing this! I’m looking into the CI failure, looks like GitHub actions is having some issues unrelated to the changes.

SophisticaSean:
> Thanks for fixing this! I’m looking into the CI failure, looks like GitHub actions is having some issues unrelated to the changes.

No problem! Let me know if there's anything I can do to help or anything you'd like changed.

mitchellhenke:
Sorry for the hassle, but would you be able to rebase? There were some CI issues I fixed in #469 that will hopefully get the tests running for your PR.

SophisticaSean:
@mitchellhenke done! thank you :)

mitchellhenke:
Thank you for fixing this!

PR #470, opened 2021-04-07
Proposal: Add before_capture_exception hook

thdxr:
The existing `before_send_event` hook is useful however the error has already been processed somewhat at that point.  This means there's sensitive data that could be embedded in the error string that is hard to detect and redact.  We recently had a situation where some private keys were pushed into sentry and alerted into Slack causing the need for a key rotation.

This PR has an example implementation of `before_capture_exception` which allows us to hook into things early and redact the raw error before it is processed.  If this is something that can be added to the main library I can wrap this up with documentation/tests

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -121,8 +121,28 @@ defmodule Sentry do
       :excluded
     else
       exception
+      |> maybe_call_before_capture_exception()
       |> Event.transform_exception(opts)
       |> send_event(opts)
+    end
+  end
+
+  @spec maybe_call_before_capture_exception(Exception.t()) :: Exception.t()
+  def maybe_call_before_capture_exception(exception) do
+    case Config.before_capture_exception() do
+      function when is_function(function, 1) ->
+        function.(exception) || false
+
+      {module, function} ->
+        apply(module, function, [exception]) || false
+
+      nil ->
+        exception
+
+      _ ->
+        raise ArgumentError,
+          message:
+            ":before_capture_exception must be an anonymous function or a {Module, Function} tuple"
     end
   end
 
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -182,6 +182,10 @@ defmodule Sentry.Config do
     get_config(:hackney_opts, default: [], check_dsn: false)
   end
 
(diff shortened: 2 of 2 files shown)


mitchellhenke:
Thanks for opening a PR (and sorry to hear about going through the hassle of key rotation). I'm curious if the `before_send_event` may still be an option, though I'm open to something like this too.

The original exception is available on the event, and could be used to modify the event. There's a simplified example showing how it could be used in the tests: 

https://github.com/getsentry/sentry-elixir/blob/0b7f8e3d786d626dbf945c421a4850ce7dbcde51/test/support/test_before_send_event.exs#L7-L15

Sentry uses [Exception.message/1](https://hexdocs.pm/elixir/Exception.html#message/1), so a custom exception that changes what `message` includes is a potential strategy.

Relevant code if it's helpful:
https://github.com/getsentry/sentry-elixir/blob/0b7f8e3d786d626dbf945c421a4850ce7dbcde51/lib/sentry/event.ex#L187-L190

Would you be able to explore if the above would help address your issue?

thdxr:
That was actually what I tried initially:
```
 def before_send(event) do
    event.original_exception
    |> Redact.redact()
    |> Sentry.Event.transform_exception([])
  end
```

However this loses the stack trace / other opts since I don't have the original opts passed into `capture_exception` here.  I guess I could copy those fields back in but wasn't sure which fields needed to be copied back

mitchellhenke:
It is technically feasible to redact the fields that are derived from `Exception.message`, e.g.:

```elixir
def before_send(event) do
  new_message = redact_message(event.original_exception)
  new_event_exception = redact_exception(event.original_exception)
  Map.put(event, :message, new_message)
  |> Map.put(:exception, [new_event_exception])
end
```

I suspect that approach is kind of whack-a-mole and could tend to be more brittle. If possible, preventing Sentry from handling any sensitive information at the boundary may be easier. Custom exceptions that define `message` or structs that redact sensitive information from the outset may be options, and they'd also cover information being leaked in logs, console sessions, etc. I'm not familiar with your issue in particular, but deriving specific attributes for [Inspect](https://hexdocs.pm/elixir/Inspect.html#module-deriving) is one strategy.

PR #472, opened 2021-04-07
Use Logger.add_backend/1

josevalim:
Configuring the backend on config/config.exs may be too early
and cause issues if trying to loading logger config before Sentry
is compiled.

Or it can be an issue in umbrella projects when only some of them
depend on Sentry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,13 +93,13 @@ defmodule MyAppWeb.ErrorView do
 
 ### Capture Crashed Process Exceptions
 
-This library comes with an extension to capture all error messages that the Plug handler might not.  This is based on [Logger.Backend](https://hexdocs.pm/logger/Logger.html#module-backends).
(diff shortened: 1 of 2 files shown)


RodolfoSilva (approved):

josevalim:
Note the previous snippet also wouldn't work for releases. :)

thbar:
> it can be an issue in umbrella projects when only some of them
depend on Sentry.

Thanks @josevalim - it is one of the pain points I've had in the umbrella app I'm managing, this will help greatly.

mitchellhenke:
@josevalim Thank you so much!! 💖

josevalim:
Oh, I am so sorry, I thought you were referring to Elixir's getting started guide. My bad, thanks for the PR though!

mitchellhenke:
@morgz thanks for updating the docs!

PR #473, opened 2021-04-12
Update "User Feedback" URL

jcrevits:
Hello! Tiny fix but the current link to `User Feedback` leads to a "Page not found". I believe this is the new URL?


mitchellhenke:
Thank you!

Issue #479, opened 2021-05-25
Attachments


hfjallemark:
Are there any plans to support attachments like some of the other libraries does?

https://docs.sentry.io/platforms/javascript/enriching-events/attachments/
https://docs.sentry.io/platforms/native/enriching-events/attachments/
https://docs.sentry.io/platforms/java/enriching-events/attachments/

Looks like it's a separate endpoint for uploading attachments: https://develop.sentry.dev/sdk/envelopes/

mitchellhenke:
@hfjallemark thanks for opening an issue! This is related to #426 as it's part of the rewrite to enable the use of the Envelope API

Issue #426, opened 2020-09-02
Support performance/APM features


anthonator:
Does this project plan to add support for Sentry's performance/APM features? Elixir's/Erlang's telemetry support seems like a perfect fit for this.

mitchellhenke:
It’s something I’ve definitely been keeping an eye on. Now that it’s released, I’ll be trying to get some proofs of concept together.

anthonator:
Is there anything the community can do to help?

mitchellhenke:
Good question, thanks for asking it!

Once I have a better handle on the structure, I’ll create issues around the features which should be a great place for contributions 🙂

mitchellhenke:
It turns out that before this can happen I will need to make some bigger changes to move existing functionality to the new Unified API and Envelope data format, so I'll be tackling that first.

anthonator:
I wonder if a good MVP would be just to have the ability to send metrics to Sentry. We can manually hook up telemetry since Phoenix, Ecto, etc. have good telemetry support.

UnderFenex:
Waiting for this feature :)

mitchellhenke:
It's an exciting feature that's at the top of my list for support!

Ruby's conversion to the Unified API and Envelope format gives a bit of an insight into what needs to happen first: https://github.com/getsentry/sentry-ruby/projects/3

This library's footprint is smaller, so it should be easier in some respects 🙂

hauleth:
Is there API specification for that? I could review the possibility to implement such feature if only there is stable API for that. I could make it a backend for Erlang's OpenTelemetry implementation.

mitchellhenke:
Apologies this is taking longer than anticipated, but I'm hoping that the holidays will give me some time to finish the partial implementation I have.

@hauleth It's sort of spread across a handful of pages: https://develop.sentry.dev/sdk/event-payloads/transaction/, https://develop.sentry.dev/sdk/event-payloads/span/ have some description, but I'd be very interested in making it easier to integrate with OpenTelemetry.

speeddragon:
Is the APM feature available for free accounts? I don't mind to help to implement this feature, but I was trying to check the API to send an envelope request and I'm always getting 429.

mitchellhenke:
@speeddragon I'm not sure, I'd reach out to Sentry to see.

I've been making slow progress, but not as much as I'd like 🙂

speeddragon:
[...]

PR #490, opened 2021-10-09
Add Envelope Support

JohnDoneth:
This PR implements the newer envelope format https://develop.sentry.dev/sdk/envelopes/#data-model for events only.

This is a stepping stone and a blocker to implementing transactions, attachments, and other features that are requested in #426, and #479.

### DIFF
--- a/lib/sentry/envelope.ex
+++ b/lib/sentry/envelope.ex
@@ -88,9 +88,11 @@ defmodule Sentry.Envelope do
   @spec event(t()) :: Event.t() | nil
   def event(envelope) do
     envelope.items
-    |> Enum.filter(fn item -> is_struct(item, Event) end)
+    |> Enum.filter(fn item -> is_event?(item) end)
     |> List.first()
   end
+
+  defp is_event?(event), do: match?(%{__struct__: Sentry.Event}, event)
 
   #
   # Encoding


mitchellhenke (commented):
This is incredible, thank you!

I will review it this week 🙂

mitchellhenke (approved):
@JohnDoneth The implementation is great, I've been running this branch for a bit and haven't run into any issues. Again, thank you!

lib/sentry/envelope.ex
```diff
@@ -0,0 +1,186 @@
...
+  def from_binary!(binary) do
+    {:ok, envelope} = from_binary(binary)
+    envelope
+  end
+
+  # Returns the event in the envelope if one exists.
+  @spec event(t()) :: Event.t() | nil
+  def event(envelope) do
+    envelope.items
+    |> Enum.filter(fn item -> is_struct(item, Event) end)
```
mitchellhenke, line 91: Ah, I had forgotten that `is_struct/2` isn't available in `1.10.x`. Could you switch this to something like `match?(%Sentry.Event{}, %{__struct__: Sentry.Event})` ?

JohnDoneth: Replaced in 4f8603c

syfgkjasdkn:
Hi @mitchellhenke! Does it mean that `Sentry v9.1.2` would no longer be supported? Since some users are stuck on 9.1.2 due to infrastructure complexity of the newer versions, would it be possible to keep the old API?

mitchellhenke:
@syfgkjasdkn That's a good question, the change here is probably sufficient to be considered breaking. Maintenance of older major versions has typically been limited, though not dropped entirely.

PR #495, opened 2022-02-10
Add os and runtime info to events

### LINKED ISSUE #496: Send OS and runtime information to Sentry
Other language SDK supports sending the OS and runtime contexts (such as name and version) as contexts metadata.
(For example, in ruby SDK: https://github.com/getsentry/sentry-ruby/blob/2270fd6dd822fecbe3ebedd4c5df9be319e9939c/sentry-ruby/lib/sentry/scope.rb#L244 )

These information will be displayed as TAGS section, that is useful to distinguish errors that caused by specific version of runtimes.
(Example with ruby SDK:
貼り付けた画像_2022_02_10_13_43
)

sekiyama58:
This collects the OS and runtime's name and version from :os and System module,
and send them as contexts metadata to Sentry. 

These information will be displayed in the TAGS section with the corresponding icons:
tags

Closes #496

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -333,6 +333,7 @@ defmodule Sentry.Client do
       user: event.user,
       breadcrumbs: event.breadcrumbs,
       fingerprint: event.fingerprint,
+      contexts: event.contexts,
       modules: event.modules
     }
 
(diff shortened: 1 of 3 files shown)


mitchellhenke:
Thanks for opening an issue with a PR to implement it, this looks great!

PR #500, opened 2022-03-14
Use X-Forwarded-For header (if defined) to extract real client IP

davidjulien:
See https://en.wikipedia.org/wiki/X-Forwarded-For

Client IP is not correct when server is behind a proxy or a load balancer. Real client IP can be identified thanks to X-Forwarded-For header.

### DIFF
--- a/lib/sentry/plug_context.ex
+++ b/lib/sentry/plug_context.ex
@@ -123,8 +123,8 @@ defmodule Sentry.PlugContext do
       cookies: handle_data(conn, cookie_scrubber),
       headers: handle_data(conn, header_scrubber),
       env: %{
-        "REMOTE_ADDR" => remote_address(conn.remote_ip),
-        "REMOTE_PORT" => Plug.Conn.get_peer_data(conn).port,
+        "REMOTE_ADDR" => remote_address(conn),
+        "REMOTE_PORT" => remote_port(conn),
         "SERVER_NAME" => conn.host,
         "SERVER_PORT" => conn.port,
         "REQUEST_ID" => Plug.Conn.get_resp_header(conn, request_id) |> List.first()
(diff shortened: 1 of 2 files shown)


test/plug_context_test.exs
```diff
@@ -38,6 +42,31 @@ defmodule Sentry.PlugContextTest do
            } = Sentry.Context.get_all()
   end
 
+  test "sets request context with real client ip if request is forwarded" do
+    Sentry.PlugContext.call(
+      conn(:get, "/test?hello=world") |> add_x_forwarded_for("10.0.0.1"),
```
speeddragon, line 47: Does it make sense to add a test with more than one IP address?

lib/sentry/plug_context.ex
```diff
@@ -123,24 +123,46 @@ defmodule Sentry.PlugContext do
...
   end
 
-  defp remote_address(address) do
-    address
-    |> :inet.ntoa()
-    |> case do
-      {:error, _} ->
-        ""
+  defp remote_address(conn) do
+    if header_value = get_header(conn, "x-forwarded-for") do
```
speeddragon, line 136: Should the module documentation mention this behaviour for people that have other headers (for this purpose) to know which header to use?

speeddragon:
Hi David, thanks for this PR. I'm not a contributor for this project, but I just did the same as you and only notice now that you had a PR open. Hope this can be merged soon since it was implemented in [another sentry client](https://github.com/getsentry/sentry-java/commit/c5f9e2d4f0d2fca8b6d54b520c8aa997bbef1fd7) way back.

speeddragon (approved):

mitchellhenke:
Thank you for this fix!

test/plug_context_test.exs
davidjulien: Yes, it will be more explicit.

You may have multiple values for this header. In that case, the first one is the client address and the next ones the succession of proxies until the final server.

`X-Forwarded-For: <client>, <proxy1>, <proxy2>`

lib/sentry/plug_context.ex
davidjulien: I found`X-Originating-IP`, `X-Remote-IP`, `X-Remote-Addr` for the same usage. However I don't known if some proxies still use these headers and the expected behavior when you have several different headers.

PR #504, opened 2022-05-02
Use Supervisor.child_spec for type

jsmestad:
Use the typespec from Elixir, [Supervisor.child_spec/0](https://hexdocs.pm/elixir/Supervisor.html#t:child_spec/0) instead of the Erlang module type.

----

By submitting this pull request, I confirm that Sentry can use, modify, copy, and redistribute this contribution, under Sentry's choice of terms.

### DIFF
--- a/lib/sentry/http_client.ex
+++ b/lib/sentry/http_client.ex
@@ -13,7 +13,7 @@ defmodule Sentry.HTTPClient do
 
   @type headers :: [{String.t(), String.t()}]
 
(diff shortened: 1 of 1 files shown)


yordis (approved):

sl0thentr0py:
dialzyer throws an error on this since `hackney_pool` implements `child_spec` as a tuple and not the map, closing since this is very old either way.

PR #508, opened 2022-06-09
Traverse full domain list when checking for excluded domains

martosaur:
As per documentation
> `:excluded_domains` - Any messages with a domain in the configured list will not be sent

However, when checking log message domains `excluded_domain?/2` function only checks the head element. 

This makes it impossible to exclude messages logged with default Elixir logger interface like `Debug.error("Message", domain: [:custom_domain])` as [their domain are always prepended with `:elixir` domain](https://github.com/elixir-lang/elixir/blob/7e4fbe657dbf9c3e19e3d2bd6c17cc6d724b4710/lib/logger/lib/logger.ex#L827-L829)


sl0thentr0py (approved):
ty @martosaur very late merge but better late than never

sonic182:
@michallepicki could you?

PR #509, opened 2022-06-10
Update default value of included_environments in README

andreyuhai:
Hi 👋 

I've seen that the default value of `included_environments` is incorrect in README.
That might confuse people, as it confused me. So I wanted to fix it. :D

---

Or was it actually supposed to include all the envs mentioned in the README? 🤔
I can create a PR for that as well, if that's the way to go.
In my opinion it would be better so people wouldn't have to mention all those separately in their configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ The full range of options is the following:
 | ------------- | -----------------|--------------|-------|
 | `dsn` | True  | n/a | |
 | `environment_name` | False  | :prod | |
-| `included_environments` | False  | `[:test, :dev, :prod]` | If you need non-standard mix env names you *need* to include it here |
+| `included_environments` | False  | `[:prod]` | If you need non-standard mix env names you *need* to include it here |
 | `tags` | False  | `%{}` | |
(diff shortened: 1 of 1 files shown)


mitchellhenke:
It matches the default config [here](https://github.com/getsentry/sentry-elixir/blob/master/config/config.exs#L4)?

Could you share a bit more on what made it confusing?

andreyuhai:
Sorry, I should've described it in more detail.

If I don't define `included_environments` in my `config.exs` explicitly and run the app, it just defaults to `[:prod]` which comes from [config.ex](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/config.ex#L8)  I think.

I was expecting that to default to `[:prod, :test, :dev]` even me without explicitly mentioning those.

To reproduce,
1. Create a new Elixir project `mix new foobar`
2. Add Sentry and other dependencies
3. Run `Sentry.Config.included_environments` in `iex` shell.

which gives you

```
iex(3)> Sentry.Config.included_environments
[:prod]
```

PR #515, opened 2022-07-27
Wrap HTTP requests in try/catch

### LINKED ISSUE #514: Possibly infinite error loop with custom http client.
👋

The way people use finch for a sentry client [[1,](https://github.com/search?q=Sentry.FinchClient+language%3AElixir&type=Code) [2](https://claudio-ortolina.org/posts/using-finch-with-sentry/)] can cause sentry to enter a possibly infinite exception generating loop.

Once `:sentry` starts sending enough events, `:nimble_pool` which `:finch` uses for managing http1 connections starts [timing out and exiting](https://github.com/dashbitco/nimble_pool/blob/f495e4d279cb5d8bd8c2dc56c33d37cdcdd1e96d/lib/nimble_pool.ex#L368) and finch [relays the exit](https://github.com/sneako/finch/blob/702c52767e5f3dd1fb3560057a85b5e62b2ca357/lib/finch/http1/pool.ex#L57) to the caller which makes [the task sending the request](https://github.com/getsentry/sentry-elixir/blob/76b23dbaff6ec560dc806a4b107b2657a772b55a/lib/sentry/client.ex#L138-L141) terminate and generate another exception thus entering an error loop generating more exceptions.

The exeption looks like this

```plain
12:12:12.866 [error] Task #PID<0.391.0> started from #PID<0.288.0> terminating
** (stop) exited in: NimblePool.checkout(#PID<0.347.0>)
    ** (EXIT) time out
    (finch 0.12.0) lib/finch/http1/pool.ex:57: Finch.HTTP1.Pool.request/5
    (finch 0.12.0) lib/finch.ex:313: anonymous fn/4 in Finch.request/3
    (telemetry 1.1.0) /Users/ruslan/Developer/sentry-finch-issue/deps/telemetry/src/telemetry.erl:320: :telemetry.span/3
    (s 0.1.0) lib/finch_client.ex:18: Sentry.FinchClient.post/3
    (sentry 8.0.6) lib/sentry/client.ex:210: Sentry.Client.request/3
    (sentry 8.0.6) lib/sentry/client.ex:192: Sentry.Client.try_request/5
    (sentry 8.0.6) lib/sentry/client.ex:167: anonymous fn/4 in Sentry.Client.do_send_event/3
    (elixir 1.13.4) lib/task/supervised.ex:89: Task.Supervised.invoke_mfa/2
Function: #Function<2.43295134/0 in Sentry.Client.do_send_event/3>
    Args: []
```

Please see [this gist](https://gist.github.com/ruslandoga/9bb8c0a27f8ffd7661d34b8957c9d9ac) for more info, it also contains two possible workarounds (`try/catch` and `protocol: :http2`) but I don't know if they are good enough to suggest to other people.

### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

### Steps to Reproduce

```bash
export SENTRY_DSN=https://1a77dc7a687646b28c1b1b6cee1ef4a9@o1012425.ingest.sentry.io/6601933
curl -LO https://gist.githubusercontent.com/ruslandoga/9bb8c0a27f8ffd7661d34b8957c9d9ac/raw/c0eb3a448c0493f08bef38286a92436c56333308/error_loop.exs
elixir error_loop.exs
```

### Expected Result

Sentry doesn't enter that kind of an error loop regardless of what the client does. Or docs make it clear that custom clients shouldn't raise.

### Expected Result (from opening this issue)

Not sure. One one hand, this is more of a problem with the custom client that `:sentry` is not responsible for, on the other hand, I haven't been able to find a client that foresaw this problem which could mean that docs regarding custom clients are not clear enough.

Maybe it would be ok to include a finch client by default or add it in the docs for `Sentry.HTTPClient` as an example?

ruslandoga:
closes #514


lib/sentry/client.ex
```diff
@@ -218,6 +218,10 @@ defmodule Sentry.Client do
       e ->
         {:error, e}
     end
+  catch
+    :exit, data -> {:error, {:exit, data, __STACKTRACE__}}
+    :throw, data -> {:error, {:throw, data, __STACKTRACE__}}
+    :error, data -> {:error, {:error, data, __STACKTRACE__}}
```
ruslandoga, line 224: It'd be easier to log directly instead of reassembling the error in `maybe_log_result/1`.

test/client_test.exs
```diff
@@ -353,4 +354,56 @@ defmodule Sentry.ClientTest do
...
+        client: Sentry.MockClient
+      )
+
+      :ok
+    end
+
+    test "handles exits" do
+      expect(Sentry.MockClient, :post, fn _url, _headers, _body -> exit(:timeout) end)
+
+      try do
```
ruslandoga, line 373: these tests would need `capture_log` after https://github.com/getsentry/sentry-elixir/pull/517

lib/sentry/client.ex
mitchellhenke: I'd be happy to see what that would look like!

lib/sentry/client.ex
```diff
@@ -115,107 +115,116 @@ defmodule Sentry.Client do
...
-        |> maybe_call_after_send_event(event)
+              retry_request(endpoint, auth_headers, {event, body}, Config.send_max_attempts())
+              |> maybe_call_after_send_event(event)
+            end)
 
-      {:error, :invalid_dsn} ->
-        {:error, :invalid_dsn}
-    end
-  end
+            {:ok, ""}
```
ruslandoga, line 145: since in `:none` mode `do_send_event` returns `{:ok, String.t}`...

lib/sentry/client.ex
```diff
@@ -51,7 +51,7 @@ defmodule Sentry.Client do
   require Logger
 
   @type send_event_result ::
-          {:ok, Task.t() | String.t() | pid()} | {:error, any()} | :unsampled | :excluded
```
ruslandoga, line 54: I removed `{:ok, pid}` from `send_event_result` type.

lib/sentry/client.ex
```diff
@@ -100,6 +99,7 @@ defmodule Sentry.Client do
         :unsampled
 
       {%Event{}, true} ->
+        put_whether_should_log(should_log?(event))
```
ruslandoga, line 102: Whether or not we should log for this event is saved to process dict.

lib/sentry/client.ex
```diff
@@ -115,107 +115,116 @@ defmodule Sentry.Client do
...
+        case result do
+          :none ->
+            should_log? = should_log?()
 
-      {:error, :invalid_dsn} ->
-        {:error, :invalid_dsn}
-    end
-  end
+            Task.Supervisor.start_child(Sentry.TaskSupervisor, fn ->
+              put_whether_should_log(should_log?)
```
ruslandoga, line 139: And `should_log?` is copied into the the task's process dict.

lib/sentry/client.ex
```diff
@@ -115,107 +115,116 @@ defmodule Sentry.Client do
...
   end
 
   @doc """
   Makes the HTTP request to Sentry using the configured HTTP client.
   """
-  @spec request(String.t(), list({String.t(), String.t()}), String.t()) ::
+  @spec request(String.t(), list({String.t(), String.t()}), String.t(), pos_integer()) ::
           {:ok, String.t()} | {:error, term()}
-  def request(url, headers, body) do
+  def request(url, headers, body, current_attempt \\ 1) do
```
ruslandoga, line 198: I've added attempt counts to error logs for `request`

ruslandoga: It looks like this:

```
10:06:20.729 [warning] Failed to send Sentry event. Attempt 1.
** (throw) :error
    test/client_test.exs:392: anonymous fn/3 in Sentry.ClientTest."test exceptions, throws, and exits handles throws"/1
    (sentry 8.0.6) lib/sentry/client.ex:202: Sentry.Client.request/4
    (sentry 8.0.6) lib/sentry/client.ex:183: Sentry.Client.retry_request/5
    (sentry 8.0.6) lib/sentry/client.ex:154: anonymous fn/5 in Sentry.Client.do_send_event/3
    (elixir 1.13.4) lib/task/supervised.ex:89: Task.Supervised.invoke_mfa/2
    (elixir 1.13.4) lib/task/supervised.ex:34: Task.Supervised.reply/4
    (stdlib 4.0.1) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
```

lib/sentry/client.ex
```diff
@@ -307,6 +316,16 @@ defmodule Sentry.Client do
     end
   end
 
+  defp maybe_log(message) do
+    if should_log?() do
+      Logger.log(
+        Config.log_level(),
+        ["Failed to send Sentry event. " | message],
+        domain: [:sentry]
+      )
```
ruslandoga, line 325: When logging, process dict is checked for whether we should log or not.

lib/sentry/client.ex
ruslandoga: I've added another commit with the changes I had in mind. I hope you like them :)

lib/sentry/client.ex
ruslandoga: I can split it into a separate PR like "fix type send_event_result type in client.ex"

lib/sentry/client.ex
ruslandoga: @mitchellhenke ping :)

ruslandoga:
@mitchellhenke Hi! Sorry to bother you about this, but I wonder if there is anything else that needs to be done for this PR to be merged?

sl0thentr0py:
@ruslandoga I'm breathing new life into this repo, would you still like to see this merged? 
can you merge latest master and push again if so?

ruslandoga:
@sl0thentr0py Sure!

ruslandoga:
Done. I'll go over this PR tomorrow to check if I still think if the correct solution.

ruslandoga:
👋 @sl0thentr0py 

Sorry for taking so long to get back to it.

I've cleaned up the PR to only fix the original issue. It should make it easier to review. I'll open another PR for direct logging from the client once this one is merged.

lib/sentry/client.ex
ruslandoga: Moved to https://github.com/getsentry/sentry-elixir/pull/545

test/client_test.exs
```diff
@@ -354,4 +355,93 @@ defmodule Sentry.ClientTest do
...
+        apply(Event, :not_a_function, [])
```
whatyouhide, line 375: To simplify this test, could you use `Sentry.capture_message/2` to avoid the `try/catch`?

ruslandoga: I'll try using `Sentry.capture_message/2`, not sure why I chose `try/catch` back then.

whatyouhide: @ruslandoga awesome. Lmk when this is ready for review. 👍

ruslandoga: @whatyouhide should be ready :)

test/test_helper.exs
```diff
@@ -5,6 +5,8 @@ Code.require_file("test/support/test_filter.exs")
 Code.require_file("test/support/test_gen_server.exs")
 Code.require_file("test/support/test_error_view.exs")
 
+Mox.defmock(Sentry.MockClient, for: Sentry.HTTPClient)
```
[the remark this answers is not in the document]
ruslandoga: Renamed in https://github.com/getsentry/sentry-elixir/pull/515/commits/61231a2add1d34ec20634689a2c3710bfd073139 :)

whatyouhide (approved):
Awesome 👏

PR #517, opened 2022-07-27
allow logging from tasks

ruslandoga:
Continuing from https://github.com/getsentry/sentry-elixir/pull/516

### DIFF
--- a/test/client_test.exs
+++ b/test/client_test.exs
@@ -330,22 +330,23 @@ defmodule Sentry.ClientTest do
       log_level: :error
     )
 
-    capture_log(fn ->
-      try do
-        apply(Event, :not_a_function, [])
-      rescue
-        e ->
-          {:ok, task} =
-            Sentry.capture_exception(
-              e,
-              stacktrace: __STACKTRACE__,
-              result: :async
-            )
+    assert capture_log(fn ->
+             try do
+               apply(Event, :not_a_function, [])
+             rescue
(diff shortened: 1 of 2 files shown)


test/client_test.exs
```diff
@@ -330,22 +330,23 @@ defmodule Sentry.ClientTest do
       log_level: :error
     )
 
-    capture_log(fn ->
```
ruslandoga, line 333: This test case was missing an `assert` and was passing without testing what it was supposed to. After adding `assert` it doesn't pass because in `:async` or `:none` modes [`maybe_log_result`](https://github.com/getsentry/sentry-elixir/blob/76b23dbaff6ec560dc806a4b107b2657a772b55a/lib/sentry/client.ex#L351-L379)'s `result` is 

```elixir
{:ok,
 %Task{
   owner: #PID<0.393.0>,
   pid: #PID<0.406.0>,
   ref: #Reference<0.1323136707.3469803524.187842>
 }}
```

That makes me think [`{:error, {:request_failure, last_error}}`](https://github.com/getsentry/sentry-elixir/blob/76b23dbaff6ec560dc806a4b107b2657a772b55a/lib/sentry/client.ex#L360-L361) is unreachable under default config and client errors would never be logged. Maybe instead of logging on task / result, logging for invalid request could happen from [within the task?](https://github.com/getsentry/sentry-elixir/pull/517)

PR #519, opened 2022-09-02
Add :remote_address_reader PlugContext option

michallepicki:
When working on masking user IP from Sentry at work, after scrubbing the IP from headers I noticed that user's IP is also visible under Environment's `REMOTE_ADDR`. One way to hide/mask the IP I thought about was to have additional plugs before and after `PlugContext`, to change and re-set the `conn.remote_ip` value. But I think a better solution would be to make reading of this value configurable.

When working on the patch I noticed the logic on the `master` branch has changed to read the `x-forwarded-for` header when present, so I made this the new default remote address reader.


michallepicki:
@mitchellhenke bump

sl0thentr0py (approved):
ty @michallepicki  for the contribution!

PR #523, opened 2022-11-11
feat: Adds `:all` option for metadata to logger backend

### LINKED ISSUE #522: Allow `metadata` for logger backend to be `:all`
We want all of our existing telemetry to integrate with Sentry, which includes adding all of our metadata. We can't analyze every app, so I want to enable "all" metadata.

isaacsanders:
Closes #522


isaacsanders:
@cleptric the CI seems broken, outside of my change.

isaacsanders:
I've tried fixing the CI, but it looks like it needs more help than I'm able to give right now.

sl0thentr0py:
hey @isaacsanders finally looking at this, sorry for the delay.
If we include all metadata, should we maybe exclude the things listed [here which are added by default](https://hexdocs.pm/logger/1.12.3/Logger.html#module-metadata)? Especially `crash_reason` which cannot be serialized as mentioned in #533 .

isaacsanders:
When is the soonest we know we can’t serialize something? If you want to start a list of things we exclude, we can start it here, but we could also move this to the request serialization part of the code.

whatyouhide:
This is obsolete now that we merged #605. Thanks @isaacsanders 💟

PR #526, opened 2022-12-14
Fix excluding domains

sonic182:
closes https://github.com/getsentry/sentry-elixir/issues/527

Using elixir and specifying a domain when logging, it doesn't exclude, because elixir uses `:elixir` domain as first key always, and sentry package only checks the first key versus the list specified in config

image

This pr is to check all domains specified in the meta `:domain` key

I'll appreciate if you make a release with this


sl0thentr0py:
duplicate of #508

PR #529, opened 2023-01-13
Wrap sentry event when the sample_rate is 0

technicalcapt:
In order to configure `sample_rate` 0 in our current project and we want to return the log before sending an event to Sentry for such environments (dev, test, staging)


technicalcapt:
Open by mistake. Closing

PR #533, opened 2023-01-30
Fix :crash_reason metadata handling

pojiro:
- [x] mix format
- [x] mix test
---

### Use case

This PR fixes following use case with `config :logger, Sentry.LoggerBackend, metadata: [:crash_reason]`.

```
  def test_with_exception() do
    try do
      raise RuntimeError
    rescue
      exception ->
        Logger.error("sentry test with exception", crash_reason: {exception, __STACKTRACE__})
    end
  end
```

Currently we cannot log for other backends and Sentry.LoggerBackend at the same time with one line.
So we need to write like following,

```
  def test_with_exception() do
    try do
      raise RuntimeError
    rescue
      exception ->
        Logger.error("sentry test with exception") # ex. for console
        Sentry.capture_exception(error, stacktrace: __STACKTRACE__)
    end
  end
```

### Why cannot currently

The following code in the current master branch cannot work because it fails to encode JSON.
Because the :crash_reason value is a tuple.

https://github.com/getsentry/sentry-elixir/blob/0e238fed32bab527c9e88c86e1ef004c737e29dd/lib/sentry/logger_backend.ex#L101-L102

[:crash_reason](https://hexdocs.pm/logger/1.14.3/Logger.html#module-metadata) is special metadata, so I think we don't need to use it for `:logger_metadata` in `:extra`.

### DIFF
--- a/lib/sentry/logger_backend.ex
+++ b/lib/sentry/logger_backend.ex
@@ -95,9 +95,10 @@ defmodule Sentry.LoggerBackend do
   end
 
   defp log(level, msg, meta, state) do
+    {crash_reason, meta} = Keyword.pop(meta, :crash_reason)
     opts = build_opts(level, meta, state)
 
-    case meta[:crash_reason] do
+    case crash_reason do
       {%_{__exception__: true} = exception, stacktrace} when is_list(stacktrace) ->
         Sentry.capture_exception(exception, [stacktrace: stacktrace] ++ opts)
 
--- a/test/logger_backend_test.exs
+++ b/test/logger_backend_test.exs
@@ -339,6 +339,37 @@ defmodule Sentry.LoggerBackendTest do
     end)
   end
 
+  test "includes Logger.metadata except for :crash_reason" do
+    Logger.configure_backend(Sentry.LoggerBackend, metadata: [:string, :crash_reason])
+    bypass = Bypass.open()
+    Process.flag(:trap_exit, true)
+    pid = self()
+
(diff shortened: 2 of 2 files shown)


pojiro:
@mitchellhenke Thank you for your great work!! Could you review this PR?

sl0thentr0py:
@pojiro thx for the PR, standalone this PR is weird since people could just not add it to the `:metadata` config for `LoggerBackend`. 
But if we merge https://github.com/getsentry/sentry-elixir/pull/523 in, this makes sense (with maybe some other defaults excluded as well), so I'll decide what to do on this once we resolve that other PR.

pojiro:
@sl0thentr0py thx for your reply and following comment.

> standalone this PR is weird since people could just not add it to the :metadata config for LoggerBackend.

I fixed my 1st comment to explain the purpose of this PR.
I will try to re-create a PR again as I may be able to resolve #523 and this.  Thanks.

pojiro:
I tried, but could not satisfy `:all`. The reason is that there is no way to determine which keys in `meta` are or are not serializable.

sl0thentr0py:
I'll need to think about this too, not that familiar with elixir. But we need some kind of general `serializable?` abstraction I guess.

pojiro:
It might be. It would be nice to be able to make changes in a direction that is consistent with the current source, but I just can't take the time to do to that.

whatyouhide:
This is obsolete now that we merged #605. Thank you @pojiro! 💟

PR #537, opened 2023-03-02
Use module attribute for dictionary key consistently

tmecklem:
get_sentry_context/0 uses @logger_metadata_key but several other places use a hardcoded :sentry atom for access. This PR makes usage consistent.


whatyouhide (approved):

whatyouhide:
Thanks @tmecklem! 💟

Issue #540, opened 2023-04-19
Sentry fails without hackney at runtime but is fine at compile time


andrew-combs:
### Environment

`~> 8.0`

### Steps to Reproduce

1. Fail to include `:hackney` in your application deps; use the default hackney adapter
2. Include something like `{:excoveralls, "~> 0.15", only: [:dev, :test]}` in `mix.exs`
3. Add `Logger.add_backend(Sentry.LoggerBackend) to `application.ex`
4. Compile and run

### Expected Result

Sentry should fail to compile due to the missing dependency.

### Actual Result

Sentry compiles and runs in dev and test environments (because `excoveralls` happens to require `hackney`), but crashes at runtime in prod.

### Notes

This is really more of an improvement to dev/x. Because of a circumstantial issue where `hackney` was brought in from another project in a non-prod environment, things appeared to work fine out of the box. However, once deployed to prod, sentry blows up because suddenly hackney is missing. Since this is something that can be known at compile-time, it would save a bit of headache to catch it then if possible.

rlopzc:
I wouldn't say is just a dev/x improvement. It literally broke our production system by not having hackney at runtime

whatyouhide:
Fixed in #554.

PR #541, opened 2023-04-19
replace runtime check with compile time check

andrew-combs:
Resolves https://github.com/getsentry/sentry-elixir/issues/540

Possible improvement might be to resolve `Application.get_env` warnings and use `Application.compile_env` when available.


whatyouhide:
Closed by #554.

PR #543, opened 2023-05-31
Fix example configuration on Sentry.Sources

scudelletti:
Hello there, this PR aims to fix and improve the example on `Sentry.Sources`.

If I understood correctly:
* `root_source_code_path` is deprecated  and expects a single path
* `root_source_code_paths` expects a list of paths.

Based on that there are 2 issues:
1. The example sets `root_source_code_path` with a list of paths. That seems an incorrect value.
2. `root_source_code_paths` should be preferred over the deprecated `root_source_code_path` config.

Thanks for your time.


whatyouhide (approved):

whatyouhide:
Thank you @scudelletti! 💟

PR #544, opened 2023-06-09
feat: add ability to scrub additional params / headers

StephaneRob:
Allow to define extra params / headers to be scrubbed by Sentry.PlugContext default scrubber.

### DIFF
--- a/lib/sentry/plug_context.ex
+++ b/lib/sentry/plug_context.ex
@@ -37,6 +37,10 @@ defmodule Sentry.PlugContext do
 
   *Please Note*: If you are sending large files you will want to scrub them out.
 
+  Alternatively, you can add extra params to be scrubbed by default scrubber.
+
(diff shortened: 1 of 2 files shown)


klacointe (approved):

whatyouhide:
`Sentry.Plug.default_body_scrubber/1` is public API, and it's documented to be the default. So, you can pass your function that scrubs the params you want, and then calls back to `default_body_scrubber/1`. I don't think there's a need here to add additional API surface 🙃

StephaneRob:
Hey @whatyouhide 
Thanks for your reply, actually it's why I did in my project (custom function + call to `default_body_scrubber/1`). But It looked like easier (at least faster) to add ability to add  custom fields to config (for built in scrubber) than create a new function that scrub custom fields and call default scrubber to scrub default one.

PR #547, opened 2023-06-13
add dependabot

ruslandoga:
This PR adds dependabot to keep with with dependencies.


whatyouhide:
Let's skip this for now, we don't need to constantly update dependencies since this is a *library*, so the `mix.lock` only matters when working on this library itself, and won't matter for things that depend on this.

PR #555, opened 2023-07-11
Remove use of deprecated Mix.Config

whatyouhide:


test/test_helper.exs
```diff
@@ -5,6 +5,7 @@ Code.require_file("test/support/test_filter.exs")
 Code.require_file("test/support/test_gen_server.exs")
 Code.require_file("test/support/test_error_view.exs")
 
-ExUnit.start()
+ExUnit.start(assert_receive_timeout: 500)
```
whatyouhide, line 8: Bonus thing to remove some config 🙃

yordis:
Do you need all those config files, btw? 

It is a lib so probably only a `test` environment is needed here (or just `config/config.exs`, no?

whatyouhide:
@yordis yes, that's next step 👍 Unless you wanna get on a PR, I'd be happy to have some help 😄

yordis:
@whatyouhide count me on it!

PR #562, opened 2023-07-23
Make Sentry.Util + Sentry.CrashError + Sentry.Config private

whatyouhide:
As discussed with @sl0thentr0py, this is targeted at the 9.0.0 release we're working on.

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -27,7 +27,7 @@ defmodule Sentry do
 
       config :sentry, dsn: "https://public:secret@app.getsentry.com/1",
         included_environments: [:prod],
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):

axelson:
@whatyouhide was this PR intended to add `@moduledoc false` to Sentry.CrashError? From the title it looks like it. But Sentry.CrashError isn't marked with `@moduledoc false` currently: https://github.com/getsentry/sentry-elixir/blob/6119acab255d341c0d4684db584530f3745a0ce3/lib/sentry/crash_error.ex

PR #564, opened 2023-07-25
 allow setting all as metadata for Logger.Backend config

mikeover:


whatyouhide:
Hey @mikeover, is this similar to https://github.com/getsentry/sentry-elixir/pull/523? I haven't had time to review either, but they _sound_ similar 😄

PR #568, opened 2023-07-28
Refactor small things here and there

whatyouhide:
Commented in the code below.

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -121,7 +121,7 @@ defmodule Sentry do
     end
 
     validate_json_config!()
-    validate_log_level_config!()
+    Config.validate_log_level!()
 
     opts = [strategy: :one_for_one, name: Sentry.Supervisor]
     Supervisor.start_link(children, opts)
@@ -226,16 +226,6 @@ defmodule Sentry do
                     """),
                     __STACKTRACE__
         end
-    end
-  end
-
-  defp validate_log_level_config!() do
-    value = Config.log_level()
-
-    if value in Config.permitted_log_level_values() do
-      :ok
-    else
-      raise ArgumentError.exception("#{inspect(value)} is not a valid :log_level configuration")
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/sentry.ex
```diff
@@ -121,7 +121,7 @@ defmodule Sentry do
     end
 
     validate_json_config!()
-    validate_log_level_config!()
+    Config.validate_log_level!()
```
whatyouhide, line 124: If we move this to the `Sentry.Config` module, we don't need to expose the "permitted log level values" from there.

lib/sentry/config.ex
```diff
@@ -142,8 +148,6 @@ defmodule Sentry.Config do
     get_config(:max_breadcrumbs, default: 100, check_dsn: false)
   end
 
-  def permitted_log_level_values, do: @permitted_log_level_values
```
whatyouhide, line 145: This is not needed now.

lib/sentry/sources.ex
```diff
@@ -66,12 +66,10 @@ defmodule Sentry.Sources do
...
 
-  def load_files do
-    Enum.reduce(
-      Config.root_source_code_paths(),
-      %{},
-      &load_files_for_root_path/2
-    )
+  @doc false
+  @spec load_files([Path.t()]) :: source_map()
+  def load_files(paths \\ Config.root_source_code_paths()) do
```
whatyouhide, line 71: If we default to `Config.root_source_code_paths/0`, then we can have tests that don't have to modify the env, and it'll work better when revamping configs, because this is really a compile-time config.

lib/sentry/config.ex
```diff
@@ -13,9 +13,15 @@ defmodule Sentry.Config do
   @default_send_max_attempts 4
   @default_log_level :warning
 
-  @permitted_log_level_values ~w(debug info warning warn error)a
+  @spec validate_log_level!() :: :ok
+  def validate_log_level!() do
+    value = log_level()
 
-  def validate_config! do
```
whatyouhide, line 18: Was unused.

sl0thentr0py (approved):

PR #572, opened 2023-07-30
Clean up Sentry.Plug.Context

whatyouhide:
* Cleaned up docs
* Made `default_remote_address_reader/1` private
* Made `scrub_map/3` and `scrub_list/3` private
* Made `build_request_interface_data/2` private

### DIFF
--- a/lib/sentry/plug_context.ex
+++ b/lib/sentry/plug_context.ex
@@ -1,100 +1,123 @@
 defmodule Sentry.PlugContext do
   @moduledoc """
   This module adds Sentry context metadata during the request in a Plug
-  application. It includes defaults for scrubbing sensitive data, and
-  options for customizing it by default.
+  application.
 
-  It is intended for usage with `Sentry.PlugCapture` as metadata added here
-  will appear in events captured.
+  It includes defaults for scrubbing sensitive data, and options for customizing it
+  by default. It is intended for usage with `Sentry.PlugCapture` as metadata added
+  here will appear in events captured.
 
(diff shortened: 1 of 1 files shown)


lib/sentry/plug_context.ex
```diff
@@ -103,16 +126,7 @@ defmodule Sentry.PlugContext do
   @scrubbed_value "*********"
```
sl0thentr0py, line 126: on most SDKs, we use `[Filtered]` to signify scrubbing just in case you want to make it consistent.

sl0thentr0py (approved):

lib/sentry/plug_context.ex
whatyouhide: Awesome, yes I can do that in another PR 👍

sc-yan:
@whatyouhide 
hi just wonder is there any specific reason to make `scrub_map/3` private again? because it was made public at [8.0.1](https://github.com/getsentry/sentry-elixir/blob/master/CHANGELOG.md#801-2020-08-08).
now when upgrading to v9, this is supposed to be a somewhat breaking change.

whatyouhide:
@sc-yan a function like `scrub_map/3` is better suited to be in your application with your application-specific logic in it, rather than exposed by an error reporting library. Removing it also allows us to be more flexible with internal changes without risking breaking APIs.

sc-yan:
@whatyouhide thanks for the prompt reply! that surely make sense

PR #576, opened 2023-07-30
Polish docs for HTTP client

whatyouhide:


lib/sentry/hackney_client.ex
```diff
@@ -22,10 +29,9 @@ defmodule Sentry.HackneyClient do
       Sentry.Config.hackney_opts()
       |> Keyword.put_new(:pool, @hackney_pool_name)
 
-    with {:ok, status, headers, client} <-
-           :hackney.request(:post, url, headers, body, hackney_opts),
-         {:ok, body} <- :hackney.body(client) do
-      {:ok, status, headers, body}
+    case :hackney.request(:post, url, headers, body, [:with_body] ++ hackney_opts) do
```
whatyouhide, line 32: Using `:with_body` avoids the extra call to `:hackney.body/1`.

lib/sentry/http_client.ex
```diff
@@ -1,21 +1,45 @@
 defmodule Sentry.HTTPClient do
```
whatyouhide, line 1: Added a few types here.

sl0thentr0py (approved):

PR #589, opened 2023-08-10
Remove Sentry.HTTPClient.child_spec/0

### LINKED ISSUE #587: Make `Sentry.HTTPClient.child_spec/1` callback be optional
My use-case is I want to use Tesla as an HTTP Client and use pre-configure Finch that is already defined in the application tree. But the `HTTPClient` requires the implementor to implement `child_spec`. 

My proposal is to make `child_spec` optional and attach it to the supervisor if it's available. I can help working on this if you're ok with this proposal. :)

Thanks

wingyplus:
Fixes #587

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -304,11 +304,12 @@ defmodule Sentry do
 
   def start(_type, _opts) do
     children = [
-      {Task.Supervisor, name: Sentry.TaskSupervisor},
-      Config.client().child_spec()
+      {Task.Supervisor, name: Sentry.TaskSupervisor}
     ]
 
-    if Config.client() == Sentry.HackneyClient do
+    client = Config.client()
+
+    if client == Sentry.HackneyClient do
       unless Code.ensure_loaded?(:hackney) do
         raise """
         cannot start the :sentry application because the HTTP client is set to \
@@ -322,6 +323,13 @@ defmodule Sentry do
         {:error, reason} -> raise "failed to start the :hackney application: #{inspect(reason)}"
       end
     end
+
+    children =
+      if client == Sentry.HackneyClient do
+        children ++ [Sentry.HackneyClient.child_spec()]
+      else
+        children
+      end
 
     Config.warn_for_deprecated_env_vars!()
     validate_json_config!()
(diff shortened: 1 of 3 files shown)


whatyouhide:
I'm wondering if we should instead just remove the `child_spec/0` callback altogether, and have users start their HTTP client in their own supervision tree.

Thoughts?

wingyplus:
> 

@whatyouhide I'm ok with this but it's ok to break the hackney configuration?

whatyouhide:
@wingyplus we don't need to break the hackney configuration. We can do something like

```elixir
client = Config.client()

children = [Task...]

children =
  if client == Sentry.HackneyClient do
    children ++ [Sentry.HackneyClient]
  else
    children
  end
```

wingyplus:
@whatyouhide Ahh, I see. PR updated. 🙇

lib/sentry.ex
```diff
@@ -304,11 +304,19 @@ defmodule Sentry do
...
 
-    if Config.client() == Sentry.HackneyClient do
+    client = Config.client()
+
+    children =
+      if client == Sentry.HackneyClient do
+        children ++ [Sentry.HackneyClient.child_spec()]
+      else
+        children
+      end
```
wingyplus, line 317: I think it should be move under hackney check.

wingyplus:
@whatyouhide what else I need do to make it merge?

whatyouhide:
Hey @wingyplus I’m figuring out whether the best strategy is to make this an optional callback or go with [my suggestion here](https://github.com/getsentry/sentry-elixir/pull/589#issuecomment-1672920725). I'll need some time since there are other things to figure out first in this lib 😄

wingyplus:
> Hey @wingyplus I’m figuring out whether the best strategy is to make this an optional callback or go with [my suggestion here](https://github.com/getsentry/sentry-elixir/pull/589#issuecomment-1672920725). I'll need some time since there are other things to figure out first in this lib 😄

Understood. I'm ok with both directions. Let me know which one you prefer. 🙇

whatyouhide:
@wingyplus I was heads down on Sentry stuff and ended up doing this myself, but it's your original solution with the optional callback. Thanks! 💟

PR #590, opened 2023-08-15
Clean up Sentry.Envelope

whatyouhide:
* Improved encoding performance
* Reduced API surface (this is a private module, so no public-facing changes)
* Reduced amount of code

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -72,8 +72,8 @@ defmodule Sentry.Client do
   @spec encode_and_send(Event.t(), result()) :: send_event_result()
   defp encode_and_send(event, result) do
     result =
-      Sentry.Envelope.new()
-      |> Sentry.Envelope.add_event(event)
+      event
+      |> Sentry.Envelope.new()
       |> Sentry.Envelope.to_binary()
       |> case do
         {:ok, body} ->
--- a/lib/sentry/envelope.ex
+++ b/lib/sentry/envelope.ex
@@ -1,122 +1,100 @@
 defmodule Sentry.Envelope do
   @moduledoc false
+  # https://develop.sentry.dev/sdk/envelopes/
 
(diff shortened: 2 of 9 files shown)


lib/sentry/client.ex
```diff
@@ -72,8 +72,8 @@ defmodule Sentry.Client do
   @spec encode_and_send(Event.t(), result()) :: send_event_result()
   defp encode_and_send(event, result) do
     result =
-      Sentry.Envelope.new()
-      |> Sentry.Envelope.add_event(event)
```
whatyouhide, line 76: We only ever add a single event, so no need for `add_event/2`.

lib/sentry/envelope.ex
```diff
@@ -1,145 +1,127 @@
...
-    |> Map.put(:event_id, event_id)
-    |> Map.update!(:items, fn items ->
-      items ++ [event]
-    end)
+  @spec to_binary(t()) :: {:ok, binary()} | {:error, any()}
+  def to_binary(%__MODULE__{} = envelope) do
+    json_library = Config.json_library()
+
+    encoded_items =
+      Enum.map(envelope.items, fn item ->
```
whatyouhide, line 34: `Enum.map/2` + `throw`/`catch` using iodata is really good for speed 😉

lib/sentry/envelope.ex
```diff
@@ -1,145 +1,127 @@
...
-    envelope
-  end
-
-  # Returns the event in the envelope if one exists.
-  @spec event(t()) :: Event.t() | nil
-  def event(envelope) do
-    envelope.items
-    |> Enum.filter(fn item -> is_event?(item) end)
-    |> List.first()
-  end
```
whatyouhide, line 93: Moved these to a test helper.

sl0thentr0py:
i would revert this tbh, we need this later

PR #592, opened 2023-08-15
Make Sentry.send_event/1,2 private

whatyouhide:
This is in preparation of making the Sentry event struct(s) private, since they changed in Sentry's API and we want to be able to modify these flexibly.

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -341,7 +341,7 @@ defmodule Sentry do
   the `:included_environments`. See the [*Configuration* section](#module-configuration)
   in the module documentation.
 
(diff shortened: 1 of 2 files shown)


whatyouhide:
Got it... Well, what if we do make all "payloads" public then? I’m ok with that, but we better make stuff like stacktraces, frames, exceptions, and so on public IMO. What do you think @sl0thentr0py? 🙃

sl0thentr0py:
yes I think exposing all that stuff is fine

whatyouhide:
Okay, #595 was merged so this is not necessary anymore. 🙃

PR #601, opened 2023-09-03
Introduce `Sentry.Transport` and a sender pool

whatyouhide:
This is a big-ass PR, but I was having trouble making it smaller 🙃 

Here are the things I’m doing:

  1. I introduced a private module, `Sentry.Transport`, which takes care of actually making the HTTP requests to Sentry. This is in line with the terminology described in the [unified API docs](https://develop.sentry.dev/sdk/unified-api/#event-pipeline). `Sentry.Transport` only knows how to POST a single envelope (with potentially multiple events in it) to Sentry.
  2. Now that `Sentry.Client` is private, I could restructure it to take advantage of `Sentry.Transport` in order to send events.
  3. I introduced `Sentry.Transport.Sender` and `Sentry.Transport.SenderPool`. These two modules start a pool of GenServers that send stuff to Sentry when `result: :none`. Each sender keeps a queue of events which it flushes whenever either 10 events are queued, or 500ms pass. This also takes advantage of multiple events in a single envelope.


config/config.exs
```diff
@@ -8,6 +8,7 @@ config :sentry,
   root_source_code_paths: [File.cwd!()],
   hackney_opts: [recv_timeout: 50],
   send_result: :sync,
-  send_max_attempts: 1
+  send_max_attempts: 1,
+  __sender_module__: Sentry.TransportSenderMock
```
whatyouhide, line 12: Internal config to make testing the Logger backend easier. This is needed since the Logger backend forces `result: :none`, so it's really hard to test it if it's going through the async sender pool.

lib/sentry.ex
```diff
@@ -251,7 +251,7 @@ defmodule Sentry do
 
       config :sentry,
         before_send_event: {MyModule, :before_send},
-        before_send_event: {MyModule, :after_send}
+        after_send_event: {MyModule, :after_send}
```
whatyouhide, line 254: Typo

PR #602, opened 2023-09-03
Fix event data that is not JSON-encodable

### LINKED ISSUE #521: Handle events with metadata that can't be JSON-encoded
### Environment

8.0.6

### Steps to Reproduce

* Have a configured logger backend with metadata fields that can't be JSON encoded.

```
config :logger, Sentry.LoggerBackend,
  level: :error,
  metadata: [:mfa],
  capture_log_messages: true
```
* Use Logger to log an error message.

### Expected Result

Some way to know that the event can't be sent to Sentry because it was not JSON encodable. This library could handle encoding the common types, but the user still might add something to the metadata that isn't JSON encodable. Since deadlocks can happen when logging in a logger maybe send an event to Sentry that another event can't be JSON encoded ?

### Actual Result

Logged error is not sent to Sentry and there is no indication to why.

### Notes

This issue is very important to us as we have missed multiple errors in production not knowing they were happening.

I think this is related to #520 since pids are also not JSON encodable by default.

whatyouhide:
Closes #521.

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -104,11 +104,16 @@ defmodule Sentry.Client do
 
   @spec render_event(Event.t()) :: map()
   def render_event(%Event{} = event) do
+    json_library = Config.json_library()
+
     event
     |> Map.from_struct()
     |> update_if_present(:message, &String.slice(&1, 0, @max_message_length))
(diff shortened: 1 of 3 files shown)


sl0thentr0py (commented):
nice thx for this!

PR #619, opened 2023-09-28
Attempt to scrub all `Plug.Conn`s in `Sentry.PlugCapture`

### LINKED ISSUE #477: Sensitive data not scrubbed out of `Phoenix.ActionClauseError`
## Important Details

How are you running Sentry?

Saas (sentry.io)

## Description

Sensitive data is not scrubbed from exception stack traces.

This is related to https://github.com/getsentry/sentry/issues/9309, which didn't get any answers.

## Steps to Reproduce

We have a Phoenix application. The endpoint looks something like this:

```elixir
defmodule MyAppWeb.Endpoint do
  use Sentry.PlugCapture
  use Phoenix.Endpoint, otp_app: :my_app_web

  @scrubbed_param_keys [
    "current_password",
    "email",
    "password",
    "password_confirmation"
  ]

  # ...

  plug Sentry.PlugContext, body_scrubber: &MyAppWeb.Endpoint.body_scrubber/1

  # ...
  plug MyAppWeb.Router

  def body_scrubber(conn) do
    Sentry.PlugContext.scrub_map(conn.params, @scrubbed_param_keys)
  end
end
```

We found an `Phoenix.ActionClauseError` logged in Sentry, but the parameters were not scrubbed. The log entry looks like this (irrelevant information redacted for this issue with `...`; values we expected to be scrubbed are replaced with `UNREDACTED`):

```
Phoenix.ActionClauseError: no function clause matching in MyAppWeb.MyController.update/2

The following arguments were given to MyAppWeb.MyController.update/2:

    # 1
    %Plug.Conn{adapter: {Plug.Cowboy.Conn, :...}, assigns: %{changeset: #Ecto.Changeset<action: nil, changes: %{}, errors: [...], data: #User<>, valid?: false>, ...}, ..., body_params: %{"_method" => "put", "current_password" => "UNREDACTED", "user" => %{"password" => "UNREDACTED", "password_confirmation" => "UNREDACTED"}}, ..., params: %{..., "current_password" => "UNREDACTED", "user" => %{"password" => "UNREDACTED", "password_confirmation" => "UNREDACTED"}}, ...}

    # 2
    %{"_csrf_token" => "...", "_method" => "put", "current_password" => "UNREDACTED", "user" => %{"password" => "UNREDACTED", "password_confirmation" => "UNREDACTED"}}

  File "lib/my_app_web/controllers/my_controller.ex", line 24, in MyAppWeb.MyController.update/2
  File "lib/my_app_web/controllers/my_controller.ex", line 1, in MyAppWeb.MyController.action/2
  File "lib/my_app_web/controllers/my_controller.ex", line 1, in MyAppWeb.MyController.phoenix_controller_pipeline/2
  File "lib/phoenix/router.ex", line 352, in Phoenix.Router.__call__/2
  File "lib/my_app_web/endpoint.ex", line 1, in MyAppWeb.Endpoint.plug_builder_call/2
  File "lib/my_app_web/endpoint.ex", line 1, in MyAppWeb.Endpoint."call (overridable 3)"/2
  File "lib/my_app_web/endpoint.ex", line 1, in MyAppWeb.Endpoint.call/2
  File "lib/phoenix/endpoint/cowboy2_handler.ex", line 65, in Phoenix.Endpoint.Cowboy2Handler.init/4
```

### What you expected to happen

We expected all fields set in `@scrubbed_param_keys` to be scrubbed from logged exceptions. However, they are sent to Sentry in clear text. Did we miss some configuration parameter or is this a bug?

whatyouhide:
Closes #477.

This is not foolproof, but I think it's pretty effective.

### DIFF
--- a/lib/sentry/plug_capture.ex
+++ b/lib/sentry/plug_capture.ex
@@ -1,11 +1,14 @@
 defmodule Sentry.PlugCapture do
   @moduledoc """
-  Provides basic functionality to handle and send errors occurring within
+  Provides basic functionality to capture and send errors occurring within
   Plug applications, including Phoenix.
 
-  It is intended for usage with `Sentry.PlugContext`.
+  It is intended for usage with `Sentry.PlugContext`, which adds relevant request
+  metadata to the Sentry context before errors are captured.
 
   ## Usage
+
+  ### With Phoenix
 
   In a Phoenix application, it is important to use this module **before**
   the Phoenix endpoint itself. It should be added to your `endpoint.ex` file:
@@ -17,7 +20,9 @@ defmodule Sentry.PlugCapture do
         # ...
       end
 
-  In a Plug application, it can be added *below* your router:
+  ### With Plug
+
+  In a Plug application, you can add this module *below* your router:
 
       defmodule MyApp.PlugRouter do
         use Plug.Router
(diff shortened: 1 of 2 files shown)


lib/sentry/plug_capture.ex
```diff
@@ -66,4 +117,38 @@ defmodule Sentry.PlugCapture do
       end
     end
   end
+
+  @doc false
+  def __capture_exception__(exception, stacktrace, scrubber) do
+    # We can't pattern match here, because we're not guaranteed to have
+    # Phoenix available.
+    exception =
+      if is_struct(exception, Phoenix.ActionClauseError) do
```
sl0thentr0py, line 126: isn't doing this just for `ActionClauseError` too limited?
like it solves the issue but similar leakage could happen in other exceptions too right?

sl0thentr0py (approved):
left a comment if we want to add other exception types, make that list configurable? 
otherwise lgtm

whatyouhide:
@sl0thentr0py sadly we can't really do that, because we're being very specific with the shape of the `Phoenix.ActionClauseError` exception here. This exception has a `args` field that other exceptions might not. In any case, users can always filter or change reported events with `:before_send_event`, so this is mostly a convenience so that Phoenix user get this scrubbing done out of the box.

PR #636, opened 2023-10-29
Add docs for custom HTTP clients in umbrellas

### LINKED ISSUE #631: Configuring HTTP client in an umbrella app
I have an Umbrella app with multiple independently-deployable apps.
Each of these apps has their own HTTP client, e.g. `foo_app` would have:
```elixir
    children = [
      {Finch, name: FooApp.Finch},
    ]
  ```
  while `bar_app` would have:
  ```elixir
    children = [
      {Finch, name: BarApp.Finch},
    ]
  ```
  
 When configuring Sentry, I can only give one custom client:
 ```elixir
 config :sentry, client: MyApp.SentryFinchHTTPClient
  ```
Which forces me to have only one Finch instance, and so, create a new Umbrella app with a shared Finch instance and shared configurations.

whatyouhide:
Closes #631.

### DIFF
--- a/lib/sentry/http_client.ex
+++ b/lib/sentry/http_client.ex
@@ -66,6 +66,30 @@ defmodule Sentry.HTTPClient do
       config :sentry,
         client: MyApp.SentryFinchHTTPClient
 
+  ### Umbrella Apps
+
+  The HTTP client for Sentry is configured globally for the `:sentry` application. In an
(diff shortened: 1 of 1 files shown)


thojanssens:
Was this the only way? It wouldn't have been possible to allow for individual config (per app)?

I will also refer to this:
https://hexdocs.pm/elixir/1.15.7/library-guidelines.html#avoid-application-configuration

PR #643, opened 2023-11-08
Fix typo in the docs

dvic:


whatyouhide:
Great catch, thanks @dvic 💟

PR #648, opened 2023-11-15
Fix dialyzer warning

### LINKED ISSUE #647: Dialyzer warning when extra_return is set
### Environment

Master from github

### Steps to Reproduce

1. Add `:unmatched_returns` to dialyzer configuration

```diff
modified   mix.exs
@@ -14,6 +14,7 @@ defmodule Sentry.Mixfile do
       deps: deps(),
       elixirc_paths: elixirc_paths(Mix.env()),
       dialyzer: [
+        flags: [:unmatched_returns, :error_handling, :extra_return],
         plt_file: {:no_warn, "priv/plts/dialyzer.plt"},
         plt_core_path: "priv/plts",
         plt_add_deps: :app_tree,

[back]
```

2. Run `mix dialyzer` 

### Expected Result

Dialyzer should not find any errors. 

### Actual Result

Dialyzer finds the following error:

```
lib/sentry/logger_backend.ex:96:unmatched_return
The expression produces a value of type:


  :excluded
  | :ignored
  | nil
  | :ok
  | :unsampled
  | {:error, :invalid_dsn | {:invalid_json, _} | {:request_failure, _}}
  | {:ok, binary()}


but this value is unmatched.
```

This may not be a problem for sentry-elixir itself (after all, you can just choose not to add that extra configuration to the dialyzer options, but any application that uses `use Sentry.PlugCapture` and has that extra configuration set, will be affected.

Munksgaard:
When dialyzer is run on this project with the `:unmatched_returns` setting, a warning is generated for this piece of code. Here is one attempt at fixing the warning, but it should be noted that this fix will cause a change in behavior: If the call to `log` does not return `:ok`, an error will occur. I couldn't really tell what the desired outcome was here, perhaps you'd rather we match on `_`?

Fixes #647

### DIFF
--- a/lib/sentry/logger_backend.ex
+++ b/lib/sentry/logger_backend.ex
@@ -93,7 +93,8 @@ defmodule Sentry.LoggerBackend do
 
     if Logger.compare_levels(level, state.level) != :lt and
          not LoggerUtils.excluded_domain?(meta[:domain] || [], state.excluded_domains) do
-      log(level, msg, meta, state)
+      _ = log(level, msg, meta, state)
+      :ok
     end
 
     {:ok, state}


lib/sentry/logger_backend.ex
```diff
@@ -93,7 +93,8 @@ defmodule Sentry.LoggerBackend do
 
     if Logger.compare_levels(level, state.level) != :lt and
          not LoggerUtils.excluded_domain?(meta[:domain] || [], state.excluded_domains) do
-      log(level, msg, meta, state)
+      :ok = log(level, msg, meta, state)
+      :ok
```
whatyouhide, line 97: I think this should rather be

```suggestion
      _ = log(level, msg, meta, state)
      :ok
```

as we're not guaranteed that `log/4` always returns `:ok`, right?

Munksgaard: That's correct, and it's essentially what I wanted your feedback on. I'll change it as you suggest.

whatyouhide (approved):

PR #649, opened 2023-11-17
Rename `:before_send_event` to `:before_send` (and hard-deprecate)

whatyouhide:
This is in line with all other SDKs. Straight from the `common` section of sentry-docs:

> All Sentry SDKs support the `before_send` callback method

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -99,7 +99,7 @@ defmodule Sentry do
   ## Filtering Exceptions
 
   If you would like to prevent Sentry from sending certain exceptions, you can
-  use the `:before_send_event` configuration option. See the [*Event Callbacks*
+  use the `:before_send` configuration option. See the [*Event Callbacks*
   section](#module-event-callbacks) below.
 
   Before v9.0.0, the recommended way to filter out exceptions was to use a *filter*,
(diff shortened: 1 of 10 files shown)


HazAT:
Isn't this a breaking change? ^^
Also, is hard_deprecate = breaking change? Never heard it before tbh

HazAT:
Ahh this is for a major, got it 👍

lib/sentry.ex
```diff
@@ -108,14 +108,14 @@ defmodule Sentry do
...
+  You can configure the `:before_send` and `:after_send_event` options to
+  customize what happens before and/or after sending an event. The `:before_send`
   callback must be of type `t:before_send_event_callback/0` and the `:after_send_event`
   callback must be of type `t:after_send_event_callback/0`. For example, you
   can set:
 
       config :sentry,
-        before_send_event: {MyModule, :before_send},
+        before_send: {MyModule, :before_send},
         after_send_event: {MyModule, :after_send}
```
HazAT, line 119: Should we also rename this then? (It doesn't exist in other SDKs)
`after_send_event`

sl0thentr0py (approved):

lib/sentry.ex
whatyouhide: @HazAT it doesn't exist in other SDKs as you said, and that's exactly why I left it untouched. It's one less change folks have to deal with. Honestly, I’m not even sure it's very much needed but hey, it's there 🤷 🙃

whatyouhide:
> Also, is hard_deprecate = breaking change? Never heard it before tbh

I'm not sure how standard of a term it is, but "hard-deprecate" would mean deprecate *and emit a warning*, while "soft-deprecate" would mean deprecate (in changelog) and hide in the docs, but without a warning.

krns:
Hey, just read the changelog of v10 and found the use of `before_send_event` is the Fingerprinting section of the readme.   I could prepare a PR if desired. 

https://github.com/getsentry/sentry-elixir#fingerprinting

whatyouhide:
Hey @krns thanks for the catch! Fixed in `master` 🙃

PR #650, opened 2023-11-20
Implement event deduplication

whatyouhide:


sl0thentr0py (changes requested):
fine with me, this is a slightly different implementation than other SDKs.

In other SDKs we have something called an 'event processor' (on scope or global)
https://github.com/getsentry/sentry-python/blob/52eb367442856d5e17a3eb6bb3343986f65cd662/sentry_sdk/scope.py#L518-L520
https://github.com/getsentry/sentry-python/blob/52eb367442856d5e17a3eb6bb3343986f65cd662/sentry_sdk/scope.py#L58

that run on `scope.apply_to_event` and `dedupe` registers an event processor that uses a thread local variable to just dedupe the last seen exception.

This one is storing some state in memory for 30 seconds but that seems fine and should do the job.

The only thing I'd ask is to make this on by default and make it possible to opt-out with a config.
 We also have this abstraction called 'integrations' in other SDKs to configure these things but maybe that's overkill right now, we can add something like that later if we have more configurable high-level behaviors.

sl0thentr0py (approved):
ty!

PR #652, opened 2023-11-22
Fix the example of using Task.Supervisor in upgrade guide.

RudolfMan:


whatyouhide (approved):

whatyouhide:
Great catch, thanks @RudolfMan! 💟

PR #653, opened 2023-11-23
Fix reading of config in "mix sentry.package_source_code"

whatyouhide:


lib/mix/tasks/sentry.package_source_code.ex
```diff
@@ -56,9 +56,12 @@ defmodule Mix.Tasks.Sentry.PackageSourceCode do
   def run(args) do
     {opts, _args} = OptionParser.parse!(args, strict: @switches)
 
+    # We can read the config here because this task depends on the `app.config` task.
+    config = Application.get_all_env(:sentry)
```
whatyouhide, line 60: Due to the way we cache config, loading `app.config` is not enough, we also need to validate it in `Sources.load_files/1`. We don't want to start the `:sentry` application for this task anyways.

sl0thentr0py (approved):
is there another test that covers the source code reporting logic?

whatyouhide:
@sl0thentr0py yes there are other tests 🙃

PR #659, opened 2023-11-27
Add upgrade guide links to the changelog

axelson:
Currently from the changelog it is difficult to find the upgrade guide since they aren't linked.


whatyouhide:
Great idea, thanks @axelson! 💟

PR #673, opened 2023-12-11
Add support for attachments

### LINKED ISSUE #479: Attachments

whatyouhide:
Closes #479.

### DIFF
--- a/lib/sentry/envelope.ex
+++ b/lib/sentry/envelope.ex
@@ -2,11 +2,11 @@ defmodule Sentry.Envelope do
   @moduledoc false
   # https://develop.sentry.dev/sdk/envelopes/
 
-  alias Sentry.{Config, Event, UUID}
+  alias Sentry.{Attachment, Config, Event, UUID}
 
   @type t() :: %__MODULE__{
           event_id: UUID.t(),
-          items: [Event.t(), ...]
+          items: [Event.t() | Attachment.t(), ...]
         }
 
   @enforce_keys [:event_id, :items]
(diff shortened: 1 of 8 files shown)


whatyouhide:
@sl0thentr0py ready for review!

lib/sentry/envelope.ex
```diff
@@ -2,37 +2,35 @@ defmodule Sentry.Envelope do
...
+  Creates a new envelope containing the given event and all of its attachments.
   """
-  @spec new([Event.t(), ...]) :: t()
-  def new([%Event{event_id: event_id}] = events) do
+  @spec from_event(Event.t()) :: t()
+  def from_event(%Event{event_id: event_id} = event) do
     %__MODULE__{
       event_id: event_id,
-      items: events
+      items: [event] ++ event.attachments
```
sl0thentr0py, line 22: maybe later let's add an `Envelope.add_item` method too when we need other types so that envelope is general enough

sl0thentr0py (approved):
lgtm 
just added an observation for later

lib/sentry/envelope.ex
whatyouhide: Yep we'll probably need something like that down the line!

PR #678, opened 2023-12-22
Fix ArgumentError when using custom scrubber with PlugCapture

paulstatezny:
[The docs for Sentry.PlugCapture](https://hexdocs.pm/sentry/10.1.0/Sentry.PlugCapture.html#module-scrubbing-sensitive-data) reference the custom `:scrubber` option for `use Sentry.PlugCapture`:

> a term of type `{module, function, args}` that will be invoked to scrub sensitive data from [Plug.Conn](https://hexdocs.pm/plug/1.11.1/Plug.Conn.html) structs.

However, when you provide `{module, function, args}` as instructed:

```elixir
use Sentry.PlugCapture, scrubber: {MyModule, :scrub_conn, []}
```

You get this compilation error:

```
== Compilation error in file my_endpoint.ex ==
** (ArgumentError) expected :scrubber to be a {module, function, args} tuple, got: {:{}, [line: 4], [MyModule, :scrub_conn, []]}
```

This PR removes the `Macro.escape` that was causing this.

### DIFF
--- a/lib/sentry/plug_capture.ex
+++ b/lib/sentry/plug_capture.ex
@@ -72,7 +72,7 @@ defmodule Sentry.PlugCapture do
 
   defmacro __using__(opts) do
     quote do
-      opts = unquote(Macro.escape(opts))
+      opts = unquote(opts)
       default_scrubber = {unquote(__MODULE__), :default_scrubber, []}
 
       scrubber =
--- a/test/plug_capture_test.exs
+++ b/test/plug_capture_test.exs
@@ -34,6 +34,22 @@ defmodule Sentry.PlugCaptureTest do
     plug PhoenixRouter
   end
 
+  defmodule Scrubber do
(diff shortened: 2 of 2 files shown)


paulstatezny:
You can reproduce the compilation error by running the test against the code in `master`.

whatyouhide:
Ah, great catch, thank you @paulstatezny 💟

whatyouhide (approved):

paulstatezny:
@whatyouhide Thank you!

Question: I intended to use this feature to redact certain segments of certain URLs. But it seems the URL passed to Sentry uses the original conn, not the scrubbed one.

Do you have a recommendation for redacting sensitive URL segments? I know it's an odd case. Would you be open to a PR that adds that feature similar to the header / body / param scrubbers?

whatyouhide:
Yes absolutely!

PR #679, opened 2023-12-30
Add `Sentry.Interfaces.Span`

whatyouhide:
Introducing the data structure.

### DIFF
--- a/lib/sentry/interfaces.ex
+++ b/lib/sentry/interfaces.ex
@@ -277,8 +277,6 @@ defmodule Sentry.Interfaces do
 
     @moduledoc since: "11.0.0"
 
-    alias Sentry.Interfaces
-
     @typedoc since: "11.0.0"
     @type t() :: %__MODULE__{
             start_timestamp: String.t() | number(),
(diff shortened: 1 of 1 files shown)


lib/sentry/interfaces.ex
```diff
@@ -267,4 +267,62 @@ defmodule Sentry.Interfaces do
...
+    @type t() :: %__MODULE__{
+            start_timestamp: String.t() | number(),
+            timestamp: String.t() | number(),
+            span_id: String.t(),
+            parent_span_id: String.t() | nil,
+            description: String.t(),
+            tags: map(),
+            trace_id: String.t() | nil,
+            op: String.t(),
+            contexts: Interfaces.context(),
```
sl0thentr0py, line 292: Spans do not have contexts themselves in the protocol, in some SDKs this is held on the `Transaction` object as a reference and then merged into the top level `contexts` on the event payload.
It depends on implementation if you want to also have this reference here or not.

lib/sentry/interfaces.ex
```diff
@@ -267,4 +267,62 @@ defmodule Sentry.Interfaces do
...
+            spans: [t()],
+            status: String.t(),
+            data: map(),
+            fingerprint: [String.t()],
```
sl0thentr0py, line 296: I don't see the `fingerprint` in the protocol either

sl0thentr0py (commented):
couple of comments, but looks fine and we can tweak later.

In general, I need to find out what the status of our 'only spans' project on the ingest side is. I'll get back to you after I find or make people write some up to date docs. Feel free to merge this,

PR #686, opened 2024-02-01
Add support for ignoring exceptions from `Sentry.LoggerHandler`

gsychev:
Adds ability to configure a list of exceptions which will not be sent to Sentry from `LoggerHandler`

### DIFF
--- a/lib/sentry/logger_handler.ex
+++ b/lib/sentry/logger_handler.ex
@@ -80,6 +80,11 @@ defmodule Sentry.LoggerHandler do
       the configured list will not be sent. Defaults to `[:cowboy]` to avoid
       double-reporting events from `Sentry.PlugCapture`.
 
+    * `:ignored_exceptions` (list of `t:module/0` of Elixir exceptions) - any messages
+      with a crash reason containing exception from configured list will not be sent.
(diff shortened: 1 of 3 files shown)


whatyouhide:
Hi @gsychev, thanks for the contribution! I think you can already do this with the [`:before_send`](https://hexdocs.pm/sentry/10.1.0/Sentry.html#t:before_send_event_callback/0) config today, right?

gsychev:
@whatyouhide , perhaps. As far as I can see in doc and examples, `before_send` can only modify event, while this PR adds filtering

gsychev:
Aha, it is possible to filter, but it is unclear from documentation

gsychev:
Alright, then it's up to you if you want to merge this PR or not, since it is just a little bit more efficient filtering, which filters exception before building the event

whatyouhide:
> Aha, it is possible to filter, but it is unclear from documentation

Would you be able to send a PR for improved docs? 🙃 

> Alright, then it's up to you if you want to merge this PR or not, since it is just a little bit more efficient filtering, which filters exception before building the event

If we already have a mechanism, let's rely on that instead of increasing the API surface.

whatyouhide:
Closing this as we're not planning to move forward with it.

PR #687, opened 2024-02-01
Add url_scrubber for redacting URLs

paulstatezny:
@whatyouhide indicated [here](https://github.com/getsentry/sentry-elixir/pull/678#issuecomment-1868330165) that this PR would be welcome. 🙂 

This PR adds support for a `url_scrubber` which works just like `body_scrubber`, `header_scrubber`, and `cookie_scrubber`.

The "default scrubber" behavior is to not redact anything. (Just runs `Plug.Conn.request_url/1` as before.)

### DIFF
--- a/lib/sentry/plug_context.ex
+++ b/lib/sentry/plug_context.ex
@@ -44,7 +44,7 @@ defmodule Sentry.PlugContext do
 
   You can also pass it in as a `{module, fun}`, like so:
 
-      plug Sentry.PlugContext, body_scrubber: {MyModule, :scrub_params}
+      plug Sentry.PlugContext, body_scrubber: {MySentryScrubber, :scrub_params}
 
   > #### Large Files {: .tip}
   >
@@ -57,7 +57,7 @@ defmodule Sentry.PlugContext do
   configured similarly to body params, through the `:header_scrubber` configuration
   option:
 
-      defmodule MyHeaderScrubber do
+      defmodule MySentryScrubber do
         def scrub_headers(conn) do
           # In this example, we do not want to include Content-Type or User-Agent
           # in reported headers, so we drop them.
@@ -70,11 +70,11 @@ defmodule Sentry.PlugContext do
 
   Then, pass it into `Sentry.PlugContext`:
 
-      plug Sentry.PlugContext, header_scrubber: &MyHeaderScrubber.scrub_headers/1
+      plug Sentry.PlugContext, header_scrubber: &MySentryScrubber.scrub_headers/1
 
   It can also be passed in as a `{module, fun}` like so:
 
-      plug Sentry.PlugContext, header_scrubber: {MyModule, :scrub_headers}
+      plug Sentry.PlugContext, header_scrubber: {MySentryScrubber, :scrub_headers}
 
   ### Scrubbing Cookies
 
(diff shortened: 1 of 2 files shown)


lib/sentry/plug_context.ex
```diff
@@ -152,7 +175,7 @@ defmodule Sentry.PlugContext do
       |> Plug.Conn.fetch_query_params()
 
     %{
-      url: Plug.Conn.request_url(conn),
+      url: apply_fun_with_conn(conn, url_scrubber),
```
paulstatezny, line 178: Using `apply_fun_with_conn` here is consistent with how the other scrubbers are called.

However, it also has the effect that passing in `nil` for `url_scrubber` will cause `%{}` to be sent instead of `Plug.Conn.request_url(conn)`.

Does this matter? Should `apply_fun_with_conn` get a new 3rd `default` argument that defaults to `%{}`?

```elixir
defp apply_fun_with_conn(conn, function, default \\ %{})
defp apply_fun_with_conn(_conn, _function = nil, default), do: default
defp apply_fun_with_conn(conn, {module, fun}, _default), do: apply(module, fun, [conn])
defp apply_fun_with_conn(conn, fun, _default) when is_function(fun, 1), do: fun.(conn)
```

```elixir
url: apply_fun_with_conn(conn, url_scrubber, Plug.Conn.request_url(conn)),
```

lib/sentry/plug_context.ex
```diff
@@ -70,11 +70,11 @@ defmodule Sentry.PlugContext do
 
   Then, pass it into `Sentry.PlugContext`:
 
-      plug Sentry.PlugContext, header_scrubber: &MyHeaderScrubber.scrub_headers/1
+      plug Sentry.PlugContext, header_scrubber: &MySentryScrubber.scrub_headers/1
 
   It can also be passed in as a `{module, fun}` like so:
 
-      plug Sentry.PlugContext, header_scrubber: {MyModule, :scrub_headers}
+      plug Sentry.PlugContext, header_scrubber: {MySentryScrubber, :scrub_headers}
```
paulstatezny, line 77: It's an insignificant detail, but there was some inconsistency with what module names were used. Since I was updating related docs and wanted to be consistent, thought it felt clean to set them all to `MySentryScrubber`.

Happy to revert if this isn't preferred!

lib/sentry/plug_context.ex
whatyouhide: Yes let's go with a third argument, but without the `\\` clause. Just do `apply_fun_with_conn(conn, function, default)` and explicitly pass in `%{}` in the other instances where we call `apply_fun_with_conn/2` today.

test/sentry/plug_context_test.exs
```diff
@@ -105,6 +109,13 @@ defmodule Sentry.PlugContextTest do
     assert %{"not-secret" => "not-secret"} == Sentry.Context.get_all().request.cookies
   end
 
+  test "allows configuring url scrubber" do
```
whatyouhide, line 112: Tiny but mighty 😉 

```suggestion
  test "allows configuring URL scrubber" do
```

lib/sentry/plug_context.ex
```diff
@@ -84,7 +84,29 @@ defmodule Sentry.PlugContext do
 
   For example:
 
-      plug Sentry.PlugContext, cookie_scrubber: &MyCookieScrubber.scrub_cookies/1
+      plug Sentry.PlugContext, cookie_scrubber: &MySentryScrubber.scrub_cookies/1
+
+  ### Scrubbing URLs
```
whatyouhide, line 89: We need to say that this has not always been available

```suggestion
  ### Scrubbing URLs

  *Available since v10.2.0.*
```

lib/sentry/plug_context.ex
paulstatezny: Will do! I like the explicitness.

lib/sentry/plug_context.ex
paulstatezny: Done! Thanks for the `suggestion`.

test/sentry/plug_context_test.exs
paulstatezny: Done!

paulstatezny:
@whatyouhide Thank you for the feedback! Addressed all of it in https://github.com/getsentry/sentry-elixir/pull/687/commits/de2fc94a82bbb73e2f1787083744191d0c3935da.

CI is failing in the same way as before; I think this is a fluke, as I cannot reproduce locally.

```
  1) test reporting from non-allowed child processes (Sentry.TestTest)
Error:      test/sentry/test_test.exs:145
     Assertion failed, no matching message after 1000ms
     The following variables were pinned:
       monitor_ref = #Reference<0.1770092520.2537553923.95488>
     The process mailbox is empty.
     code: assert_receive {:DOWN, ^monitor_ref, _, _, :normal}
     stacktrace:
       test/sentry/test_test.exs:171: (test)
```

Please let me know if you think I need to adjust the changes in any way!

whatyouhide:
@paulstatezny yeah CI is unrelated, no worries. Thank you! 💟

PR #691, opened 2024-02-12
feat(source packaging): allow custom output path in task

razielgn:
This fixes packaging an app's source code for Sentry v10 when building in highly reproducible environments (like nix), where the deps' sources are read-only (i.e. cannot write to Sentry's `priv` dir).
I've added an example to the docs.

### DIFF
--- a/lib/mix/tasks/sentry.package_source_code.ex
+++ b/lib/mix/tasks/sentry.package_source_code.ex
@@ -43,9 +43,34 @@ defmodule Mix.Tasks.Sentry.PackageSourceCode do
   > source-context-related options in compile-time config files (like `config/config.exs`
   > or `config/prod.exs`).
 
+  > #### Building with Nix {: .tip}
+  >
+  > If you build your application using nixpkgs' `mixRelease`,
+  > `mix sentry.package_source_code` will fail, because Sentry's source is read-only.
+  >
+  > To fix this, you can use the `--output` option to write the map file to a writable location,
+  > then copy it to Sentry's `priv` in the final OTP release.
+  >
+  > Example:
+  >
+  > ```nix
+  > mixRelease {
+  >   preBuild = ''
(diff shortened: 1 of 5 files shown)


test/mix/sentry.package_source_code_test.exs
```diff
@@ -20,15 +20,26 @@ defmodule Mix.Tasks.Sentry.PackageSourceCodeTest do
...
+    put_test_config(
+      root_source_code_paths: [File.cwd!()],
+      enable_source_code_context: true
+    )
 
-    assert %{"version" => 1, "files_map" => source_map} =
-             :erlang.binary_to_term(contents, [:safe])
+    expected_path =
+      [System.tmp_dir!(), "sentry.map"]
+      |> Path.join()
```
whatyouhide, line 34: Instead of this and then the `try`/`after`, you can use the `@tag :tmp_dir` ExUnit tag and do `expected_path = Path.join(tmp_dir, "sentry.map")`, without having to remove the file after the test.

lib/mix/tasks/sentry.package_source_code.ex
```diff
@@ -43,9 +43,34 @@ defmodule Mix.Tasks.Sentry.PackageSourceCode do
   > source-context-related options in compile-time config files (like `config/config.exs`
   > or `config/prod.exs`).
 
+  > #### Building with nix {: .tip}
```
whatyouhide, line 46:
```suggestion
  > #### Building with Nix {: .tip}
```

whatyouhide (changes requested):
The issue with this is that the `priv/sentry.map` path is not public API. So, if allow users to change the output of the map, we should also allow them to load the map from a different path. So, let's also add a config option for this, called `:source_code_map_path`. When not present, we default to the current `priv/sentry.map`.

whatyouhide:
Thanks @razielgn! 💟

Issue #694, opened 2024-02-20
Support for "Cron Monitors"


hkrutzer:
Some SDKs have support for the new Cron Monitoring feature, e.g. https://docs.sentry.io/platforms/javascript/guides/nextjs/crons/. Could this SDK support it as well? I think libraries like Quantum are quite popular. Perhaps it could work with `:telemetry.span`?

whatyouhide:
I will take a look at this 👁️

whatyouhide:
This has been closed in #697, #698, #699.

PR #697, opened 2024-02-25
Introduce manual check-ins for crons

whatyouhide:
First pass at #694. This PR introduces the low-level API, without any integrations.

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -171,7 +171,7 @@ defmodule Sentry do
   > with `:source_code_exclude_patterns`.
   """
 
-  alias Sentry.{Client, Config, Event, LoggerUtils}
+  alias Sentry.{CheckIn, Client, Config, Event, LoggerUtils}
 
   require Logger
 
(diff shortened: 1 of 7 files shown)


sl0thentr0py (approved):

lib/sentry.ex
```diff
@@ -338,6 +338,64 @@ defmodule Sentry do
     end
   end
 
+  @doc """
+  Captures a check-in built with the given `options`.
+
+  Check-ins are used to report the status of a monitor to Sentry. This is used
+  to track the health and progress of **cron jobs**. This function is somewhat
+  low level, and mostly useful when you want to report the status of a cron
+  but you are not using any common library to manage your cron jobs.
```
hkrutzer, line 347: So this is basically meant to always be wrapped, right? As 

```
{:ok, check_in_id} = Sentry.capture_check_in(status: :in_progress, monitor_slug: "genserver-job")
```

will break in e.g. testing code if no DSN is set.

whatyouhide: `capture_check_in/1` will return `:ignored` if DSN is not set, yeah.

PR #698, opened 2024-02-26
Add cron integration for Oban

whatyouhide:
Follow up of #697. The most commonly-used library for job scheduling is definitely Oban, so seems like a reasonable place to start 🙃.

See also #694.


lib/sentry/cron/oban.ex
```diff
@@ -0,0 +1,68 @@
...
+  def attach_telemetry_handler do
+    _ = :telemetry.attach_many(__MODULE__, @events, &__MODULE__.handle_event/4, %{})
+    :ok
+  end
+
+  @spec handle_event([atom()], term(), term(), term()) :: term()
+  def handle_event(event, measurements, metadata, config)
+
+  def handle_event([:oban, :job, :start], _measurements, metadata, _config) do
+    Sentry.capture_check_in(
```
sl0thentr0py, line 20: this is not entirely clear from the docs (I've already pinged people to clarify this several times, sorry..)

but if you want these monitors to show up automatically (we call it ['upserting'](https://docs.sentry.io/platforms/ruby/crons/#upserting-cron-monitors)), you need to add a `monitor_config` with a crontab/interval pulled from the job information. 

If you don't add a `monitor_config`, the user will need to manually add a monitor with a matching `slug` in the product view otherwise the corresponding check-ins will be dropped.

This also implies that this integration should probably only deal with [periodic jobs](https://github.com/sorentwo/oban?tab=readme-ov-file#periodic-jobs) and not all general jobs.

In ruby I also expose a [mixin module](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/cron/monitor_check_ins.rb) that people can add independently to their jobs, we can do something similar (a plug?) that makes sense in the elixir world.

sl0thentr0py (changes requested):

lib/sentry/cron/oban.ex
whatyouhide: Gotcha, nice. I'll think about it, but in the meantime I opened https://github.com/sorentwo/oban/issues/1048 for more context.

whatyouhide: @sl0thentr0py this is ready for re-review. Now it's only reporting check ins for periodic jobs, so we should be good to go.

sl0thentr0py (approved):

PR #701, opened 2024-03-18
Fix wrong flag name

vshev4enko:


whatyouhide:
Nice catch @vshev4enko! 💟

PR #704, opened 2024-03-20
Fix automatic start integrations

vshev4enko:


whatyouhide:
Fantastic catch, thanks @vshev4enko! 💟

PR #705, opened 2024-03-24
Add integrated support for capturing Oban errors

whatyouhide:


sorentwo:
@whatyouhide Once this is released you should come by and add mention of it to the Oban docs 🙂

michallepicki:
Hi @whatyouhide , thank you for this! I have custom instrumentation already and comparing it to the code from this PR it looks more straightforward. I wonder e.g. why did you decide not to use the stacktrace and other fields from `meta` which are exposed for error reporting purposes as Oban documents [here](https://github.com/sorentwo/oban#reporting-errors)? And would you care to explain in what scenarios the stacktrace could be empty and why it's important to fill it in with something?

sorentwo:
@michallepicki there are details about why the stacktrace is empty and other tidbits in this article: https://getoban.pro/articles/enhancing-error-reporting

PR #711, opened 2024-04-04
WIP otel auto setup

sl0thentr0py:


lib/sentry/integrations/opentelemetry.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  alias Sentry.Config
+
+  def setup do
+    if Config.dsn() do
+      Application.put_env(:opentelemetry, :traces_exporter, :otlp)
+      Application.put_env(:opentelemetry_exporter, :otlp_protocol, :http_protobuf)
+      Application.put_env(:opentelemetry_exporter, :otlp_traces_endpoint, Config.spans_endpoint())
+      Application.put_env(:opentelemetry_exporter, :otlp_headers, Config.auth_headers())
+    end
```
yordis, line 12: I honestly believe this is a suboptimal idea.

The OTEL spec expects environment configuration to work correctly, among other things. DevOps wouldn't like this, and troubleshooting will require knowing this code exists instead of, for the most part, checking the `config` directory.
I have already encountered problems with adding or expecting global environment configuration, which has caused significant headaches.

Proper documentation significantly goes a long way when these configurations barely change.

solnic:
@sl0thentr0py I suppose this should be closed in favor of #784?

sl0thentr0py:
yep, no longer relevant

PR #721, opened 2024-04-17
FIx monitor slug in Oban + Quantum integrations

### LINKED ISSUE #720: Cron Monitor integration for Oban
For a particular client, I'm in the process of upgrading their `sentry` client from `8.x` to `10.x`, while I was doing that I noticed the addition of two **great integrations** for Oban. So, I wanted to try those and after doing that, I wanted to share a couple of observations.

The Sentry documentation about the configuration options doesn't specify a minimum required Oban or Oban Pro version.  I think the docs should mention that the cron integration will only work if you have Oban Pro [v0.14 ](https://oban.pro/docs/pro/1.4.3/changelog.html#enhancements-1), or Oban [v.2.17.6](https://github.com/sorentwo/oban/releases/tag/v2.17.6).

Screenshot 2024-04-16 at 1 40 44 PM

Another issue that I found is the current `monitor_slug`. While I see the telemetry handler implementation working and sending the check-ins, it seems that the `monitor_slug` is the raw `job.worker` (e.g., `MyApp.MyDummyWorker`), while the **Sentry UI** doesn't allow you to define such slug, even if the Monitor Cron title is `MyApp.MyDummyWorker`, Sentry lowercase that input and removes the dots, So, there is a mismatch in the config and the monitor doesn't catch the check-ins. I haven't tried creating the Cron Monitor programmatically to see if Sentry allows setting the monitor slug as is sent by the cron integration.

For now, at least in my case, is more important the upgrade from the sentry client from `8.x` to `10.x`. So, I will revisit the cron integration later, but I wanted to start a discussion here.

whatyouhide:
Closes #720.

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -5,7 +5,8 @@ defmodule Sentry.Config do
     oban: [
       type: :keyword_list,
       doc: """
-      Configuration for the [Oban](https://github.com/sorentwo/oban) integration. *Available
+      Configuration for the [Oban](https://github.com/sorentwo/oban) integration. The Oban
+      integration requires at minumum Oban Pro v0.14 or Oban v.2.17.6. *Available
       since v10.2.0*.
       """,
       keys: [
--- a/lib/sentry/integrations/oban/cron.ex
+++ b/lib/sentry/integrations/oban/cron.ex
@@ -72,7 +72,7 @@ defmodule Sentry.Integrations.Oban.Cron do
       [
         check_in_id: "oban-#{job.id}",
         # This is already a binary.
-        monitor_slug: job.worker,
+        monitor_slug: slugify(job.worker),
         monitor_config: [schedule: schedule_opts]
       ]
     else
@@ -98,5 +98,13 @@ defmodule Sentry.Integrations.Oban.Cron do
     duration
     |> System.convert_time_unit(:native, :millisecond)
     |> Kernel./(1000)
+  end
+
+  # MyApp.SomeWorker -> "my-app-some-worker"
+  defp slugify(worker_name) do
+    worker_name
+    |> String.split(".")
+    |> Enum.map(&(&1 |> Macro.underscore() |> String.replace("_", "-")))
+    |> Enum.join("-")
   end
 end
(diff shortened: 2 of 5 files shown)


lib/sentry/integrations/oban/cron.ex
```diff
@@ -99,4 +99,12 @@ defmodule Sentry.Integrations.Oban.Cron do
     |> System.convert_time_unit(:native, :millisecond)
     |> Kernel./(1000)
   end
+
+  # MyApp.SomeWorker -> "my-app-some-worker"
+  defp slugify(worker_name) do
+    worker_name
+    |> String.split(".")
+    |> Enum.map(&(&1 |> Macro.underscore() |> String.replace("_", "-")))
+    |> Enum.join("-")
```
milmazz, line 108:
```suggestion
    worker_name
    |> String.split(".")
    |> Enum.map_join("-", &(&1 |> Macro.underscore() |> String.replace("_", "-")))
    |> String.slice(0, 50)
```

But even this will force me to _edit_ every single Cron Monitor that I created, why? Because the Sentry form to add a monitor doesn't allow to define the monitor slug right away, so, if you want to adapt that slug, you have to edit. Also, I just tried and the web form only allow 50 characters as the maximum.

So, the slug by default is something like:

```elixir
"Core.MyDummyWorker" |> String.duplicate(5) |> String.downcase() |> String.replace(".", "") |> String.slice(0, 50)
```

lib/sentry/config.ex
```diff
@@ -5,7 +5,8 @@ defmodule Sentry.Config do
     oban: [
       type: :keyword_list,
       doc: """
-      Configuration for the [Oban](https://github.com/sorentwo/oban) integration. *Available
+      Configuration for the [Oban](https://github.com/sorentwo/oban) integration. The Oban
+      integration requires at minumum Oban Pro v0.14 or Oban v.2.17.6. *Available
```
milmazz, line 9: :bowtie:

lib/sentry/integrations/oban/cron.ex
whatyouhide: Yeah there's an issue here because [the Sentry check-in API](https://develop.sentry.dev/sdk/check-ins/) doesn't allow to programmatically change the *name*, only the monitor slug.

Great catch on the character limit though.

whatyouhide:
@milmazz this is a first pass to get things working, but I suspect we'll want some level of customization here anyway. I don't want to figure out the API now though, I just want to get this fixed for now so that we have time to give it some more thought.

milmazz (approved):
After editing the monitor slug via Sentry UI, it's working 🎉

PR #742, opened 2024-06-27
Fix `on_mount` LiveView hook when given `:not_mounted_at_router`

### LINKED ISSUE #738: Error using `on_mount Sentry.LiveViewHook` for child LiveView
### Environment

Elixir 1.16.3
Phoenix 1.7.12
Phoenix LiveView 0.20.14
Sentry 10.6.1

### Steps to Reproduce

I added the `on_mount Sentry.LiveViewHook` to `MyAppWeb.live_view` so it applies to all my LiveViews. I also have a LiveView that does `use MyAppWeb, :live_view` that I also embed in a dead view via `live_render`.

Repro one-file: https://gist.github.com/benregn/2e64c494ac8db44c0d3aa48660182d93

### Expected Result

On navigating to the page with the embedded LiveView, I expect to see no errors.

### Actual Result

On navigating to the page with the embedded LiveView, I see this error:

`[error] Sentry.LiveView.on_mount hook errored out: ** (RuntimeError) cannot attach hook with id Sentry.LiveViewHook on :handle_params because the view was not mounted at the router with the live/3 macro`

### Workaround

I made this wrapper to alleviate the problem in my app:

```elixir
defmodule SentryLiveViewHookWrapper do
  def on_mount(_arg, :not_mounted_at_router, session, socket) do
    {:cont, socket}
  end
  
  def on_mount(:default, params, session, socket),
    do: Sentry.LiveViewHook.on_mount(:default, params, session, socket)
end
```

savhappy:
Closes #738.

* Added new on_mount for params atom :not_mounted_at_router and bypasses logging error when a liveview process is a child of a non-live controller/view
* added test - checks to ensure logger has no errors in metadata

### DIFF
--- a/test/sentry/live_view_hook_test.exs
+++ b/test/sentry/live_view_hook_test.exs
@@ -31,11 +31,44 @@ defmodule SentryTest.LiveComponent do
   end
 end
 
+defmodule SentryTest.DeadLive do
+  use Phoenix.LiveView
+  on_mount Sentry.LiveViewHook
+
+  @impl true
+  def render(assigns) do
+    ~H"""
+    <div>I'm being live_rendered!</div>
+    """
+  end
+end
+
+defmodule SentryTest.ErrorView do
+  def render(template, _), do: Phoenix.Controller.status_message_from_template(template)
+end
+
+defmodule SentryTest.PageController do
+  use Phoenix.Controller
+  use Phoenix.Component
+
+  def page(conn, _params) do
+    assigns = %{conn: conn}
+
+    rendered = ~H"""
+    I'm a controller! <br><%= live_render(@conn, SentryTest.DeadLive, id: "live-render-id") %>
+    """
+
+    content = Phoenix.HTML.Safe.to_iodata(rendered)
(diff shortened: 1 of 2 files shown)


test/sentry/live_view_hook_test.exs
```diff
@@ -31,11 +31,44 @@ defmodule SentryTest.LiveComponent do
...
+  def render(assigns) do
+    ~H"""
+    <div>I'm being live_rendered!</div>
+    """
+  end
+end
+
+defmodule SentryTest.ErrorView do
+  def render(template, _), do: Phoenix.Controller.status_message_from_template(template)
+end
```
whatyouhide, line 48: Do you need this considering you're just returning some text with `text/2`?

whatyouhide (commented):
Fantastic! Just left one comment but looks great.

test/sentry/live_view_hook_test.exs
savhappy: Nope! Removed!

PR #745, opened 2024-07-01
Add new client error exception

savhappy:
-add client error module to handle Sentry client errors

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -174,7 +174,7 @@ defmodule Sentry do
   > with `:source_code_exclude_patterns`.
   """
 
-  alias Sentry.{CheckIn, Client, Config, Event, LoggerUtils}
+  alias Sentry.{CheckIn, Client, ClientError, Config, Event, LoggerUtils}
 
   require Logger
 
@@ -212,10 +212,15 @@ defmodule Sentry do
 
   @type send_result() ::
           {:ok, event_or_envelope_id :: String.t()}
-          | {:error, term()}
+          | {:error, Sentry.error()}
           | :ignored
           | :unsampled
           | :excluded
+
+  @typedoc """
+  An error that is returned from Sentry Client.
+  """
+  @type error :: ClientError.t()
 
   @doc """
   Parses and submits an exception to Sentry.
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -268,7 +268,7 @@ defmodule Sentry do
   ## Interpolation (since v10.1.0)
 
   The `message` argument supports interpolation. You can pass a string with formatting
-  markers as `%s`, ant then pass in the `:interpolation_parameters` option as a list
+  markers as `%s`, and then pass in the `:interpolation_parameters` option as a list
   of positional parameters to interpolate. For example:
 
       Sentry.capture_message("Error with user %s", interpolation_parameters: ["John"])
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -212,15 +212,10 @@ defmodule Sentry do
 
   @type send_result() ::
           {:ok, event_or_envelope_id :: String.t()}
-          | {:error, Sentry.error()}
+          | {:error, ClientError.t()}
           | :ignored
           | :unsampled
           | :excluded
-
-  @typedoc """
-  An error that is returned from Sentry Client.
-  """
-  @type error :: ClientError.t()
 
   @doc """
   Parses and submits an exception to Sentry.
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -5,7 +5,17 @@ defmodule Sentry.Client do
   # and sampling.
   # See https://develop.sentry.dev/sdk/unified-api/#client.
 
-  alias Sentry.{CheckIn, Config, Dedupe, Envelope, Event, Interfaces, LoggerUtils, Transport}
+  alias Sentry.{
+    CheckIn,
+    Config,
+    Dedupe,
+    Envelope,
+    Event,
+    Interfaces,
+    LoggerUtils,
+    Transport,
+    ClientError
+  }
 
   require Logger
 
@@ -139,8 +149,8 @@ defmodule Sentry.Client do
       :excluded ->
         :excluded
 
-      {:error, _reason} ->
-        result
+      {:error, reason} ->
+        {:error, ClientError.new(reason)}
     end
   end
 
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -112,7 +112,7 @@ defmodule Sentry.Client do
   # See: https://develop.sentry.dev/sdk/unified-api/#event-pipeline
   @spec send_event(Event.t(), keyword()) ::
           {:ok, event_id :: String.t()}
-          | {:error, Sentry.error()}
+          | {:error, ClientError.t()}
           | :unsampled
           | :excluded
   def send_event(%Event{} = event, opts) when is_list(opts) do
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -7,14 +7,14 @@ defmodule Sentry.Client do
 
   alias Sentry.{
     CheckIn,
+    ClientError,
     Config,
     Dedupe,
     Envelope,
     Event,
     Interfaces,
     LoggerUtils,
-    Transport,
-    ClientError
+    Transport
   }
 
   require Logger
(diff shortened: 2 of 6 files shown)


savhappy:
This does not change anything in the logs. Should the logging be changed to match the correct message response coming from client error module?

lib/sentry/client.ex
```diff
@@ -5,7 +5,17 @@ defmodule Sentry.Client do
...
+  alias Sentry.{
+    CheckIn,
+    Config,
+    Dedupe,
+    Envelope,
+    Event,
+    Interfaces,
+    LoggerUtils,
+    Transport,
+    ClientError
```
whatyouhide, line 17: This is a silly nit but these were alpha-sorted, can we move this up? 🙃

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
+defmodule Sentry.ClientError do
+  @moduledoc """
+  An exception struct that represents an error returned by Sentry.
```
whatyouhide, line 3: What do you think about

```suggestion
  An exception struct that represents an error returned by Sentry when
  reporting an error or a message.
```

?

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  This struct is designed to manage and handle errors originating from operations
+  in the client application. The `:reason` field signifies the cause of the error
```
whatyouhide, line 6: I think "client application" can be confusing?

```suggestion
  in the Sentry client. The `:reason` field signifies the cause of the error
```

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+  as an atom or tuple.
+
+  To raise instances of this exception, you can use `Kernel.raise/1`. When crafting
+  formatted error messages for purposes such as logging or presentation, consider
+  leveraging `Exception.message/1`.
```
whatyouhide, line 11: Fantastique, perfectly worded.

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  @doc """
+  The exception struct for a Sentry error.
+  """
+  defexception [:reason]
+
+  @typedoc """
+  The type for a Sentry error exception.
+  """
+  @type t() :: %__MODULE__{reason: term}
```
whatyouhide, line 24: Nit, but can we introduce

```elixir
@type reason() :: :too_many_retries | ...
```

and use that in `new/1` as well, instead of `term/0`?

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  defp format({:request_failure, reason}) when is_atom(reason) do
+    lookup_reason =
+      case :inet.format_error(reason) do
+        ~c"unknown POSIX error" -> inspect(reason)
+        formatted -> List.to_string(formatted)
+      end
+
+    "#{lookup_reason}"
```
whatyouhide, line 56: `lookup_reason` is already a string, you can just do

```suggestion
    case :inet.format_error(reason) do
      ~c"unknown POSIX error" -> inspect(reason)
      formatted -> List.to_string(formatted)
    end
```

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  defp format({:request_failure, reason}) do
+    "#{inspect(reason)}"
```
whatyouhide, line 60: No need for interpolation:

```suggestion
    inspect(reason)
```

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  defp format({kind, data, stacktrace}) do
+    "Exception: #{inspect(kind)} with data: #{inspect(data)} and stacktrace: #{inspect(stacktrace)}"
```
whatyouhide, line 64: You can leverage `Exception.format/3` here. I'd do this:

```suggestion
    """
    Sentry failed to report event due to an unexpected error:
    
    #{Exception.format(kind, data, stacktrace)}\
    """
```

Thoughts?

test/sentry/client_error_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule Sentry.ClientErrorTest do
+  use Sentry.Case
+  alias Sentry.ClientError
+
+  describe "message/1" do
+    test "With atom - returns message " do
```
whatyouhide, line 6: Supa dupa nit: let's start tests with a lowercase and no trailing space

```suggestion
    test "with atom - returns message" do
```

in the whole file here.

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+  The exception struct for a Sentry error.
+  """
+  defexception [:reason]
+
+  @typedoc """
+  The type for a Sentry error exception.
+  """
+  @type t() :: %__MODULE__{reason: term}
+
+  @doc false
```
whatyouhide, line 26: Nice 👏

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+  @spec new(term) :: t
+  def new(reason) do
+    %__MODULE__{reason: reason}
+  end
+
+  @impl true
+  def message(%__MODULE__{reason: reason}) do
+    "Request failure reason: #{format(reason)}"
```
whatyouhide, line 34: Nit, but usually in Elixir we start exception messages with lowercase letters.

It would be

```suggestion
    "request failure reason: #{format(reason)}"
```

here. Can you fix that up in this file?

lib/sentry/client_error.ex
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  defp format(:too_many_retries) do
+    "Sentry responded with status 429 - Too Many Requests"
+  end
+
+  defp format({:invalid_json, reason}) do
+    "Invalid JSON -> #{inspect(reason)}"
```
whatyouhide, line 42: This should be a `Jason.DecodeError`, right? So we can do

```suggestion
    "Invalid JSON: #{Exception.message(reason)}"
```

whatyouhide (commented):
Left A BUNCH of comments, sorry! Fantastic work though, this is exactly what I had in mind. 🙃

lib/sentry/client_error.ex
savhappy: Yep yep!

lib/sentry/client_error.ex
savhappy: hmmm actually I think I'm wrong here. This exception get returned on the :request_failure tuple, not with :invalid_json. After following the path it looks like the error produced for this atom is one of these:

`  def __validate_json_library__(nil) do
    {:error, "nil is not a valid value for the :json_library option"}
  end

  def __validate_json_library__(mod) when is_atom(mod) do
    try do
      with {:ok, %{}} <- mod.decode("{}"),
           {:ok, "{}"} <- mod.encode(%{}) do
        {:ok, mod}
      else
        _ ->
          {:error,
           "configured :json_library #{inspect(mod)} does not implement decode/1 and encode/1"}
      end
    rescue
      UndefinedFunctionError ->
        {:error,
         """
         configured :json_library #{inspect(mod)} is not available or does not implement decode/1 and encode/1.
         Do you need to add #{inspect(mod)} to your mix.exs?
         """}
    end
  end

  def __validate_json_library__(other) do
    {:error, "expected :json_library to be a module, got: #{inspect(other)}"}
  end`

Is that correct?

whatyouhide: I don't think that's it, that's something we do when the `:sentry` application starts to validate that the JSON library chosen by the user quacks like a JSON library indeed.

IIRC we return `:invalid_json` when the response from the Sentry server is not valid JSON, don't we?

https://github.com/getsentry/sentry-elixir/blob/fba608c72a403cb0b21e9ebd8cd2cd313a1e3ecb/lib/sentry/transport.ex#L23-L30

lib/sentry/client_error.ex
savhappy: like this:

```elixir
  @type t() :: %__MODULE__{
          reason:
            :too_many_retries
            | {:invalid_json, Exception.t() / 0}
            | {:request_failure, String.t()}
            | {:request_failure, atom}
            | {:request_failure, {atom(), term(), [term()]}}
        }```

Comment (whatyouhide) [reply lib/sentry/client_error.ex:24]: yep!

File: lib/sentry.ex [hunk 13]
```diff
@@ -212,11 +212,16 @@ defmodule Sentry do
...
-          | {:error, term()}
+          | {:error, Sentry.error()}
           | :ignored
           | :unsampled
           | :excluded
 
+  @typedoc """
+  An error that is returned from Sentry Client.
+  """
+  @type error :: ClientError.t()
```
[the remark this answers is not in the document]
savhappy: I was thinking preemptive in case we introduce more error types but removed this. You're right, it's redundant.

whatyouhide (approved):

whatyouhide:
Yeeehaaaw

PR #747, opened 2024-07-04
Update links related to logger

HarshBalyan:
Updated links are one of the following.
- broken 
- works after redirection
- has updated content


whatyouhide (approved):
Sweet 🍬

PR #748, opened 2024-07-04
Add `http_resp` to client error

savhappy:
-add :server_error as new error path

### DIFF
--- a/lib/sentry/client_error.ex
+++ b/lib/sentry/client_error.ex
@@ -17,15 +17,20 @@ defmodule Sentry.ClientError do
   @doc """
   The exception struct for a Sentry error.
   """
-  defexception [:reason]
+  defexception [:reason, :http_response]
 
   @typedoc """
   The type for a Sentry error exception.
   """
-  @type t :: %__MODULE__{reason: reason()}
+  @type t :: %__MODULE__{
+          reason: reason(),
+          http_response:
+            nil | {status :: 100..199, headers :: [{String.t(), String.t()}], body :: binary()}
+        }
 
   @type reason() ::
           :too_many_retries
+          | :server_error
           | {:invalid_json, Exception.t()}
           | {:request_failure, String.t()}
           | {:request_failure, atom}
@@ -33,14 +38,22 @@ defmodule Sentry.ClientError do
           | {atom(), term(), [term()]}
 
   @doc false
-  @spec new(reason()) :: t
-  def new(reason) do
-    %__MODULE__{reason: reason}
+  @spec new(
+          reason(),
+          nil | {status :: 100..599, headers :: [{String.t(), String.t()}], body :: binary()}
+        ) :: t
+  def new(reason, http_response \\ nil) do
+    %__MODULE__{reason: reason, http_response: http_response}
   end
 
   @impl true
-  def message(%__MODULE__{reason: reason}) do
+  def message(%__MODULE__{reason: reason, http_response: http_response})
+      when is_nil(http_response) do
     "request failure reason: #{format(reason)}"
+  end
+
+  def message(%__MODULE__{reason: reason, http_response: http_response}) do
+    "request failure reason: #{format(reason, http_response)}"
   end
 
   defp format(:too_many_retries) do
@@ -71,6 +84,15 @@ defmodule Sentry.ClientError do
     Sentry failed to report event due to an unexpected error:
 
     #{Exception.format(kind, data, stacktrace)}\
+    """
+  end
+
+  defp format(:server_error, {status, headers, body}) do
+    """
+    Sentry failed to report the event due to a server error.\n\
+    HTTP Status: #{status}\n\
+    Response Headers: #{inspect(headers)}\n\
+    Response Body: #{inspect(body)}
     """
   end
 end
--- a/lib/sentry/client_error.ex
+++ b/lib/sentry/client_error.ex
@@ -43,6 +43,7 @@ defmodule Sentry.ClientError do
     %__MODULE__{reason: reason}
   end
 
+  @doc false
   @spec server_error(
           status :: 100..599,
           headers ::
--- a/lib/sentry/client_error.ex
+++ b/lib/sentry/client_error.ex
@@ -31,6 +31,8 @@ defmodule Sentry.ClientError do
   @type reason() ::
           :too_many_retries
           | :server_error
+          | {:server_error,
+             {status :: 100..199, headers :: [{String.t(), String.t()}], body :: binary()}}
           | {:invalid_json, Exception.t()}
           | {:request_failure, String.t()}
           | {:request_failure, atom}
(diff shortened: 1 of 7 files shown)


lib/sentry/client_error.ex
```diff
@@ -17,15 +17,22 @@ defmodule Sentry.ClientError do
...
-  @type t :: %__MODULE__{reason: reason()}
+  @type t :: %__MODULE__{
+          reason: reason(),
+          http_response:
+            nil | {status :: 100..199, headers :: [{String.t(), String.t()}], body :: binary()}
+        }
 
   @type reason() ::
           :too_many_retries
+          | :server_error
```
savhappy, line 33: This felt convoluted but needed to pass :server_error by itself as a type so we don't return the headers twice when we call `new`

lib/sentry/client_error.ex
```diff
@@ -17,15 +17,22 @@ defmodule Sentry.ClientError do
...
+          | {:server_error,
+             {status :: 100..199, headers :: [{String.t(), String.t()}], body :: binary()}}
```
savhappy, line 35: Shouldn't the range be to 599?

lib/sentry/transport.ex
```diff
@@ -73,12 +76,12 @@ defmodule Sentry.Transport do
...
         {:retry_after, delay_ms}
 
-      {:ok, status, headers, _body} ->
+      {:ok, status, headers, body} ->
         error_header =
           :proplists.get_value("X-Sentry-Error", headers, nil) ||
             :proplists.get_value("x-sentry-error", headers, nil) || ""
 
-        {:error, "Received #{status} from Sentry server: #{error_header}"}
+        {:error, {:server_error, {status, error_header, body}}}
```
savhappy, line 84: This is the biggest question - do we want to return only the headers that match x-sentry-error or all headers @whatyouhide?

lib/sentry/client_error.ex
whatyouhide: Add `server_error/3`, it's gonna make things simpler 🙃 `new/*` are private anyway so we don't need to care too much, we can always change later.

```elixir
def server_error(status, headers, body) do
  %__MODULE__{reason: :server_error, http_response: {status, headers, body}}
end
```

lib/sentry/client_error.ex
whatyouhide: Yes sorry typo on my end 🙃

lib/sentry/client_error.ex
```diff
@@ -34,15 +41,24 @@ defmodule Sentry.ClientError do
 
   @doc false
   @spec new(reason()) :: t
+  def new({:server_error, http_response}) do
+    %__MODULE__{reason: :server_error, http_response: http_response}
+  end
+
   def new(reason) do
-    %__MODULE__{reason: reason}
+    %__MODULE__{reason: reason, http_response: nil}
```
whatyouhide, line 49: It's `nil` by default, you can leave this alone:

```suggestion
    %__MODULE__{reason: reason}
```

lib/sentry/transport.ex
whatyouhide: All headers, it's pretty much free to return all of them and I can't think of a good reason to only return `x-sentry-*` headers.

test/sentry/client_test.exs
```diff
@@ -226,6 +226,34 @@ defmodule Sentry.ClientTest do
       assert log =~ "Received 400 from Sentry server: Rate limiting."
     end
 
+    test "returns an error if Sentry server responds with error status", %{bypass: bypass} do
```
savhappy, line 229: Added two new test for the newly returned errors for send_event

lib/sentry/transport.ex
```diff
@@ -49,6 +49,14 @@ defmodule Sentry.Transport do
         Process.sleep(sleep_interval)
         post_envelope_with_retries(client, endpoint, headers, payload, retries_left)
 
+      {:error, _status, _headers, _body} when retries_left != [] ->
```
savhappy, line 52: Pattern match causes issues on retry. This was my solution -> open to another idea if you have one @whatyouhide

whatyouhide: > Pattern match causes issues on retry

What issues? You have control over what `request/4` returns so you can return `{:error, {status, headers, body}}` from there and let this fall into the `{:error, reason}` pattern match?

savhappy: When I pass as a tuple it catches in this new error block for a few tests rather than :request_failure. But after thinking about it. I think returning the full error is probably what we want anyhow. Thx for input!! 


elixir
```
 test post_envelope/2 returns an error if the JSON library crashes when decoding the response (Sentry.TransportTest)
     test/sentry/transport_test.exs:121
     match (=) failed
     code:  assert {:error, {:request_failure, reason}} =
              Transport.post_envelope(envelope, HackneyClient, _retries = [])
     left:  {:error, {:request_failure, reason}}
     right: {
              :error,
              {
                :error,
                %RuntimeError{message: "I'm a really bad JSON library"},
                [{Sentry.TransportTest.CrashingJSONLibrary, :decode, 1, [file: ~c"test/sentry/transport_test.exs", line: 129, error_info: %{module: Exception}]}, {Sentry.Transport, :request, 4, [file: ~c"lib/sentry/transport.ex", line: 67]}, {Sentry.Transport, :post_envelope_with_retries, 5, [file: ~c"lib/sentry/transport.ex", line: 34]}, {Sentry.TransportTest, :"test post_envelope/2 returns an error if the JSON library crashes when decoding the response", 1, [file: ~c"test/sentry/transport_test.exs", line: 139]}, {ExUnit.Runner, :exec_test, 2, [file: ~c"lib/ex_unit/runner.ex", line: 472]}, {:timer, :tc, 2, [file: ~c"timer.erl", line: 270]}, {ExUnit.Runner, :"-spawn_test_monitor/4-fun-1-", 6, [file: ~c"lib/ex_unit/runner.ex", line: 394]}]
              }
            }
```

lib/sentry/client_error.ex
```diff
@@ -38,11 +43,26 @@ defmodule Sentry.ClientError do
     %__MODULE__{reason: reason}
   end
 
+  @spec server_error(
```
whatyouhide, line 46:
```suggestion
  @doc false
  @spec server_error(
```

test/sentry/client_error_test.exs
```diff
@@ -5,51 +5,53 @@ defmodule Sentry.ClientErrorTest do
...
+      {status, headers, body} =
+        {400, "Rate limiting.", "{}"}
+
+      assert "request failure reason: Sentry failed to report the event due to a server error.\nHTTP Status: 400\nResponse Headers: \"Rate limiting.\"\nResponse Body: \"{}\"\n" =
+               ClientError.server_error(status, headers, body) |> ClientError.message()
+    end
+  end
+
+  defp result_msg(reason) do
+    ClientError.new(reason) |> ClientError.message()
```
whatyouhide, line 55:
```suggestion
    reason |> ClientError.new() |> ClientError.message()
```

whatyouhide (approved):

PR #750, opened 2024-07-17
Cron integration bug

savhappy:
-added simple ets table to map cron id to UUID as required by Sentry


savhappy:
@whatyouhide I need to cleanup tests and add typepecs but is this what you had in mind? I left some pseudo-code that I'll remove after draft.

lib/sentry/integrations/oban/cron.ex
```diff
@@ -68,9 +70,19 @@ defmodule Sentry.Integrations.Oban.Cron do
...
+    # lookup job.id -> if exists check-in-id is this uuid. Othereise store new key and create UUID value
     if schedule_opts = schedule_opts(job) do
+      {:ok, id} =
+        case Mapping.lookup(:cron, "oban-#{job.id}") do
+          {:ok, value} ->
+            {:ok, value}
+
+          :error ->
+            Mapping.insert(:cron, "oban-#{job.id}", UUID.uuid4_hex())
+        end
```
whatyouhide, line 82: Hrm, I’m thinking this is not exactly what we want. I think that we want:

  * that if this is a "start" event, then the ID **must not** be there, so we want to insert it and
  * that if this is a "stop" or "exception" event, then the ID **must** be there, so we expect it.

Does that make sense?

lib/sentry/envelope.ex
```diff
@@ -29,7 +29,7 @@ defmodule Sentry.Envelope do
   @spec from_check_in(CheckIn.t()) :: t()
   def from_check_in(%CheckIn{} = check_in) do
     %__MODULE__{
-      event_id: UUID.uuid4_hex(),
+      event_id: check_in.check_in_id,
```
whatyouhide, line 32: Fantastique!

lib/sentry/mapping.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Sentry.Mapping do
+  @moduledoc false
+
+  def new() do
+    :ets.new(:cron, [:named_table, :set, :public, read_concurrency: true])
+  end
+
+  def insert(table, key, value) do
+    :ets.insert(table, {key, value})
```
whatyouhide, line 9: The table is always gonna be called the same. You could extract the name to a module attribute:

```elixir
@table :cron
```

and then you won't have to pass it around in fucntions.

lib/sentry/application.ex
```diff
@@ -48,16 +52,18 @@ defmodule Sentry.Application do
   end
 
   defp start_integrations(config) do
-    if config[:oban][:cron][:enabled] do
-      Sentry.Integrations.Oban.Cron.attach_telemetry_handler()
-    end
+    if Code.ensure_loaded?(Sentry.Integrations.CheckInIDMappings) do
```
whatyouhide, line 55: Wait, why would this not be loaded? This will always be loaded. We can also always start it even if integrations are not started for the time being, it's very cheap to run a GenServer 🙃

lib/sentry/integrations/check_in_id_mappings.ex
```diff
@@ -0,0 +1,33 @@
+defmodule Sentry.Integrations.CheckInIDMappings do
+  @moduledoc false
+
+  use GenServer
+
+  @table :cron
```
whatyouhide, line 6: Just a nit to avoid a very-common name that could be in use by users or other applications.

```suggestion
  @table :sentry_cron_mappings
```

lib/sentry/integrations/check_in_id_mappings.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link([] = _opts) do
+    GenServer.start_link(__MODULE__, nil, name: __MODULE__)
+  end
+
+  @spec lookup(String.t()) :: {:ok, <<_::256>>}
+  def lookup(key) do
```
whatyouhide, line 14: Nit, but can we call this something like `lookup_or_insert_new` to make it clear that this'll generate a new UUID if needed?

Also, the return value can just be `UUID.t()` here, no need for the `{:ok, _}` tuple either.

whatyouhide (commented):
This is looking fantastic! I've left just a couple more comments and then we're good to go. 🎉

lib/sentry/application.ex
savhappy: Do you think it's unnecessary to make sure the table was created?

savhappy:
@whatyouhide should probably add a full start/stop test to quantum? If so please wait to merge (:

lib/sentry/application.ex
whatyouhide: 1. `:ets.new/2` doesn't really fail, so the table will pretty much always be created.
2. That's not what `Code.ensure_loaded?/1` does: it only checks that the bytecode for the module is loaded in the VM and that the module is available, **not** that the GenServer was successfully started.
3. If the supervisor started without issues, then by definition the GenServer also started (at least the first time) without issues.

whatyouhide:
@savhappy yep let's do that test~

whatyouhide:
Fantastic work @savhappy 💟

PR #751, opened 2024-07-17
Update documentation to remove `if`

BobbyMcWho:


whatyouhide:
Oooph, good catch! Thanks for fixing this up 💟

Issue #757, opened 2024-07-29
Selective reporting on Oban errors


Rodeoclash:
Currently failing Oban jobs will always report to Sentry, that is, any job that raises and exception or returns an `{:error, "reason"}` tuple.

In some circumstances we have jobs that should fail with expected errors. For example, the delivery of our webhooks can sometimes fail with an error like `%Req.TransportError{reason: :econnrefused}` - this indicates that the recipient of the webhook has misconfigured their handler. In this scenario, we *don't* want to alert Sentry but we do want the job to fail (and eventually retry as the user will be notified of the failing webhook and can then fix the problem preventing delivery).

I'm not quite sure what the solution to this should look like, it could be something like:

1. Error reporting on Oban jobs can be selectively applied.
2. Error reporting can ignore certain error responses to jobs (this is my preference as I would like to know about other *unexpected* errors!). I imagine this taking the form of a list of error tuples that when returned from the job, don't trigger the error reporting.

I think this can currently be worked around by using a combination of the snooze response and manually tracking the attempts in the job, but this would be a workaround, the ability to selectively report errors somehow would solve this.

whatyouhide:
@Rodeoclash thanks for the feature suggestion!

One approach you can always take is go for an [event filter](https://hexdocs.pm/sentry/10.6.2/Sentry.html#module-event-callbacks), which would allow you to configure a global filter for events reported to Sentry, with any custom logic you desire.

@sorentwo pinging you here though in case you have good ideas on how to approach this on the Oban side?

Rodeoclash:
I was pointed to the event_filter feature after I raised this by a colleague, I'll dig into that before I bother you again 🙂

whatyouhide:
Nah no worries, I think for Oban we might do something more ergonomic because I also encountered the same "annoyance" of returning `{:error, ...}` and not wanting it to be reported much.

sorentwo:
There's an [old recipe](https://hexdocs.pm/oban/expected-failures.html) that describes a way to do this using telemetry and protocols. However, that's not useful when Sentry is doing the reporting automatically.

One thing you can consider is that all error tuples are wrapped in an`Oban.PerformError`. If you filter out that particular error you'd effectively silence failures from manual errors while retaining real exceptions.

whatyouhide:
@sorentwo oh, if the job crashes (exit/throw/raise) isn't that wrapped as an `Oban.PerformError` as well?

sorentwo:
[...]

PR #759, opened 2024-08-01
Integration - Oban - max attempts flag

oliver-kriska:
`max_attempts_only` config flag will allow to capture exception only in case max attempts were reached and Oban Job will not be retried

[#757](https://github.com/getsentry/sentry-elixir/issues/757)

### DIFF
--- a/lib/sentry/application.ex
+++ b/lib/sentry/application.ex
@@ -57,7 +57,7 @@ defmodule Sentry.Application do
     end
 
     if config[:oban][:capture_errors] do
-      Sentry.Integrations.Oban.ErrorReporter.attach()
+      Sentry.Integrations.Oban.ErrorReporter.attach(config[:oban][:max_attempts_only])
     end
 
     if config[:quantum][:cron][:enabled] do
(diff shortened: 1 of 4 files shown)


oliver-kriska:
Maybe it will be better solution based on state `discarded` than on attempts.

whatyouhide:
Instead of this approach, I would rather add an `:integration_meta` key to the `Sentry.Event` struct. It would not be reported to the Sentry server, but it would be documented and passed over to `before_send` filters. In there, the Oban integration could store:

```elixir
integeration_meta: %{
  oban: %{job_meta: ...}
}
```

@savhappy do you want to work on this?

savhappy:
Question for  you, @whatyouhide. For the `Sentry.Event` approach -> Do we want users to be able to set a flag still for max_attempts specifically? And would we add the exclude exception filter as a default in the `DefaultEventFilter` module?

whatyouhide:
@savhappy so we'd do this:

  - We add `integration_meta` as a possible key to `Sentry.Event`
  - We update the Oban integration to insert `%{oban: %{job_meta: ...}}` under `integration_meta` whenever reporting an event from a job.
  - We document these changes.

That's it. Then, users can deal with their own logic for filtering out events by looking at this integration metadata and deciding whether the Oban error can be skipped due to attempts and stuff like that.

whatyouhide:
Closing in favor of #771 which implements the generic `integration_meta` mechanism.

PR #764, opened 2024-08-05
Periodically purge check-ins ETS table

### LINKED ISSUE #762: Periodically purge check-ins ETS table
> @nsweeting yep, you're right! I'll add some sweep functionality that clears the tables. 
> 
> @whatyouhide, is the best approach to clear the table -> remove any check-in ids that have a 'stop' or 'exception' event? or something like `memsup` to check the available RAM and then clear based on the oldest entries?
> 
> I'm going to reopen this to address what @nsweeting mentions. 

 _Originally posted by @savhappy in [#749](https://github.com/getsentry/sentry-elixir/issues/749#issuecomment-2264149450)_

Let's go with the GenServer that owns the table periodically sweeping it. Every ~30 seconds, it can send a message to itself, go through the all table and use `ets:select_delete` or something like that to remove all check-ins that are older than, say, 10 minutes. If needed we can add a timestamp to the check ins, I don't recall if we do that already.

savhappy:
-add sweep functionality to CheckInIDMappings ets table to clear out stale jobs
-add configurable option to corn integrations (TTL) for long standing jobs
-closes #762

### DIFF
--- a/lib/sentry/application.ex
+++ b/lib/sentry/application.ex
@@ -19,11 +19,13 @@ defmodule Sentry.Application do
         []
       end
 
+    integrations = Keyword.fetch!(config, :integrations)
+
     children =
       [
         {Registry, keys: :unique, name: Sentry.Transport.SenderRegistry},
         Sentry.Dedupe,
-        Sentry.Integrations.CheckInIDMappings
+        {Sentry.Integrations.CheckInIDMappings, integrations[:ttl]}
       ] ++
         maybe_http_client_spec ++
         [Sentry.Transport.SenderPool]
@@ -33,7 +35,7 @@ defmodule Sentry.Application do
 
     with {:ok, pid} <-
            Supervisor.start_link(children, strategy: :one_for_one, name: Sentry.Supervisor) do
-      start_integrations(Keyword.fetch!(config, :integrations))
+      start_integrations(integrations)
       {:ok, pid}
     end
   end
--- a/lib/sentry/application.ex
+++ b/lib/sentry/application.ex
@@ -25,7 +25,8 @@ defmodule Sentry.Application do
       [
         {Registry, keys: :unique, name: Sentry.Transport.SenderRegistry},
         Sentry.Dedupe,
-        {Sentry.Integrations.CheckInIDMappings, integrations[:ttl]}
+        {Sentry.Integrations.CheckInIDMappings,
+         [max_expected_check_in_time: integrations_config[:max_expected_check_in_time]]}
       ] ++
         maybe_http_client_spec ++
         [Sentry.Transport.SenderPool]
@@ -35,7 +36,7 @@ defmodule Sentry.Application do
 
     with {:ok, pid} <-
            Supervisor.start_link(children, strategy: :one_for_one, name: Sentry.Supervisor) do
-      start_integrations(integrations)
+      start_integrations(integrations_config)
       {:ok, pid}
     end
   end
--- a/lib/sentry/application.ex
+++ b/lib/sentry/application.ex
@@ -26,7 +26,10 @@ defmodule Sentry.Application do
         {Registry, keys: :unique, name: Sentry.Transport.SenderRegistry},
         Sentry.Dedupe,
         {Sentry.Integrations.CheckInIDMappings,
-         [max_expected_check_in_time: integrations_config[:max_expected_check_in_time]]}
+         [
+           max_expected_check_in_time:
+             Keyword.fetch!(integrations_config, :max_expected_check_in_time)
+         ]}
       ] ++
         maybe_http_client_spec ++
         [Sentry.Transport.SenderPool]
--- a/lib/sentry/application.ex
+++ b/lib/sentry/application.ex
@@ -19,7 +19,7 @@ defmodule Sentry.Application do
         []
       end
 
-    integrations = Keyword.fetch!(config, :integrations)
+    integrations_config = Keyword.fetch!(config, :integrations)
 
     children =
       [
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -2,6 +2,16 @@ defmodule Sentry.Config do
   @moduledoc false
 
   integrations_schema = [
+    ttl: [
+      type: :integer,
+      default: 600_000,
+      doc: """
+      The time in milliseconds that a check-in id will live after it has been created.
+
+      For long running cron jobs, set ttl (time to live) to longest running job time in milliseconds.
+      This avoids jobs being sweeped from table before they've been completed. Defaults to ten minutes.
+      """
+    ],
     oban: [
       type: :keyword_list,
       doc: """
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -14,6 +14,8 @@ defmodule Sentry.Config do
       if a check-in end event is reported after the specified `max_expected_check_in_time`,
       the SDK will not report it. This behavior helps manage resource usage effectively while still
       providing necessary tracking for your jobs.
+      
+      *Available since 10.6.3*.
       """
     ],
     oban: [
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -6,12 +6,14 @@ defmodule Sentry.Config do
       type: :integer,
       default: 600_000,
       doc: """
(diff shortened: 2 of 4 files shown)


lib/sentry/application.ex
```diff
@@ -19,11 +19,13 @@ defmodule Sentry.Application do
         []
       end
 
+    integrations = Keyword.fetch!(config, :integrations)
```
whatyouhide, line 22:
```suggestion
    integrations_config = Keyword.fetch!(config, :integrations)
```

lib/sentry/config.ex
```diff
@@ -2,6 +2,16 @@ defmodule Sentry.Config do
   @moduledoc false
 
   integrations_schema = [
+    ttl: [
```
whatyouhide, line 5: We cannot call this `:ttl` and put it at the top of the `:integrations` subschema: it's too generic. I would call this `:max_expected_check_in_time`.

Also, in the docs you need to explain the use case here: you need to explain that the SDK reports check-in starts and check-in ends, and with this key then you will stop reporting check-in ends if they come `:max_expected_check_in_time` *after* the check-in start.

Also, try to avoid leaking implementation details in the docs. Don't mention the ETS table or sweeping. You can say something like "if check-in ends come after this time then the SDK won't report them, and this is to avoid memory leaks" yada yada.

lib/sentry/config.ex
```diff
@@ -2,6 +2,16 @@ defmodule Sentry.Config do
...
+      type: :integer,
+      default: 600_000,
+      doc: """
+      The time in milliseconds that a check-in id will live after it has been created.
+
+      For long running cron jobs, set ttl (time to live) to longest running job time in milliseconds.
+      This avoids jobs being sweeped from table before they've been completed. Defaults to ten minutes.
```
whatyouhide, line 12:
```suggestion
      This avoids jobs being sweeped from table before they've been completed.
```

nimble_options already documents the default you used in the `:default` key (try it with `mix docs`).

lib/sentry/integrations/check_in_id_mappings.ex
```diff
@@ -5,30 +5,58 @@ defmodule Sentry.Integrations.CheckInIDMappings do
...
+    _ = :ets.select_delete(@table, match_spec)
+
+    schedule_sweep(state)
+    {:noreply, state}
+  end
+
+  ## Helpers
+
+  defp schedule_sweep(ttl_millisec) do
+    Process.send_after(self(), {:sweep, ttl_millisec}, @sweep_interval_millisec)
```
whatyouhide, line 55: Do you need to send over the `ttl_millisec` if it's the state of the GenServer anyway?

whatyouhide (commented):
Left a few preliminary comments, but it's a great start. 🙃

lib/sentry/integrations/check_in_id_mappings.ex
savhappy: Nope! Thx for pointing that out!

lib/sentry/config.ex
savhappy, line 1: Elaborate in here or somewhere else in docs on what check-in id is? @whatyouhide

lib/sentry/integrations/check_in_id_mappings.ex
savhappy: Okay by removing this meant I had to make this genserver configurable for testing since we're testing both a singleton and isolated process. So this caused some extra changes in the code

lib/sentry/application.ex
```diff
@@ -19,11 +19,14 @@ defmodule Sentry.Application do
...
 
+    integrations_config = Keyword.fetch!(config, :integrations)
+
     children =
       [
         {Registry, keys: :unique, name: Sentry.Transport.SenderRegistry},
         Sentry.Dedupe,
-        Sentry.Integrations.CheckInIDMappings
+        {Sentry.Integrations.CheckInIDMappings,
+         [max_expected_check_in_time: integrations_config[:max_expected_check_in_time]]}
```
whatyouhide, line 29: We always expect this:

```suggestion
         [max_expected_check_in_time: Keyword.fetch!(integrations_config, :max_expected_check_in_time]}
```

lib/sentry/config.ex
```diff
@@ -2,6 +2,18 @@ defmodule Sentry.Config do
   @moduledoc false
 
   integrations_schema = [
+    max_expected_check_in_time: [
+      type: :integer,
+      default: 600_000,
+      doc: """
+      The time in milliseconds that a check-in id will live after it has been created.
```
whatyouhide, line 9:
```suggestion
      The time in milliseconds that a check-in ID will live after it has been created.
```

lib/sentry/config.ex
```diff
@@ -2,6 +2,18 @@ defmodule Sentry.Config do
...
+
+      The SDK reports the start and end of each check-in. However, to optimize performance and prevent
+      potential memory issues, if a check-in end event is reported after the specified `max_expected_check_in_time`,
+      the SDK will not report it. This behavior helps manage resource usage effectively while still providing necessary
+      tracking for your jobs.
```
whatyouhide, line 14: @savhappy elaborate here what a check-in is and why this option is necessary

test/sentry/check_in_id_mappings_test.exs
```diff
@@ -0,0 +1,30 @@
...
+        start:
+          {CheckInIDMappings, :start_link, [[max_expected_check_in_time: 0, name: TestMappings]]}
+      }
+
+      pid = start_supervised!(child_spec)
+
+      CheckInIDMappings.lookup_or_insert_new(cron_key)
+      assert :ets.lookup(@table, cron_key) != []
+
+      Process.sleep(5)
```
whatyouhide, line 23: Why sleep here?

savhappy: We need to sleep to allow the `:sweep` message to be processed by the target process so that the final assertion accurately reflects the updated state of the ETS table. Otherwise, it returns the key/value pair before being deleted.

lib/sentry/config.ex
```diff
@@ -2,6 +2,20 @@ defmodule Sentry.Config do
...
+      default: 600_000,
+      doc: """
+      The time in milliseconds that a check-in ID will live after it has been created.
+
+      The SDK reports the start and end of each check-in. A check-in is used to track the
+      progress of a specific check-in event associated with cron job telemetry events that are a part
+      of the same job. However, to optimize performance and prevent potential memory issues,
+      if a check-in end event is reported after the specified `max_expected_check_in_time`,
+      the SDK will not report it. This behavior helps manage resource usage effectively while still
+      providing necessary tracking for your jobs.
```
whatyouhide, line 16:
```suggestion
      providing necessary tracking for your jobs.
      
      *Available since 10.6.3*.
```

test/sentry/check_in_id_mappings_test.exs
whatyouhide: If you call `:sys.get_state(pid)`, you are making a sync call to the GenServer. Since messages are ordered, you are guaranteed that by the time the GenServer replies to the `:sys.get_state/1` call then the `:sweep` message will have been processed. So, you can remove this 🙃

whatyouhide (commented):
Last two comments and we're good to go!

test/sentry/check_in_id_mappings_test.exs
savhappy: Maybe I'm missing something? When I remove the sleep it returns the pair and the final assert fails

```   test/sentry/check_in_id_mappings_test.exs:9
     Assertion with == failed
     code:  assert :ets.lookup(@table, cron_key) == []
     left:  [{"quantum_123", "ec2ded4d0c1a43b580d2e567396e3bee", 1723541054291}]
     right: []
     stacktrace:
       test/sentry/check_in_id_mappings_test.exs:27: (test)
```

whatyouhide: Ok no rush I can debug this locally.

whatyouhide (approved):

lib/sentry/config.ex
```diff
@@ -2,6 +2,22 @@ defmodule Sentry.Config do
...
+
```
whatyouhide, line 17:
```suggestion

```

PR #765, opened 2024-08-06
Fix logger handler support for structured log messages

martosaur:
STR:
1. In a configured project, issue a log with structured message: `Logger.error(foo: "bar")`
2. Check message didn't get to Sentry

Here's my attempt on fixing this. Looks like the case statement is trying to work on both `report` and `msg`. Also added a basic test as I don't believe there are any for `{:report, report}` clause.

### DIFF
--- a/lib/sentry/logger_handler.ex
+++ b/lib/sentry/logger_handler.ex
@@ -341,17 +341,10 @@ defmodule Sentry.LoggerHandler do
   # "report" here is of type logger:report/0, which is a map or keyword list.
   defp log_unfiltered(%{msg: {:report, report}}, sentry_opts, %__MODULE__{} = config) do
     case Map.new(report) do
-      %{report: %{reason: {exception, stacktrace}}}
+      %{reason: {exception, stacktrace}}
       when is_exception(exception) and is_list(stacktrace) ->
         sentry_opts = Keyword.merge(sentry_opts, stacktrace: stacktrace, handled: false)
         capture(:exception, exception, sentry_opts, config)
-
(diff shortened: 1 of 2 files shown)


lib/sentry/logger_handler.ex
```diff
@@ -341,18 +341,11 @@ defmodule Sentry.LoggerHandler do
...
       when is_exception(exception) and is_list(stacktrace) ->
         sentry_opts = Keyword.merge(sentry_opts, stacktrace: stacktrace, handled: false)
         capture(:exception, exception, sentry_opts, config)
 
-      %{report: %{reason: {reason, stacktrace}}} when is_list(stacktrace) ->
-        sentry_opts = Keyword.put(sentry_opts, :stacktrace, stacktrace)
-        capture(:message, "** (stop) " <> Exception.format_exit(reason), sentry_opts, config)
-
-      %{report: report_info} ->
-        capture(:message, inspect(report_info), sentry_opts, config)
```
whatyouhide, line 354: We were never hitting these clauses? Did `mix coveralls.html` confirm that? 😮

martosaur: I think so? I mean, the whole function head can be deleted and tests still pass...
CleanShot 2024-08-07 at 08 57 11

whatyouhide: Uh, interesting. Well, not good 😄 I suspect that maybe some of those were for older Erlang versions but your change passes on all supported Erlang/Elixir versions, so we're good to go!

whatyouhide (approved):

PR #771, opened 2024-08-17
Add integration_meta key to event

savhappy:
-Added an integration_meta key to store the metadata of a cron job, allowing the user to filter what events are reported based on the values. 
-Alternate solution to PR #759 
-Related to #757

### DIFF
--- a/lib/sentry/event.ex
+++ b/lib/sentry/event.ex
@@ -59,7 +59,7 @@ defmodule Sentry.Event do
           modules: %{optional(String.t()) => String.t()},
           extra: map(),
           fingerprint: [String.t()],
-          integration_meta: map() | nil,
+          integration_meta: map(),
 
           # Interfaces.
           breadcrumbs: [Interfaces.Breadcrumb.t()],
--- a/lib/sentry/event.ex
+++ b/lib/sentry/event.ex
@@ -349,31 +349,30 @@ defmodule Sentry.Event do
     source = Keyword.get(opts, :event_source)
     handled? = Keyword.fetch!(opts, :handled)
 
-    event =
-      %__MODULE__{
-        attachments: attachments_context,
-        breadcrumbs: breadcrumbs,
-        contexts: generate_contexts(),
-        culprit: culprit_from_stacktrace(Keyword.get(opts, :stacktrace, [])),
-        environment: Config.environment_name(),
-        event_id: UUID.uuid4_hex(),
-        exception: List.wrap(coerce_exception(exception, stacktrace, message, handled?)),
-        extra: extra,
-        fingerprint: Keyword.fetch!(opts, :fingerprint),
-        level: Keyword.fetch!(opts, :level),
-        message: message && build_message_interface(message, opts),
-        modules: :persistent_term.get({:sentry, :loaded_applications}),
-        original_exception: exception,
-        release: Config.release(),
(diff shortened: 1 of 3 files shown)


lib/sentry/event.ex
```diff
@@ -59,6 +59,7 @@ defmodule Sentry.Event do
           modules: %{optional(String.t()) => String.t()},
           extra: map(),
           fingerprint: [String.t()],
+          integration_meta: map() | nil,
```
whatyouhide, line 62:
```suggestion
          integration_meta: map(),
```

lib/sentry/event.ex
```diff
@@ -117,6 +118,7 @@ defmodule Sentry.Event do
     transaction: nil,
     threads: nil,
     user: %{},
+    integration_meta: nil,
```
whatyouhide, line 121:
```suggestion
    integration_meta: %{},
```

lib/sentry/event.ex
```diff
@@ -255,6 +257,14 @@ defmodule Sentry.Event do
       `Sentry.capture_message/2`. *Available since v10.1.0*.
       """
     ],
+    integration_meta: [
+      type: :map,
+      doc: """
+      A map to store metadata specific to the integration.
+      It allows cron integrations, such as the Oban integration,
+      to store job-related metadata. *Available since v10.1.0*.
+      """
```
whatyouhide, line 266: For now this can be `doc: false` instead since integrations are all internal to the Sentry SDK.

lib/sentry/event.ex
```diff
@@ -339,29 +349,31 @@ defmodule Sentry.Event do
...
-      request: struct(%Interfaces.Request{}, request),
-      sdk: @sdk,
-      server_name: Config.server_name() || to_string(:net_adm.localhost()),
-      source: source,
-      tags: tags,
-      timestamp: timestamp,
-      user: user
-    }
+    event =
+      %__MODULE__{
```
whatyouhide, line 353: Can you not change the formatting of this? Makes it harder to review 🙃

lib/sentry/integrations/oban/error_reporter.ex
```diff
@@ -41,7 +41,9 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
           inspect(job.worker),
           Exception.message(exception)
         ],
-        extra: Map.take(job, [:args, :attempt, :id, :max_attempts, :meta, :queue, :tags, :worker])
+        extra:
+          Map.take(job, [:args, :attempt, :id, :max_attempts, :meta, :queue, :tags, :worker]),
```
whatyouhide, line 45: We need tests for this 🧪

whatyouhide (changes requested):

lib/sentry/event.ex
savhappy: Ah yes! My linter did this..

lib/sentry/integrations/oban/error_reporter.ex
savhappy: on it!

lib/sentry/event.ex
whatyouhide: Silly goose linter

whatyouhide (approved):

PR #772, opened 2024-08-19
Validate options in dev mode

### LINKED ISSUE #766: Validate all params passed to sentry even in dev/test
👋 
Me again, and thanks again!

It'd be great if even in dev/test mode Sentry validated all parameters/options passed. It does validate some but not all:

```
tobi@qiqi:~$ iex -S mix
iex(1)> Sentry.capture_message("hu", client: [oha: :nonsense])
:ignored
tobi@qiqi:~$ MIX_ENV=prod iex -S mix
iex(1)> Sentry.capture_message("hu", client: [oha: :nonsense])
** (NimbleOptions.ValidationError) invalid value for :client option: expected atom, got: [oha: :nonsense]
    (nimble_options 1.1.1) lib/nimble_options.ex:359: NimbleOptions.validate!/2
    (sentry 10.6.2) lib/sentry/client.ex:106: Sentry.Client.send_event/2
    iex:1: (file)
iex(1)> 

```

I suspect this behavior is because Sentry is deactivated, but it allows some bad options to slip through dev and test to production. I.e. I had somehow (don't ask me how) falsely assumed that I could pass an environment name to the client in `capture_message`'s optins: `client: [environment_name: System.get_env("ENV)]` - which made it all the way to production 😅 

So still checking that even if dsn isn't configured would be nice I think.

savhappy:
-validate all options passed to send_event even when in dev mode
-closes #766

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -335,7 +335,8 @@ defmodule Sentry do
         Client.send_event(event, opts)
 
       !Config.dsn() ->
-        Client.validate_and_ignore(opts)
+        _opts = Client.validate_options!(opts)
+        :ignored
 
       included_envs == :all or to_string(Config.environment_name()) in included_envs ->
         Client.send_event(event, opts)
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -335,7 +335,7 @@ defmodule Sentry do
         Client.send_event(event, opts)
 
       !Config.dsn() ->
-        :ignored
+        Client.validate_and_ignore(opts)
 
       included_envs == :all or to_string(Config.environment_name()) in included_envs ->
         Client.send_event(event, opts)
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -160,6 +160,12 @@ defmodule Sentry.Client do
     end
   end
 
+  @spec validate_and_ignore(keyword()) :: :ignored
+  def validate_and_ignore(opts) when is_list(opts) do
+    _opts = NimbleOptions.validate!(opts, @send_event_opts_schema)
+    :ignored
+  end
+
   defp sample_event(sample_rate) do
     cond do
       sample_rate == 1 -> :ok
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -116,7 +116,7 @@ defmodule Sentry.Client do
           | :unsampled
           | :excluded
   def send_event(%Event{} = event, opts) when is_list(opts) do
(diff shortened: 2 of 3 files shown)


lib/sentry.ex
```diff
@@ -335,7 +335,7 @@ defmodule Sentry do
         Client.send_event(event, opts)
 
       !Config.dsn() ->
-        :ignored
+        Client.validate_and_ignore(opts)
```
whatyouhide, line 338: This is gonna make it harder to track the return value–I have to jump to another module. Maybe we can do

```elixir
Client.validate_options!(opts)
:ignored
```

instead?

lib/sentry/client.ex
```diff
@@ -160,6 +160,12 @@ defmodule Sentry.Client do
     end
   end
 
+  @spec validate_and_ignore(keyword()) :: :ignored
+  def validate_and_ignore(opts) when is_list(opts) do
```
whatyouhide, line 164: We can also re-use this function above where we call `NimbleOptions.validate!` if we change it to return the options, like suggested in the other comment.

test/sentry_test.exs
```diff
@@ -119,6 +119,16 @@ defmodule SentryTest do
     assert logged_count == 2
   end
 
+  test "raises error with validate_and_ignore/1 in dev mode if opts passed are invalid " do
+    put_test_config(dsn: nil, test_mode: false)
+
+    assert_raise(NimbleOptions.ValidationError, fn ->
+      Sentry.Client.validate_and_ignore(client: [bad_key: :nada])
+    end)
```
whatyouhide, line 127: Nit:

```suggestion
    assert_raise NimbleOptions.ValidationError, fn ->
      Sentry.Client.validate_and_ignore(client: [bad_key: :nada])
    end
```

test/sentry_test.exs
```diff
@@ -119,6 +119,16 @@ defmodule SentryTest do
...
+
+    assert :ignored == Sentry.Client.validate_and_ignore(client: :hackney)
```
whatyouhide, line 129: Nit, but matches other tests:

```suggestion
    assert :ignored = Sentry.Client.validate_and_ignore(client: :hackney)
```

whatyouhide (commented):
Left just a couple of comments but this is great.

lib/sentry.ex
savhappy: Yes, good idea!

whatyouhide (approved):

PR #776, opened 2024-08-26
fix(tasks): allow passing `--no-compile` and `--no-check-deps` to `package_source_code` task

razielgn:
10.7.0 broke invoking `mix sentry.package_source_code` during a build using nix (xkcd 1172 😂 )

That's because inside the nix build env, running `loadpaths` and `compile` must be avoided because all deps are already compiled separately and loaded read-only via env vars.

These flags are somewhat "standard" and used in various phoenix/ecto tasks:
- https://github.com/phoenixframework/phoenix/blob/e99f657f1cc9062fca0f2b8b79bc90659d8bd514/lib/mix/tasks/phx.routes.ex#L66-L68
- https://github.com/elixir-ecto/ecto_sql/blob/master/lib/mix/tasks/ecto.load.ex#L54-L55


whatyouhide (approved):

razielgn:
Thanks!

PR #778, opened 2024-08-28
Slightly re-haul how we document options

### LINKED ISSUE #763: Document other options for capture_message/capture_exception
👋 Thanks for sentry,

right now both `capure_message/2` and `capture_exception/2` claim the following:

> The opts argument is passed as the second argument to [send_event/2](https://hexdocs.pm/sentry/Sentry.html#send_event/2).

This leaves you relatively confused as [`send_event/2`](https://hexdocs.pm/sentry/Sentry.html#send_event/2) supports relatively few options.

For instance the README itself uses `Sentry.capture_exception(my_exception, stacktrace: __STACKTRACE__)` which isn't documented in there. 

Some digging in the code reveals that the opts are split between client and event opts, the latter of which end up at some form of [`Event.create_event/1`](https://hexdocs.pm/sentry/Sentry.Event.html#create_event/1).

This includes some params I'd consider essential:

* stacktrace
* level
* extra

These are hard to find now from the README & docs (at least I have completely missed them without digging into the code).

I think it'd be nice to either point people to the doc for the other function, or since it seems rather internal, document these options in another public facing common place so they're easier to find and adjust :)

I'm not PRing it, as I'm not sure where this should best live :)

savhappy:
-add details pointing users toward valid accepted options 
-closes #763

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -314,6 +314,9 @@ defmodule Sentry do
   information about an exception, a message, or any other event that you want to
   report. To manually build events, see the functions in `Sentry.Event`.
 
+    ## Options
+    #{NimbleOptions.docs(Options.get_client_options())}
+
   > #### Async Send {: .error}
   >
   > Before v9.0.0 of this library, the `:result` option also supported the `:async` value.
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -16,6 +16,9 @@ defmodule Sentry do
       [*Setup with Plug and Phoenix* guide](setup-with-plug-and-phoenix.html), and the
       `Sentry.PlugCapture` and `Sentry.PlugContext` modules.
 
+    * Through integrations for various ecosystem tools, like [Oban](oban-integration.html)
+      or [Quantum](quantum-integration.html).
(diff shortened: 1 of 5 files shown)


whatyouhide:
@savhappy I don't think we should refer to how we split options internally (users don't really have to know that we are using nimble_options and stuff). I think what we had in mind here is that we want to *list* all the options in one place and link to that place as much as possible.

savhappy:
Ah very clear! I'll take that approach. Thanks 👌

savhappy:
@whatyouhide this what you had in mind?

whatyouhide:
@savhappy I think `Sentry.Options` is a great idea to store all the options, validation, and so on. However, if we make it public then we are committing to supporting the `Sentry.Options` module being available. What if, instead, we do:

```elixir
defmodule Sentry.Options do
  @moduledoc false

  ...
end
```

and then document all options in a `## Options` section in `Sentry`? You can always say "see the *Options* section in the `Sentry` documentation" if you want to link people to options. Thoughts?

whatyouhide (approved):

PR #780, opened 2024-08-29
Accept structs as structured logs in Sentry.LoggerHandler

mrdotb:
👋 

While testing my custom logger handler, which is based on the Sentry Elixir handler, I encountered an issue where passing a struct to `Logger.error` causes the handler to detach. The error handler only supports keywords and maps.

The problem arose when using the Tesla library, which passes a `Tesla.Env` struct to `Logger.error`. Interestingly, `Logger.info` does support passing structs, and the struct is displayed as a keyword list  or map with the `__struct__ `field

```elixir
iex(3)> Logger.error(%Foo{})
:ok
23:36:43.117 [error] [__struct__: Foo, bar: nil]
```

I added a test that reproduce the issue and a fix proposition.

### DIFF
--- a/lib/sentry/logger_handler.ex
+++ b/lib/sentry/logger_handler.ex
@@ -338,7 +338,12 @@ defmodule Sentry.LoggerHandler do
     log_from_crash_reason(log_event.meta[:crash_reason], unicode_chardata, sentry_opts, config)
   end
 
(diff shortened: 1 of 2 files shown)


whatyouhide (approved):
Great catch, thanks @mrdotb 💟

PR #784, opened 2024-09-04
[WIP] Support for Transactions/Spans via Opentelemetry Span Processor

solnic:
## 🚧 WIP

TBD


test/sentry/telemetry/span_processor_test.exs
```diff
@@ -0,0 +1,30 @@
...
+    TestEndpoint.instrumented_function()
+    assert [%Sentry.Transaction{} = transaction] = Sentry.Test.pop_sentry_transactions()
+
+    assert length(transaction.spans) == 1
+    assert [span] = transaction.spans
+    assert span.op == "instrumented_function"
+    assert span.timestamp > span.start_timestamp
+    assert transaction.timestamp > transaction.start_timestamp
+  end
+end
```
solnic, line 30: Screenshot 2024-09-04 at 19 22 51

@whatyouhide do you happen to know why `start_timestamp` and `timestamp` values are negative?

whatyouhide: Depends on your system I think but it could be that these are in the `:native` time unit. How are you creating them? You getting them from OTel?

For example, see https://hexdocs.pm/opentelemetry_api/1.3.1/opentelemetry.html#timestamp_to_nano/1 or https://hexdocs.pm/elixir/System.html#convert_time_unit/3

solnic: > Depends on your system I think but it could be that these are in the :native time unit. How are you creating them? You getting them from OTel?

Just via Tracer:

```elixir
  def mount(_params, _session, socket) do
    Tracer.with_span "list_posts" do
      posts = Publishing.list_posts()

      Tracer.set_attribute("posts.count", length(posts))

      {:ok, stream(socket, :posts, posts)}
    end
  end
```

whatyouhide: Yeah getting them from OTel then. I'd try `:opentelemetry.timestamp_to_nano` on those values and see if it gives you a reasonable nanosecond Unix timestamp 😄

whatyouhide:
@solnic can you rebase this one now that #797 has been merged? Gonna make it easier to review I think 😄

solnic:
> @solnic can you rebase this one now that #797 has been merged? Gonna make it easier to review I think 😄

@whatyouhide done!

.github/workflows/main.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       run: |
         mix deps.get --check-locked
         cd test_integrations/phoenix_app
-        mix deps.get --check-locked
+        mix deps.get
```
solnic, line 63: FYI this would not pass under 1.16 for some reason. Lock is different under this version of Elixir :/

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,88 @@
+defmodule Sentry.Opentelemetry.SpanStorage do
```
whatyouhide, line 1: This design won't work for scaling out. You will have a *single* GenServer serializing access through the whole application. You could have thousands of processes calling this at the same time, and all those requests would have to go through this process and wait for their turn. We will have to rework this pretty substantially.

For something like this, a good tool to reach for is ETS. Erlang's official OTel libraries use that as well for storing spans:

https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry/src/otel_span_ets.erl

We _might_ be able to reuse Erlang's implementation itself to store and then retrieve the spans. I'd suggest to ask around (https://opentelemetry.io/community/) or let me know if you want me to put you in touch with one of the devs of the OTel/Erlang integration—I would help but I don't have the bandwidth just this week.

mix.exs
```diff
@@ -111,7 +111,9 @@ defmodule Sentry.Mixfile do
       # Required by Phoenix.LiveView's testing
       {:floki, ">= 0.30.0", only: :test},
       {:oban, "~> 2.17 and >= 2.17.6", only: [:test]},
-      {:quantum, "~> 3.0", only: [:test]}
+      {:quantum, "~> 3.0", only: [:test]},
+      {:opentelemetry, "~> 1.4"},
+      {:opentelemetry_api, "~> 1.3"}
```
whatyouhide, line 116: Are we ok with Sentry *always* depending on OTel libraries, even when users won't care about transactions?

whatyouhide (changes requested):

lib/sentry/opentelemetry/span_storage.ex
savhappy: @solnic I can help/look into this today and over the weekend!

solnic: @savhappy sounds good, thanks :)

mix.exs
solnic: @whatyouhide right, this could become optional

lib/sentry/opentelemetry/span_storage.ex
savhappy: @whatyouhide would it be better to have two tables (one for root spans and one for children) or to store the key/value pairs accordingly to differentiate like this -> 

key for root {:root_span, span_id} 
key for children {parent_span_id, span_id}

yordis:
@whatyouhide are you adding support for the otel propagator map from #820 I am failing to see where it is being done

danschultzer:
What's the status of this PR? We would love to be start using otel with Sentry.

solnic:
@danschultzer we need to finish a little rework in #817 - after it's merged into this one, we need to go through another round of reviews, assuming we're happy with the implementation and testing goes well, we should be able to wrap it up and release later this month.

solnic:
@whatyouhide seems like now span_processor_test is failing on 1.13 - I'm looking into this!

danschultzer:
I've tested this in staging enviroment and seen two issues:

1. Oban is not handled in `build_transaction`:

```elixir
  Handler "Elixir.OpentelemetryOban.PluginHandler.plugin_stop" has failed and has been detached. Class=:error
  Reason=:function_clause
  Stacktrace=[
     {Sentry.OpenTelemetry.SpanProcessor, :build_transaction,
      [
         %Sentry.OpenTelemetry.SpanRecord{
           trace_id: "450fcd62516cbff48d61066e2db6af37",
            span_id: "996b1cea0c7231f4",
             tracestate: {:tracestate, []},
              parent_span_id: nil,
               name: "Elixir.Oban.Stager process",
                kind: :internal,
                 start_time: "2024-12-10T05:03:40.820Z",
                  end_time: "2024-12-10T05:03:40.820Z",
                   attributes: %{"oban.plugin" => Oban.Stager},
                    events: {:events, 128, 128, :infinity, 0, []},
                     links: {:links, 128, 128, :infinity, 0, []},
                      status: :undefined,
                       trace_flags: 1,
                        is_recording: false,
                         instrumentation_scope: {:instrumentation_scope, "opentelemetry_oban",
                          "1.1.1", :undefined},
                           origin: "opentelemetry_oban"
                            },
                             []
                              ], [file: ~c"lib/sentry/opentelemetry/span_processor.ex", line: 57]},
                               {Sentry.OpenTelemetry.SpanProcessor, :on_end, 2,
                                [file: ~c"lib/sentry/opentelemetry/span_processor.ex", line: 29]},
                                 {:otel_tracer_server, :"-on_end/1-fun-0-", 3,
                                  [file: ~c"/app/deps/opentelemetry/src/otel_tracer_server.erl", line: 195]},
                                   {:lists, :foldl, 3, [file: ~c"lists.erl", line: 2146]},
                                    {:otel_span, :end_span, 1, [file: ~c"src/otel_span.erl", line: 294]},
                                     {:otel_telemetry, :end_telemetry_span, 2,
                                      [file: ~c"src/otel_telemetry.erl", line: 52]},
                                       {:telemetry, :"-execute/3-fun-0-", 4,
                                        [file: ~c"/app/deps/telemetry/src/telemetry.erl", line: 167]},
                                         {:lists, :foreach_1, 2, [file: ~c"lists.erl", line: 2310]}
                                         ]
```

2. `get_data_category` fails as `Sentry.Transaction` is not handled:

```elixir
GenServer {Sentry.Transport.SenderRegistry, 7} terminating
** (FunctionClauseError) no function clause matching in Sentry.Envelope.get_data_category/1
 (sentry 10.8.0) lib/sentry/envelope.ex:68: Sentry.Envelope.get_data_category(%Sentry.Transaction{event_id: "c17ec8f1f9094d58b52e28332ba9c68a", start_timestamp: "2024-12-10T05:16:11.079Z", timestamp: "2024-12-10T05:16:11.082Z", transaction: "app.repo.query", transaction_info: %{source: "component"}, status: nil, contexts: %{trace: %{data: %{"db.instance" => nil, "db.name" => nil, "db.system" => :postgresql, "db.type" => :sql, "db.url" => "ecto:", "decode_time_microseconds" => 3, "idle_time_microseconds" => 1250815, "query_time_microseconds" => 2032, "queue_time_microseconds" => 1014, "source" => nil, "total_time_microseconds" => 3051}, origin: "opentelemetry_ecto", description: "app.repo.query", op: "db.sql.ecto", span_id: "87d6d9b1c61ced34", trace_id: "119004179fa20ee3af70641c1e331508", parent_span_id: nil}}, request: nil, measurements: %{}, spans: [], type: "transaction"})
  (sentry 10.8.0) lib/sentry/client_report/sender.ex:36: anonymous fn/3 in Sentry.ClientReport.Sender.record_discarded_events/3
   (elixir 1.17.3) lib/enum.ex:987: Enum."-each/2-lists^foreach/1-0-"/2
    (sentry 10.8.0) lib/sentry/transport.ex:69: Sentry.Transport.post_envelope_with_retries/6
     (sentry 10.8.0) lib/sentry/transport.ex:32: Sentry.Transport.encode_and_post_envelope/3
      (sentry 10.8.0) lib/sentry/transport/sender.ex:67: Sentry.Transport.Sender.handle_cast/2
       (stdlib 6.1) gen_server.erl:2371: :gen_server.try_handle_cast/3
        (stdlib 6.1) gen_server.erl:2433: :gen_server.handle_msg/6
        Process Label: Sentry.Transport.Sender
        Last message: {:"$gen_cast", {:send, Sentry.HTTPClient.app, %Sentry.Transaction{event_id: "c17ec8f1f9094d58b52e28332ba9c68a", start_timestamp: "2024-12-10T05:16:11.079Z", timestamp: "2024-12-10T05:16:11.082Z", transaction: "app.repo.query", transaction_info: %{source: "component"}, status: nil, contexts: %{trace: %{data: %{"db.instance" => nil, "db.name" => nil, "db.system" => :postgresql, "db.type" => :sql, "db.url" => "ecto:", "decode_time_microseconds" => 3, "idle_time_microseconds" => 1250815, "query_time_microseconds" => 2032, "queue_time_microseconds" => 1014, "source" => nil, "total_time_microseconds" => 3051}, origin: "opentelemetry_ecto", description: "app.repo.query", op: "db.sql.ecto", span_id: "87d6d9b1c61ced34", trace_id: "119004179fa20ee3af70641c1e331508", parent_span_id: nil}}, request: nil, measurements: %{}, spans: [], type: "transaction"}}}
```

config/config.exs
```diff
@@ -15,4 +15,6 @@ if config_env() == :test do
   config :logger, backends: []
 end
 
+config :opentelemetry, span_processor: {Sentry.OpenTelemetry.SpanProcessor, []}
```
solnic, line 18: @whatyouhide seems like this just doesn't do anything under Elixir 1.13 - do you maybe have any insights here? What would be an equivalent of this for older Elixir?

danschultzer:
Testing this, I see that the sentry environment is not set for the traces.

Also it looks like [all](https://github.com/getsentry/sentry-ruby/tree/master/sentry-opentelemetry) [other](https://github.com/getsentry/sentry-javascript/tree/develop/packages/opentelemetry) [SDK's](https://github.com/getsentry/sentry-go/tree/master/otel) makes the OTel integration be a separate package, should it be the same for the elixir sdk? At least all the otel modules needs to be wrapped in a ensure_loaded conditional as the otel deps are optional, but I think it does make sense that the sentry otel bridge its own dep?

solnic:
> I've tested this in a production enviroment and seen two issues:

@danschultzer thanks for testing it out so early, this is very helpful! I fixed both issues, but I consider Oban support still a WIP.

solnic:
> Also it looks like [all](https://github.com/getsentry/sentry-ruby/tree/master/sentry-opentelemetry) [other](https://github.com/getsentry/sentry-javascript/tree/develop/packages/opentelemetry) [SDK's](https://github.com/getsentry/sentry-go/tree/master/otel) makes the OTel integration be a separate package, should it be the same for the elixir sdk? At least all the otel modules needs to be wrapped in a ensure_loaded conditional as the otel deps are optional, but I think it does make sense that the sentry otel bridge is a separate package?

Yes I think that would make sense. Thoughts @whatyouhide?

danschultzer:
Sure, I've been testing this a lot more and I got several suggestions to it. I've forked your work @solnic and fixed some of the issues I had there along with things I think could be refactored. Let me push a draft PR so you can check it out, just need to clean it up a bit. Among other things, I think there's some considerations to be made on how much we want to manually normalize the attributes vs just deferring to otel.

solnic:
@danschultzer I made it pull opentelemetry_oban from your fork for the time being, and made some adjustments. See recent commits:

- 844313d04098e7dd1186b7226243c4e908e01295 - this makes Oban processing traces/spans follow Sentry's/OTel conventions more closely
- 65f6b116d340d7c368285dd8b0f34eb67801a8f2 - this makes it skip Oban Stager traces as they produce a lot of noise. We may want to make that configurable though

@whatyouhide could you take another look over here?

lib/sentry/client.ex
```diff
@@ -225,6 +282,19 @@ defmodule Sentry.Client do
...
 
+  @spec render_transaction(%Transaction{}) :: map()
+  def render_transaction(%Transaction{} = transaction) do
+    transaction
+    |> Transaction.to_map()
+    |> Map.merge(%{
+      platform: "elixir",
+      sdk: %{
+        name: "sentry.elixir",
+        version: "10.7.1"
```
danschultzer, line 293: We want this to be pulled dynamically right?

```suggestion
        version: Application.spec(:sentry, :vsn)
```

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,199 @@
...
+  end
+
+  @impl true
+  def force_flush(_config) do
+    :ok
+  end
+
+  # TODO: "Elixir.Oban.Stager process" is an internal span and it produces a lot of noise
+  #       so it makes sense to ignore it - should this be configurable?
+  defp build_transaction(%{name: "Elixir.Oban.Stager process"}, _), do: nil
```
danschultzer, line 67: I think ideally we want the end-user use an otel sampler for this? Other sentry SDK's don't have any notion of filtering otel spans IIRC.

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,199 @@
...
+      spans: [root_span | child_spans]
+    })
+  end
+
+  defp transaction_name(
+         %{attributes: %{unquote(to_string(MessagingAttributes.messaging_system())) => :oban}} =
+           span_record
+       ) do
+    span_record.attributes["oban.job.worker"]
+  end
```
danschultzer, line 90: FWIW this breaks the otel semantic conventions for messaging systems. Though the oban otel instrumentation also doesn't follow the guidelines either. This PR resolves it: https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/436

Instead of `{job} proces` it should be `process {job}` or `process {queue}`.

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,199 @@
...
+
+    {"db", db_query_text}
+  end
+
+  defp get_op_description(%{
+         attributes:
+           %{unquote(to_string(MessagingAttributes.messaging_system())) => :oban} = attributes
+       }) do
+    {"queue.process", attributes["oban.job.worker"]}
+  end
```
danschultzer, line 143: I think we should aim to make this agnostic instead of oban specific, but does require upstream to be fixed with https://github.com/open-telemetry/opentelemetry-erlang-contrib/pull/436 so we can `messaging.operation.type` and `messaging.destination.name` for this.

lib/sentry/envelope.ex
```diff
@@ -46,16 +48,35 @@ defmodule Sentry.Envelope do
...
+          | Event.t()
+          | Transaction.t()
+        ) ::
           String.t()
   def get_data_category(%Attachment{}), do: "attachment"
   def get_data_category(%CheckIn{}), do: "monitor"
   def get_data_category(%ClientReport{}), do: "internal"
   def get_data_category(%Event{}), do: "error"
+  # TODO: is this correct?
+  def get_data_category(%Transaction{}), do: "error"
```
danschultzer, line 79: Lookd like it should be transaction? https://github.com/getsentry/sentry-ruby/blob/ad535cc67b6084329cbf5992c3a184a4b1f80a6b/sentry-ruby/lib/sentry/envelope/item.rb#L16-L26

lib/sentry/opentelemetry/span_processor.ex
solnic: @danschultzer added a sampler via 267b82233a4b154dbd4b6a406f46fc123bbe1d89

lib/sentry/opentelemetry/span_processor.ex
solnic: @danschultzer I'm gonna hold off with Oban support until your PRs are merged, otherwise it's a moving target.

solnic:
> @solnic @danschultzer can you folks let me know when this will be Ready to Review™? I havent been paying attention cause the rate of changes is too high but I think I would take a look before we take action on merging this.

@whatyouhide it's in a flux still as it's waiting for some upstream improvements in opentelemetry packages but I would really appreciate if you could review our updated approach to span storage. We still use genserver but now it uses ets, and there's a sweeping mechanism in place. This part could really use a review and I could address your feedback while we're waiting for other things.

lib/sentry/opentelemetry/span_record.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  @fields Record.extract(:span, from_lib: "opentelemetry/include/otel_span.hrl")
+  Record.defrecordp(:span, @fields)
+
+  defstruct @fields ++ [:origin]
+
+  def new(otel_span) do
+    otel_attrs = span(otel_span)
+
+    {:attributes, _, _, _, attributes} = otel_attrs[:attributes]
```
whatyouhide, line 13: You can extract this record too

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,137 @@
...
+    schedule_cleanup(cleanup_interval)
+
+    {:ok, %{cleanup_interval: cleanup_interval}}
+  end
+
+  def store_span(span_data) when span_data.parent_span_id == nil do
+    stored_at = System.system_time(:second)
+
+    case :ets.lookup(@table, {:root_span, span_data.span_id}) do
+      [] -> :ets.insert(@table, {{:root_span, span_data.span_id}, {span_data, stored_at}})
```
whatyouhide, line 34: This is not race-condition-safe. You can use `ets:insert_new/2` here I think?

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,137 @@
...
+    cleanup_stale_spans()
+    schedule_cleanup(state.cleanup_interval)
+    {:noreply, state}
+  end
+
+  defp schedule_cleanup(interval) do
+    Process.send_after(self(), :cleanup_stale_spans, interval)
+  end
+
+  defp cleanup_stale_spans do
```
whatyouhide, line 109: You can use `match_delete` here I believe instead of `match_object` + deleting single spans

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,137 @@
+defmodule Sentry.OpenTelemetry.SpanStorage do
+  @moduledoc false
+  use GenServer
+
+  require Logger
+
+  @table :span_storage
+  @cleanup_interval :timer.minutes(5)
```
whatyouhide, line 8: This seems quite long, I think we can shorten the span TTL *and* this? What's the reasoning behind these two numbers?

whatyouhide (commented):
I think a better approach here would be to split this PR into the *data layer* and then the OTel implementation layer. The data layer should be easy and "non controversial": `Span` struct, `send_transaction`, and whatnot. Thoughts?

solnic:
> I think a better approach here would be to split this PR into the _data layer_ and then the OTel implementation layer. The data layer should be easy and "non controversial": `Span` struct, `send_transaction`, and whatnot. Thoughts?

@whatyouhide yes I agree, I'll split it up.

There's also a suggestion (somewhere in the comments here 😅) to actually separate otel stuff in a sentry-opentelemetry package. Would that make sense from the maintenance pov? We could keep it here but I'm not sure how that would affect our release pipeline (I suspect it should not be hard to configure though given that other SDKs use monorepo for multiple packages too).

whatyouhide:
I’m not sure we need `sentry-opentelemetry` if we make the `opentelemetry_*` deps optional. If we did, I'd rather put it in a separate repo.

danschultzer:
> I’m not sure we need `sentry-opentelemetry` if we make the `opentelemetry_*` deps optional. If we did, I'd rather put it in a separate repo.

FWIW all other Sentry SDK's are monorepos with individual libraries for instrumentation:
https://github.com/getsentry/sentry-ruby
https://github.com/getsentry/sentry-rust
https://github.com/getsentry/sentry-go
https://github.com/getsentry/sentry-javascript/tree/develop/packages

solnic:
I've split this into #842 and #843 /cc @whatyouhide

grantwest:
> I've split this into #842 and #843 /cc @whatyouhide

Does that mean this PR will be closed without merging?

jwaldrip:
Exciting to see progress. Been waiting a long time and can't wait to see the fruits of your labor!

hugobarauna:
@solnic any updates here? This would help us in Livebook Teams. :)

PR #787, opened 2024-09-06
Add missing cond for frames

### LINKED ISSUE #786: GenServer {Sentry.Transport.SenderRegistry, 6} terminating
### Environment

What version are you running?
10.7.1

### Steps to Reproduce

1. Integrate sentry and run the server
2. I purposely didn't turn on the DB to see the errors in the dashboard to test the integration
3. Failed

### Expected Result

See all errors reported to sentry. Just got one randomly, not sure how 🤷 .

### Actual Result

```
[error] Postgrex.Protocol (#PID<0.472.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (REDACTED): connection refused - :econnrefused
[info] [message: "notifier only receiving messages from its own node, functionality may be degraded", source: "oban", event: "notifier:switch", connectivity_status: :solitary]
[error] GenServer {Sentry.Transport.SenderRegistry, 6} terminating
** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom. This protocol is implemented for the following type(s): DBConnection.PrepareStream, DBConnection.Stream, Date.Range, Ecto.Adapters.SQL.Stream, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, Jason.OrderedObject, List, Map, MapSet, Phoenix.LiveView.LiveStream, Postgrex.Stream, Range, Req.Response.Async, Stream, Timex.Interval
    (elixir 1.17.2) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.17.2) lib/enum.ex:166: Enumerable.reduce/3
    (elixir 1.17.2) lib/enum.ex:4423: Enum.map/2
    (sentry 10.7.1) lib/sentry/client.ex:292: anonymous fn/1 in Sentry.Client.render_exception/1
    (sentry 10.7.1) lib/sentry/client.ex:360: Sentry.Client.update_if_present/3
    (elixir 1.17.2) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
    (sentry 10.7.1) lib/sentry/client.ex:360: Sentry.Client.update_if_present/3
    (sentry 10.7.1) lib/sentry/client.ex:283: Sentry.Client.render_event/1
    (sentry 10.7.1) lib/sentry/envelope.ex:61: Sentry.Envelope.item_to_binary/2
    (elixir 1.17.2) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
    (sentry 10.7.1) lib/sentry/envelope.ex:53: Sentry.Envelope.to_binary/1
    (sentry 10.7.1) lib/sentry/transport.ex:23: Sentry.Transport.post_envelope/3
    (sentry 10.7.1) lib/sentry/transport/sender.ex:46: Sentry.Transport.Sender.handle_cast/2
    (stdlib 6.0.1) gen_server.erl:2199: :gen_server.try_handle_cast/3
    (stdlib 6.0.1) gen_server.erl:2261: :gen_server.handle_msg/6
    (stdlib 6.0.1) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
Process Label: Sentry.Transport.Sender
Last message: {:"$gen_cast", {:send, Sentry.HackneyClient, %Sentry.Event{event_id: "0569d2eb222249aebc750a1fc8368031", timestamp: "2024-09-05T22:34:09.632321", breadcrumbs: [], contexts: %{runtime: %{name: "elixir", version: "1.17.2 (compiled with Erlang/OTP 27)"}, os: %{name: "linux", version: "6.6.49"}}, dist: nil, environment: "dev", exception: [%Sentry.Interfaces.Exception{type: "DBConnection.ConnectionError", value: "tcp connect (REDACTED): connection refused - :econnrefused", module: nil, stacktrace: %Sentry.Interfaces.Stacktrace{frames: nil}, mechanism: %Sentry.Interfaces.Exception.Mechanism{handled: false, data: nil, synthetic: nil, help_link: nil, meta: nil, type: "generic"}}], extra: %{domain: [:elixir], logger_level: :error, logger_metadata: %{line: 107, file: "lib/db_connection/connection.ex"}}, fingerprint: ["{{ default }}"], level: :error, logger: nil, message: nil, modules: %{"bcrypt_elixir" => "3.1.0", "iex" => "1.17.2", "esbuild" => "0.8.1", "rustler_precompiled" => "0.8.0", "hpack" => "0.3.0", "castore" => "1.0.8", "mint" => "1.6.2", "ssl_verify_fun" => "1.1.7", "gproc" => "0.9.1", "idna" => "6.1.1", "comeonin" => "5.4.0", "phoenix_template" => "1.0.4", "merge_pdf" => "0.4.0", "plug_crypto" => "2.1.0", "jason" => "1.4.4", "dialyxir" => "1.4.3", "opentelemetry_ecto" => "1.2.0", "dns_cluster" => "0.1.3", "phoenix_html_helpers" => "1.0.1", "eex" => "1.17.2", "ex_cldr_dates_times" => "2.20.3", "oban_met" => "0.1.7", "ex_cldr_calendars" => "1.26.1", "joken" => "2.6.2", "configparser_ex" => "4.0.0", "decorator" => "1.4.0", "gettext" => "0.26.1", "jose" => "1.11.10", "ranch" => "1.8.0", "phoenix_live_view" => "1.0.0-rc.6", "makeup" => "1.1.2", "cowboy" => "2.12.0", ...}, platform: :elixir, release: "0.0.1", request: %Sentry.Interfaces.Request{method: nil, url: nil, query_string: nil, data: nil, cookies: nil, headers: nil, env: nil}, sdk: %Sentry.Interfaces.SDK{name: "sentry-elixir", version: "10.7.1"}, server_name: "matrix", tags: %{}, transaction: nil, threads: nil, user: %{}, integration_meta: %{}, culprit: nil, attachments: [], source: :logger, original_exception: %DBConnection.ConnectionError{message: "tcp connect (REDACTED): connection refused - :econnrefused", severity: :error, reason: :error}}}}
```

savhappy:
-render_exception was missing a conditional and nil was being passed to Enum.map causing a `Protocol.Undefined` error
-closes #786

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -289,7 +289,7 @@ defmodule Sentry.Client do
     |> Map.from_struct()
     |> update_if_present(:mechanism, &Map.from_struct/1)
     |> update_if_present(:stacktrace, fn %Interfaces.Stacktrace{frames: frames} ->
-      %{frames: Enum.map(frames, &Map.from_struct/1)}
+      %{frames: frames && Enum.map(frames, &Map.from_struct/1)}
     end)
   end
 
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -354,10 +354,27 @@ defmodule Sentry.Client do
     end
   end
 
+  defp update_if_present(map, :stacktrace, fun) do
+    case Map.pop(map, :stacktrace) do
+      {nil, _} ->
+        map
+
+      {value, map} ->
+        if is_nil(value.frames) || value.frames == [] do
+          Map.delete(map, :stacktrace)
+        else
+          Map.put(map, :stacktrace, fun.(value))
+        end
+    end
+  end
+
   defp update_if_present(map, key, fun) do
     case Map.pop(map, key) do
-      {nil, _} -> map
-      {value, map} -> Map.put(map, key, fun.(value))
+      {nil, _} ->
+        map
+
+      {value, map} ->
+        Map.put(map, key, fun.(value))
     end
   end
 
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -287,18 +287,24 @@ defmodule Sentry.Client do
   defp render_exception(%Interfaces.Exception{} = exception) do
     exception
     |> Map.from_struct()
+    |> render_stacktrace()
     |> update_if_present(:mechanism, &Map.from_struct/1)
-    |> update_if_present(:stacktrace, fn %Interfaces.Stacktrace{frames: frames} ->
-      %{frames: frames && Enum.map(frames, &Map.from_struct/1)}
-    end)
   end
 
   defp render_thread(%Interfaces.Thread{} = thread) do
     thread
     |> Map.from_struct()
-    |> update_if_present(:stacktrace, fn %Interfaces.Stacktrace{frames: frames} ->
-      %{frames: frames && Enum.map(frames, &Map.from_struct/1)}
-    end)
+    |> render_stacktrace()
+  end
+
+  defp render_stacktrace(map) do
+    case map do
+      %{stacktrace: %{frames: %Interfaces.Stacktrace{frames: [_ | _]} = stacktrace}} ->
+        %{stacktrace | frames: Enum.map(stacktrace.frames, &Map.from_struct/1)}
+
+      map_without_stacktrace ->
+        Map.delete(map_without_stacktrace, :stacktrace)
+    end
   end
 
   defp remove_nils(map) when is_map(map) do
@@ -351,20 +357,6 @@ defmodule Sentry.Client do
     else
       {:ok, _encoded} -> :unchanged
       {:error, _reason} -> {:changed, inspect(value)}
-    end
-  end
-
-  defp update_if_present(map, :stacktrace, fun) do
-    case Map.pop(map, :stacktrace) do
-      {nil, _} ->
-        map
-
-      {value, map} ->
-        if is_nil(value.frames) || value.frames == [] do
-          Map.delete(map, :stacktrace)
-        else
-          Map.put(map, :stacktrace, fun.(value))
-        end
     end
   end
 
--- a/test/sentry/client_test.exs
+++ b/test/sentry/client_test.exs
@@ -84,22 +84,24 @@ defmodule Sentry.ClientTest do
       :code.purge(RaisingJSONClient)
     end
 
-    test "renders threads with stacktrace :frames field set to nil if empty" do
+    test "renders threads with stacktrace property deleted if :frames field set to nil if empty" do
       event =
         Event.create_event(message: "No frames in stacktrace", stacktrace: [])
 
       client = Client.render_event(event)
 
-      assert %{frames: nil} = get_in(client.threads, [Access.at(0), :stacktrace])
+      assert is_nil(get_in(client.threads, [Access.at(0), :stacktrace]))
+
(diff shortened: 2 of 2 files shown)


lib/sentry/client.ex
```diff
@@ -289,7 +289,7 @@ defmodule Sentry.Client do
     |> Map.from_struct()
     |> update_if_present(:mechanism, &Map.from_struct/1)
     |> update_if_present(:stacktrace, fn %Interfaces.Stacktrace{frames: frames} ->
-      %{frames: Enum.map(frames, &Map.from_struct/1)}
+      %{frames: frames && Enum.map(frames, &Map.from_struct/1)}
```
whatyouhide, line 292: Should `frames` be present at all if it's null, according to Sentry docs?

whatyouhide:
We also need a test for this 🙃

lib/sentry/client.ex
whatyouhide: @savhappy need to figure this ☝️ out before merging 🙃

savhappy: Yep yep did some digging in the docs and it says the `frames` attribute is required if stacktrace is present so we should remove the property entirely if `frames: nil`? @whatyouhide thoughts?

whatyouhide: Nice, perfect!

lib/sentry/client.ex
```diff
@@ -354,10 +354,27 @@ defmodule Sentry.Client do
...
+        map
+
+      {value, map} ->
+        if is_nil(value.frames) || value.frames == [] do
+          Map.delete(map, :stacktrace)
+        else
+          Map.put(map, :stacktrace, fun.(value))
+        end
+    end
+  end
```
whatyouhide, line 369: Let's not special-case `update_if_present/3` to handle `:stacktrace`. In general, that leads to somewhat-confusing code because then function behaves differently based on the key, so it's effectively two functions.



You can do this instead:

```elixir
case Map.from_struct(thread) do
  %{stacktrace: %{frames: [_ | _]} = stacktrace} ->
    %{stacktrace | frames: Enum.map(frames, &Map.from_struct/1)}

  thread_as_map ->
    Map.delete(thread_as_map, :stacktrace)
end
```

This is special-cased to handle `:stacktrace` and the `:frames` logic specifically, reads pretty clear IMO.

whatyouhide (changes requested):

lib/sentry/client.ex
```diff
@@ -285,20 +285,24 @@ defmodule Sentry.Client do
   end
 
   defp render_exception(%Interfaces.Exception{} = exception) do
-    exception
-    |> Map.from_struct()
+    case Map.from_struct(exception) do
```
whatyouhide, line 288: Ok I missed that this was done in two places. Let's get this `case` code and extract it into a `render_stacktrace` function then!

savhappy: Done!

whatyouhide (approved):

whatyouhide:
The fix here was incorrect, my bad. We were overriding the exception map completely. Tests were missing. I fixed this in https://github.com/getsentry/sentry-elixir/commit/d79919fdca3be3c7b488d065bd1da7175de3c11a @savhappy.

savhappy:
Ah I see that! Thank you for the correction and the updated tests!

Issue #788, opened 2024-09-07
Sending a test event fails with in `:proplists.get_value/3`


rlopzc:
### Environment

What version are you running? 10.7.1

### Steps to Reproduce

1. Configure prod
2. run `MIX_ENV=prod mix sentry.send_test_event`

### Expected Result

Test event is sent.

### Actual Result

An error is raised in the console.

```
Client configuration:
server: REDACTED/envelope/
public_key: REDACTED
secret_key:
current environment_name: "prod"
hackney_opts: [pool: :sentry_pool]

Sending test event...
** (FunctionClauseError) no function clause matching in :proplists.get_value/3

    The following arguments were given to :proplists.get_value/3:

        # 1
        "X-Sentry-Error"

        # 2
        %ArgumentError{message: "unknown registry: Req.Finch"}

        # 3
        nil

    (stdlib 6.0.1) proplists.erl:214: :proplists.get_value/3
    (sentry 10.7.1) lib/sentry/client.ex:386: Sentry.Client.maybe_log_send_result/2
    (sentry 10.7.1) lib/sentry/client.ex:251: Sentry.Client.encode_and_send/4
    (sentry 10.7.1) lib/sentry/client.ex:137: Sentry.Client.send_event/2
    (sentry 10.7.1) lib/mix/tasks/sentry.send_test_event.ex:88: Mix.Tasks.Sentry.SendTestEvent.send_event/1
    (mix 1.17.2) lib/mix/task.ex:495: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.17.2) lib/mix/cli.ex:96: Mix.CLI.run_task/2
    /Users/rlopez/.asdf/installs/elixir/1.17.2-otp-27/bin/mix:2: (file)
```

My config:
```
config :sentry,
  client: Gru.Sentry.HttpClient,
  release: System.get_env("APP_VERSION", "0.0.1"),
  dsn: "REDACTED",
  environment_name: config_env(),
  enable_source_code_context: true,
  root_source_code_paths: [File.cwd!()],
  report_deps: false,
  integrations: [
    oban: [
      # Capture errors:
      capture_errors: true,
      # Monitor cron jobs:
      cron: [enabled: true]
    ]
  ]
```

My HTTP client:
```elixir
defmodule App.Sentry.HttpClient do
  @moduledoc """
  Sentry HTTPClient using Req
  """
  @behaviour Sentry.HTTPClient

  @impl Sentry.HTTPClient
[...]

PR #789, opened 2024-09-10
update maybe_log_send_result/2 with proper error

savhappy:
-When `Sentry.ClientError` was created, it was not passed updated in the `maybe_log_send_result/2` function
-This was originally found from debugging issue #788

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -308,7 +308,7 @@ defmodule Sentry.Client do
     message =
       case send_result do
         {:error, {:invalid_json, error}} ->
-          "Unable to encode JSON Sentry error - #{inspect(error)}"
+          "Failed to send Sentry event. Unable to encode JSON Sentry error - #{inspect(error)}"
 
         {:error, {:request_failure, last_error}} ->
           case last_error do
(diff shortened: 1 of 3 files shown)


whatyouhide:
@savhappy can you refactor this so that we only have *one* `maybe_log_send_result`? I can't recall why we have two. Build the error wherever it makes sense, but then do the logging only in one place.

savhappy:
Yes @whatyouhide the easiest is to make it public in `Sentry.Client` module. Or we can change the arrity of `Sentry.Transport.post_envelope/3` to accommodate the two uses. Which path is preferred?

whatyouhide:
> Or we can change the arity of `Sentry.Transport.post_envelope/3` to accommodate the two uses. Which path is preferred?

Can you elaborate?

savhappy:
Actually, I can just grab what I need from the `Sentry.Envelope` struct to pass the events to `maybe_log_send_result`. So it would look like this:


``` elixir
  def post_envelope(%Envelope{} = envelope, client, retries \\ @default_retries)
      when is_atom(client) and is_list(retries) do
    result =
      case Envelope.to_binary(envelope) do
        {:ok, body} ->
          {endpoint, headers} = get_endpoint_and_headers()

          post_envelope_with_retries(client, endpoint, headers, body, retries)

        {:error, reason} ->
          {:error, {:invalid_json, reason}}
      end

    _ = maybe_log_send_result(result, envelope.items)
  end

```

Does that seem ideal?

savhappy:
But it breaks all of the cron tests... maybe making this function public is best? @whatyouhide

whatyouhide:
Let's make this public and see how it looks like.

savhappy:
Updates were made! We still need a test for `:too_many_retries` but the refactored logic should handle the bug mentioned in #792

whatyouhide (approved):

PR #793, opened 2024-09-14
Add more tests to `Sentry.Client`

savhappy:
-add test to retries path
-add spec to maybe_log_send_result


savhappy:
Updates from pairing session

whatyouhide (approved):

PR #794, opened 2024-09-15
Header validation

savhappy:
-add simple header validation and test

### DIFF
--- a/lib/sentry/transport.ex
+++ b/lib/sentry/transport.ex
@@ -58,7 +58,7 @@ defmodule Sentry.Transport do
   end
 
   defp request(client, endpoint, headers, body) do
-    with {:ok, 200, _headers, body} <- client.post(endpoint, headers, body),
+    with {:ok, 200, _headers, body} <- post_validation(client, endpoint, headers, body),
          {:ok, json} <- Config.json_library().decode(body) do
       {:ok, Map.get(json, "id")}
     else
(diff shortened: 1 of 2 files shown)


lib/sentry/transport.ex
```diff
@@ -86,6 +86,22 @@ defmodule Sentry.Transport do
     kind, data -> {:error, {kind, data, __STACKTRACE__}}
   end
 
+  defp post_validation(client, endpoint, headers, body) do
```
whatyouhide, line 89: We should also validate that the status is an integer, the body is a string, and headers are a list.

```elixir
case client.post(...) do
  {:ok, status, resp_headers, resp_body} when is_integer(status) and status in 200..599 and is_list(resp_headers) and is_binary(resp_body) ->
```

whatyouhide, line 89: Also maybe we should call this `client_post_and_validate_return_value`

savhappy:
Changes made! @whatyouhide

lib/sentry/transport.ex
```diff
@@ -86,6 +87,21 @@ defmodule Sentry.Transport do
...
 
+  defp client_post_and_validate_return_value(client, endpoint, headers, body) do
+    case client.post(endpoint, headers, body) do
+      {:ok, status, resp_headers, resp_body}
+      when is_integer(status) and status in 200..599 and is_list(resp_headers) and
+             is_binary(resp_body) ->
+        {:ok, status, resp_headers, resp_body}
+
+      {:ok, _status, _resp_headers, _resp_body} ->
+        {:error, :invalid_http_response}
```
whatyouhide, line 98: I don't think this is a fair error code: we don't know that the HTTP response is invalid. What's invalid here is that the HTTP client didn't comply to the expected behaviour. What I would do here is return

```elixir
{:error, {:malformed_http_client_response, status, resp_headers, resp_body}}`
```

so that it makes it sort of understandable for people?

savhappy: Yes, much better 💟

whatyouhide (approved):

PR #797, opened 2024-09-17
Add full e2e phoenix app test

solnic:
Here's a test setup that does e2e testing with a real phoenix app. This will be handy when testing more complex integrations like OTel stuff for a good start. It should also help us reduce the risk of introducing regressions in Phoenix apps (once we add more tests of course).

The only custom "trick" I had to use was defining endpoint in test/support because I made sentry depend on the phoenix_app from test/fixtures/phoenix_app and the endpoint needs access to Sentry.CapturePlug.

Another nice benefit is that we could set it up in the Sentry SDK project and it would make it simpler to test things out w/o having to create new setups manually.


whatyouhide:
I’m wondering if a better way to do this would be to instead have a Phoenix app in `integration_tests` or something, and then have a specific Mix task for testing that app (`mix test` in the app basically). That app can then depend on `{:sentry, path: ".."}`.

https://github.com/elixir-ecto/ecto_sql does all this setup and has been running it successfully for a while, so it could be a good source of inspiration,

solnic:
@whatyouhide yeah this makes sense, thanks for suggestion!

solnic:
> ==> Running tests for integration: phoenix_app
** (Mix) You're trying to run :phoenix_app on Elixir v1.13.4 but it has declared in its mix.exs file it supports only Elixir ~> 1.14
Integration tests for phoenix_app failed
Error: Process completed with exit code 1.

So, what's our strategy here? Can we drop 1.13 support given that more recent Phoenix versions do not support it?

solnic:
I made `mix test` skip running test.integrations when Elixir version is < 1.16 for the time being.

mix.exs
```diff
@@ -123,6 +128,54 @@ defmodule Sentry.Mixfile do
   end
 
   defp aliases do
-    [test: ["sentry.package_source_code", "test"]]
+    [
+      test: ["sentry.package_source_code", "test", &run_integration_tests_if_supported/1],
```
whatyouhide, line 132: Principle of least surprise: I think I would rather keep `mix test`'s behavior what we would expect it to be, which is run the repo's tests, and then have `mix test.integrations` separately (like you do now). This way users have to run it manually locally, but that's likely fine as this is more of a e2e test. CI will run it anyways, so it will catch broken stuff if that happens.

```suggestion
      test: ["sentry.package_source_code", "test"],
```

mix.exs
```diff
@@ -123,6 +128,54 @@ defmodule Sentry.Mixfile do
...
+    if Version.match?(System.version(), ">= 1.16.0") do
+      Mix.Task.run("test.integrations")
+    else
+      IO.puts("Skipping integration tests for Elixir versions < 1.16")
+    end
+  end
+
+  @integrations [
+    "phoenix_app"
+  ]
```
whatyouhide, line 147: Maybe we can keep this hardcoded for now as we don't have other integrations?

mix.exs
```diff
@@ -123,6 +128,54 @@ defmodule Sentry.Mixfile do
...
+
+  defp test_integrations(args) do
+    if Version.match?(System.version(), ">= 1.16.0") do
+      for integration <- @integrations do
+        run_integration_tests(integration, args)
+      end
+    else
+      IO.puts("Skipping integration tests for Elixir versions < 1.16")
```
whatyouhide, line 155:
```suggestion
      Mix.shell().info("Skipping integration tests for Elixir versions < 1.16")
```

mix.exs
```diff
@@ -123,6 +128,54 @@ defmodule Sentry.Mixfile do
...
+    end
+  end
+
+  @integrations [
+    "phoenix_app"
+  ]
+
+  defp test_integrations(args) do
+    if Version.match?(System.version(), ">= 1.16.0") do
```
whatyouhide, line 150: Sweet!

mix.exs
```diff
@@ -123,6 +128,54 @@ defmodule Sentry.Mixfile do
...
+        run_integration_tests(integration, args)
+      end
+    else
+      IO.puts("Skipping integration tests for Elixir versions < 1.16")
+    end
+  end
+
+  defp run_integration_tests(integration, args) do
+    header = if IO.ANSI.enabled?(), do: IO.ANSI.bright() <> IO.ANSI.cyan(), else: ""
+    reset = if IO.ANSI.enabled?(), do: IO.ANSI.reset(), else: ""
```
whatyouhide, line 161: You can use `IO.ANSI.format/1`, whose second (default) argument is `IO.ANSI.enabled?()`, so that it magiworks and you can do

```elixir
IO.puts(IO.ANSI.format(["\n", :bright, :cyan, "==> Running tests for integration: #{integration}"])
```

(also resets at the end)

whatyouhide: Same below btw

mix.exs
```diff
@@ -123,6 +128,54 @@ defmodule Sentry.Mixfile do
...
+
+    IO.puts("\n#{header}==> Running tests for integration: #{integration}#{reset}")
+
+    integration_dir = Path.join("test_integrations", integration)
+
+    {_, status} =
+      System.cmd("sh", ["-c", "cd #{integration_dir} && mix test #{Enum.join(args, " ")}"],
+        into: IO.binstream(:stdio, :line)
+      )
```
whatyouhide, line 170: Instead of `cd <> && ...`, you can use the `:cd` option in `System.cmd` and you won't have to use `sh`:

```suggestion
    {_, status} =
      System.cmd("mix", ["test" | args],
        into: IO.binstream(:stdio, :line),
        cd: Path.join("test_integrations", integration)
      )
```

whatyouhide: Also, this snippet is from another repo I have and it's handy to show nice test output when running in a subshell, cause it defaults to no colors:

```elixir
color_arg = if IO.ANSI.enabled?(), do: "--color", else: "--no-color"

System.cmd("mix", ["test", color_arg | args], ...)
```

test/test_helper.exs
```diff
@@ -1,4 +1,4 @@
-ExUnit.start(assert_receive_timeout: 1000)
+ExUnit.start()
```
whatyouhide, line 1: Why this change?

test_integrations/phoenix_app/mix.exs
```diff
@@ -0,0 +1,80 @@
...
+      {:telemetry_metrics, "~> 1.0"},
+      {:telemetry_poller, "~> 1.0"},
+      {:gettext, "~> 0.20"},
+      {:jason, "~> 1.2"},
+      {:dns_cluster, "~> 0.1.1"},
+      {:bandit, "~> 1.5"},
+      {:bypass, "~> 2.1", only: :test},
+      {:hackney, "~> 1.18", only: :test},
+
+      {:sentry, path: "../.."}
```
whatyouhide, line 59: Noooooice

whatyouhide (changes requested):

test/test_helper.exs
solnic: @whatyouhide oops, accidental left-over

solnic:
@whatyouhide thank you so much for all the suggestions! I applied all of it + rebased/squashed + updated #784 (which waits for reviews too 🙏🏻)

mix.exs
```diff
@@ -83,6 +85,9 @@ defmodule Sentry.Mixfile do
   defp elixirc_paths(:test), do: ["test/support"] ++ elixirc_paths(:dev)
   defp elixirc_paths(_other), do: ["lib"]
 
+  defp test_paths(nil), do: ["test"]
+  defp test_paths(integration), do: ["test_integrations/#{integration}/test"]
```
whatyouhide, line 89: I like this.

whatyouhide (commented):
I was testing this locally, and noticed these issues:

1. The first time I ran `mix test.integrations`, it didn't work because I had to go into the Phoenix app directory and run `mix deps.get`. I think we should shove that into the Mix alias itself.

2. Can we fix this compilation warning within the Phoenix app?

```
==> phoenix_app
Compiling 18 files (.ex)
warning: defining a Gettext backend by calling

    use Gettext, otp_app: ...

is deprecated. To define a backend, call:

    use Gettext.Backend, otp_app: :my_app

Then, instead of importing your backend, call this in your module:

    use Gettext, backend: MyApp.Gettext

  lib/phoenix_app_web/gettext.ex:23: PhoenixAppWeb.Gettext (module)
```

The fix is exactly what it says it is :smile:

whatyouhide (approved):

PR #799, opened 2024-09-19
Refactor `maybe_log_send_result`

savhappy:
-refactor

### DIFF
--- a/lib/sentry/transport.ex
+++ b/lib/sentry/transport.ex
@@ -134,7 +134,7 @@ defmodule Sentry.Transport do
     {dsn.endpoint_uri, auth_headers}
   end
 
-  def maybe_log_send_result(send_result, events) do
+  defp maybe_log_send_result(send_result, events) do
     if Enum.any?(events, &(Map.has_key?(&1, :source) && &1.source == :logger)) do
       :ok
     else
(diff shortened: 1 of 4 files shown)


savhappy:
@whatyouhide I'm getting hung up on this -> When I move `maybe_log_send_result` to the `Sentry.Transport.encode_and_post_envelope` I get many failing test with the corn integrations:

```
test start event and same ref stop event have same check-in id (Sentry.Integrations.Quantum.CronTest)
     test/sentry/integrations/quantum/cron_test.exs:233
     Assertion failed, no matching message after 1000ms
     The following variables were pinned:
       ref = #Reference<0.2513758090.1975255046.33127>
     The process mailbox is empty.
     code: assert_receive {^ref, :done}
     stacktrace:
       test/sentry/integrations/quantum/cron_test.exs:253: (test)
```

Advice?

whatyouhide:
@savhappy inside `Bypass.expect_once` on the failing test, put `dbg(check_in_body)` for example to figure out why it's not matching. That error means something within `expect_once` failed and the `:done` message wasn't sent.

lib/sentry/transport.ex
```diff
@@ -130,4 +133,21 @@ defmodule Sentry.Transport do
 
     {dsn.endpoint_uri, auth_headers}
   end
+
+  def maybe_log_send_result(send_result, events) do
```
whatyouhide, line 137: Does this need to be public?

whatyouhide, line 137:
```suggestion
  defp maybe_log_send_result(send_result, events) do
```

savhappy: Nope, thanks for the catch!

whatyouhide (approved):

PR #801, opened 2024-09-25
Add support for client reports

savhappy:


savhappy:
@whatyouhide yep yep I'll tag you here when ready

savhappy:
@whatyouhide ready for an initial review. I've only added a single instance of recording an event but will add in more after discussion

lib/sentry/application.ex
```diff
@@ -19,6 +19,12 @@ defmodule Sentry.Application do
         []
       end
 
+    # if Config.dsn() && Config.send_client_reports?() do
+    client_report_enabled =
+      if Config.send_client_reports?() do
+        [Sentry.ClientReport]
```
whatyouhide, line 25: Cheap enough to just start this anyway, and also allows for runtime changing of `:send_client_reports`. Let's always start `Sentry.ClientReport`.

lib/sentry/config.ex
```diff
@@ -147,6 +147,14 @@ defmodule Sentry.Config do
       [`:jason`](https://hex.pm/packages/jason) as a dependency of your application.
       """
     ],
+    send_client_reports: [
+      type: :boolean,
+      default: true,
+      doc: """
+      Send SDK Client Reports, this option defaults to `true`
+      *Available since v10.0.0*.
```
whatyouhide, line 155: 1. nimble_options already mentions the default value in the generated doc.
2. This is not available since 10.0.0, but rather will be available from 10.8.0.
3. Let's link to Client Reports documentation.

lib/sentry/event.ex
```diff
@@ -78,7 +78,6 @@ defmodule Sentry.Event do
           sdk: Interfaces.SDK.t() | nil,
           threads: [Interfaces.Thread.t()] | nil,
           user: Interfaces.user() | nil,
-
```
whatyouhide, line 81: Can you leave these lines in? Not really a necessary change here 🙃 Alos the one below.

lib/sentry/envelope.ex
```diff
@@ -2,11 +2,11 @@ defmodule Sentry.Envelope do
   @moduledoc false
   # https://develop.sentry.dev/sdk/envelopes/
 
-  alias Sentry.{Attachment, CheckIn, Config, Event, UUID}
+  alias Sentry.{Attachment, CheckIn, Config, Event, UUID, ClientReport}
```
whatyouhide, line 5: Super duper nit: can you keep the alphabetic sorting?

lib/sentry/envelope.ex
```diff
@@ -34,6 +34,17 @@ defmodule Sentry.Envelope do
     }
   end
 
+  @doc """
+  Creates a new envelope containing the client report and a map of discarded events.
```
whatyouhide, line 38:
```suggestion
  Creates a new envelope containing the client report.
```

lib/sentry/envelope.ex
```diff
@@ -92,4 +103,15 @@ defmodule Sentry.Envelope do
         throw(error)
     end
   end
+
+  defp item_to_binary(json_library, %ClientReport{} = client_report) do
+    case client_report |> Map.from_struct() |> json_library.encode() do
+      {:ok, encoded_client_report} ->
+        header = ~s({"type": "client_report", "length": #{byte_size(encoded_client_report)}})
```
whatyouhide, line 110: This is silly and I’m noticing I didn't do it for `check_in`s, but let's make the JSON ever so slightly smaller 🙃 

```suggestion
        header = ~s({"type":"client_report","length":#{byte_size(encoded_client_report)}})
```

lib/sentry/envelope.ex
```diff
@@ -34,6 +34,17 @@ defmodule Sentry.Envelope do
     }
   end
 
+  @doc """
+  Creates a new envelope containing the client report and a map of discarded events.
+  """
+  @spec from_client_report(ClientReport.t()) :: t()
```
whatyouhide, line 40: This module is private but since we have a doc maybe this helps:

```suggestion
  @doc since: "10.8.0"
  @spec from_client_report(ClientReport.t()) :: t()
```

lib/sentry/envelope.ex
```diff
@@ -34,6 +34,17 @@ defmodule Sentry.Envelope do
...
+  def from_client_report(%ClientReport{} = client_report) do
```
whatyouhide, line 41: **Blocking**: we need a test for this function and a test for `to_binary/1` with a client report as arg.

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
+defmodule Sentry.ClientReport do
+  @moduledoc """
+  **TODO** add proper doc for module
+
+  See <https://develop.sentry.dev/sdk/client-reports/>.
+  """
+
+  @moduledoc since: "10.0.0"
```
whatyouhide, line 8: Everywhere here we need `10.8.0` instead.

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  @typedoc since: "10.0.0"
```
whatyouhide, line 10: This is not needed with `@moduledoc since`

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  use GenServer
+  alias Sentry.Client
+
+  @typedoc """
+  The possible reasons of the discarded event.
+  """
+  @type reasons() ::
```
whatyouhide, line 18: The type represents what a *single* reason can be, not multiple reasons. Change this to:

```suggestion
  @type reason() ::
```

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+          | :insufficient_data
+          | :backpressure
+
+  @typedoc """
+  The struct for a **client report** interface.
+  """
+  @type t() :: %__MODULE__{
+          timestamp: String.t() | number(),
+          discarded_events:
+            list(%{reason: reasons(), category: String.t(), quantity: pos_integer()})
```
whatyouhide, line 35: Nit, but `list/1` is somewhat less common than just:

```suggestion
            [%{reason: reasons(), category: String.t(), quantity: pos_integer()}, ...]       
```

Also added `, ...` because a client report with no discarded events makes no sense, does it? (Honest question, I really don't know)

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+    :ratelimit_backoff,
+    :queue_overflow,
+    :cache_overflow,
+    :network_error,
+    :sample_rate,
+    :before_send,
+    :event_processor,
+    :insufficient_data,
+    :backpressure
+  ]
```
whatyouhide, line 50: To make the `@type t` easier to keep in sync with this list, move this module attr above the type and then define the type as

```elixir
@type type() :: unquote(Enum.reduce(@client_report_reasons, &quote(do: unquote(&1) | unquote(&2))))
```

Nifty trick!

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  @send_interval 5000
+
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link(_) do
```
whatyouhide, line 55: To make sure we don't pass in unused options:

```suggestion
  def start_link([]) do
```

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  @spec add_discarded_event({reasons(), String.t()}) :: :ok
+  def add_discarded_event({reason, category}) do
+    if Enum.member?(@client_report_reasons, reason) do
+      GenServer.cast(__MODULE__, {:add_discarded_event, {reason, category}})
+    end
+
+    :ok
+  end
```
whatyouhide, line 67: Why did you choose to silently ignore events whose `reason` is not a valid one?

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+  @typedoc """
+  The struct for a **client report** interface.
+  """
+  @type t() :: %__MODULE__{
+          timestamp: String.t() | number(),
+          discarded_events:
+            list(%{reason: reasons(), category: String.t(), quantity: pos_integer()})
+        }
+
+  defstruct [:timestamp, :discarded_events]
```
whatyouhide, line 38:
```suggestion
  defstruct [:timestamp, discarded_events: %{}]
```

No need to initialize the state with the `discarded_events == nil` check if you set a default like ☝️

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  @impl true
+  def handle_cast({:add_discarded_event, {reason, category}}, client_report) do
+    if client_report.discarded_events == nil do
+      {:noreply, %{client_report | discarded_events: %{{reason, category} => 1}}}
+    else
+      discarded_events =
+        Map.update(client_report.discarded_events, {reason, category}, 1, &(&1 + 1))
+
+      {:noreply, %{client_report | discarded_events: discarded_events}}
```
whatyouhide, line 83: Nit, but this way we make sure we check for typos in field names at compile time.

```suggestion
      {:noreply, %__MODULE__{client_report | discarded_events: discarded_events}}
```

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+    end
+  end
+
+  @impl true
+  def handle_info(:send_report, state) do
+    if state.discarded_events != nil do
```
whatyouhide, line 89: If you set the default to `%{}`, make sure to change this check to `if map_size(discarded_events) == 0`.

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,128 @@
...
+    end
+  end
+
+  defp schedule_report do
+    Process.send_after(self(), :send_report, @send_interval)
+  end
+
+  defp timestamp do
+    DateTime.utc_now()
+    |> DateTime.truncate(:microsecond)
```
whatyouhide, line 113: Why truncating here? Do you need microseconds in the date? Or do you need them to be `0`?

whatyouhide (changes requested):
Left a first pass of comments, some are nits, some are things to address, and a couple are blocking. Let me know if you wanna go through them together 🙃

lib/sentry/client_report.ex
savhappy: I decided to do this for two reasons -> From what I can tell, the Ruby SDK ignores invalid reasons 

``` ruby
    def record_lost_event(reason, data_category, num: 1)
      return unless @send_client_reports
      return unless CLIENT_REPORT_REASONS.include?(reason)

      @discarded_events[[reason, data_category]] += num
    end   
```

and the documentation says 

`In case a reason needs to be added, it also has to be added to the allowlist in [snuba](https://github.com/getsentry/snuba/blob/4e7cfdddcf7b93eacb762bc74ca2461cec9464e5/snuba/datasets/outcomes_processor.py#L24-L34).`
    
So we choose to do nothing if it's invalid unless we want to add them to this repo mentioned. Thoughts?

lib/sentry/client_report.ex
savhappy: wip. I matched another part of the code in this repo where we send a timestamp. The example shows the timestamp reported as this in the payload 

`"2020-02-07T14:16:00Z"` 

So you're right. `:second` is better

savhappy:
Thanks @whatyouhide for this! Updated with these comments except the blocking tests. Will add those shortly. Hold off on more feedback till I get the next round out. I need to add `data_category` to this somewhere too and maybe a `capture_event` function to direct how this report is classified.

savhappy:
Question @whatyouhide : I need to add in the appropriate [`data_category`](https://github.com/getsentry/relay/blob/master/relay-base-schema/src/data_category.rs#L101) to the payload. We add the type only in the headers but other SDK's have a type field. The mappings for types would look like this 

``` elixir
defmodule Sentry.DataCategory do
  @spec data_category_mappings(String.t()) :: String.t()
  def data_category_mappings(type) do
    case type do
      "session" -> "session"
      "sessions" -> "session"
      "attachment" -> "attachment"
      "transaction" -> "transaction"
      "profile" -> "profile"
      "span" -> "span"
      "check_in" -> "monitor"
      "statsd" -> "metric_bucket"
      "metric_meta" -> "metric_bucket"
      "event" -> "error"
      "client_report" -> "internal"
      _ -> "default"
    end
  end
end
```

But we only have the types `Event, Attachment, CheckIn, ClientReport` (and soon Transaction). So whats the best way to go about deriving the type that will age well as we add more in the future? Or should I just get it based on the struct that's passed and if its not one of the structs see if it fits a string that's passed as the arg for data_category instead? I hope this makes sense. If not we can go into detail on Tuesday.

whatyouhide:
@savhappy can you elaborate on the issue here? Where do we need to add `data_category`? And *what* is the data category?

savhappy:
Yep! The data categories are used to classify the type of data being ingested.

It's defined in the Client Report docs as -> `category: The data category for which the discard reason applies. These are the same data categories used for [rate limits](https://develop.sentry.dev/sdk/rate-limiting/#definitions).`

We need to add the data_category to the payload -> 

```
 "discarded_events": [
    {
      "reason": "queue_overflow",
      "category": "transaction",
      "quantity": 1
    }
```
and I think its the count used for metrics. So When I send {reason, category} over to the genserver. I need to ensure that I'm sending over the correct data category to Sentry. 

In Transport module -> 
``` elixir

  def record_discarded_event(reason, category) do
    # add logic for data category here
    # call to client report genserver
    ClientReport.add_discarded_event({reason, category})
    # end

    :ok
  end
  ```

File: lib/sentry/client.ex [hunk 20]
```diff
@@ -91,6 +92,24 @@ defmodule Sentry.Client do
     end
   end
 
+  @spec send_client_report(ClientReport.t()) ::
+          {:ok, client_report_id :: String.t()} | {:error, ClientError.t()}
+  def send_client_report(%ClientReport{} = client_report) do
+    # We don't pass options because this is internal. But should we use default client or users?
+    client = Config.client()
```
whatyouhide, line 99: No, always use the client configured by the user. They might not even have the default client as a dependency.

lib/sentry/client.ex
```diff
@@ -91,6 +92,24 @@ defmodule Sentry.Client do
...
+    # This is a "private" option, only really used in testing.
+    request_retries =
+      Application.get_env(:sentry, :request_retries, Transport.default_retries())
+
+    send_result =
+      client_report
+      |> Envelope.from_client_report()
+      |> Transport.encode_and_post_envelope(client, request_retries)
+
+    send_result
```
whatyouhide, line 110: Same as:

```suggestion
    client_report
    |> Envelope.from_client_report()
    |> Transport.encode_and_post_envelope(client, request_retries)
```

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+  defstruct [:timestamp, discarded_events: %{}]
+
+  @send_interval 5000
+
+  @spec start_link([]) :: GenServer.on_start()
+  def start_link([]) do
+    # check config to see if send_client_report is true
+    GenServer.start_link(__MODULE__, %__MODULE__{}, name: __MODULE__)
```
whatyouhide, line 46: Using `%__MODULE__{}` as the state is confusing IMO because you're conflating two responsibilities in the same place: now `%__MODULE__{}` is both the data definition of the client report, as well as the GenServer's internal state.

Instead, I'll go against my own advice and say: just use the map of discarded events as the GenServer's state, and only add the timestamp when sending the client report. Makes sense?

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,120 @@
...
+          timestamp: String.t() | number(),
+          discarded_events: [%{reason: reason(), category: String.t(), quantity: pos_integer()}]
+        }
+
+  defstruct [:timestamp, discarded_events: %{}]
+
+  @send_interval 5000
+
+  @spec start_link([]) :: GenServer.on_start()
+  def start_link([]) do
```
whatyouhide, line 44: If you choose to go with `Sentry.ClientReport` containing both the struct as well as the GenServer, you'll need to make all these functions private (`@doc false`).

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,120 @@
...
+    # check config to see if send_client_report is true
```
whatyouhide, line 45: Is this a TODO?

lib/sentry/client_report.ex
whatyouhide: Okay, sounds good. Let's silently ignore it for now, BUT leave a comment with basically what you just said to me, but in the code 🙃

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,120 @@
...
+  @impl true
+  def init(state) do
+    schedule_report()
+    {:ok, state}
+  end
+
+  @impl true
+  def handle_cast({:add_discarded_event, {reason, category}}, client_report) do
+    if map_size(client_report.discarded_events) == 0 do
+      {:noreply, %{client_report | discarded_events: %{{reason, category} => 1}}}
```
whatyouhide, line 67: This code is not needed if you use `Map.update/4` the way you do below. If the map is empty, you'll init `{reason, category}` to the third arg of `Map.update/4` (`1`, correctly). Plz remove the `if` 🙃

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,120 @@
...
+      discarded_events =
+        Map.update(client_report.discarded_events, {reason, category}, 1, &(&1 + 1))
+
+      {:noreply, %__MODULE__{client_report | discarded_events: discarded_events}}
+    end
+  end
+
+  @impl true
+  def handle_info(:send_report, state) do
+    if map_size(state.discarded_events) == 0 do
```
whatyouhide, line 78: Opposite: if map is empty, then *don't* report?

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,120 @@
...
+  defp transform_map(discarded_events_map) do
+    discarded_events_map
+    |> Enum.map(fn {{reason, category}, quantity} ->
+      %{
+        reason: reason,
+        category: category,
+        quantity: quantity
+      }
+    end)
+  end
```
whatyouhide, line 119: Nit, but you can inline this in `handle_info(:send_report, ...)` I think. It's a small function and `transform_map` is a very generic name, which means readers will have to jump around to see what it does.

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,120 @@
...
+  @send_interval 5000
+
+  @spec start_link([]) :: GenServer.on_start()
+  def start_link([]) do
+    # check config to see if send_client_report is true
+    GenServer.start_link(__MODULE__, %__MODULE__{}, name: __MODULE__)
+  end
+
+  @spec add_discarded_event({reason(), String.t()}) :: :ok
+  def add_discarded_event({reason, category}) do
```
whatyouhide, line 50: Don't use a tuple here, it's not needed. Just pass two args, `(reason, category)` and call the GenServer cast `{:add_discarded_event, reason, category}`.

lib/sentry/transport.ex
```diff
@@ -40,6 +40,11 @@ defmodule Sentry.Transport do
     result
   end
 
+  def record_discarded_event(reason, category) do
+    ClientReport.add_discarded_event({reason, DataCategory.data_category_mapping(category)})
```
whatyouhide, line 44: Why go through the extra `data_category_mapping/1` step here rather than just do the mapping in `ClientReport`, without introducing a new `Sentry.DataCategory` module?

whatyouhide (commented):
Left one more round of comments, this is shaping up! 🎉

lib/sentry/client_report.ex
savhappy: Updated but double check that this is what you meant -> 

``` elixir
@doc false
  @spec start_link([]) :: GenServer.on_start()
  def start_link([]) do
    # check config to see if send_client_report is true
    GenServer.start_link(__MODULE__, %{}, name: __MODULE__)
  end
```

lib/sentry/client_report.ex
savhappy: Yep. If the user chooses to flip to false then we don't report

lib/sentry/client_report.ex
savhappy: Whoops! Yes, thx!

lib/sentry/transport.ex
savhappy: I believe we resolved this in our pairing. Updates made so lemme know thoughts

lib/sentry/envelope.ex
savhappy: Still need to add this test and others. I'll add more test after the logic stops moving so maybe one more round of feedback from @whatyouhide

lib/sentry.ex
```diff
@@ -411,6 +412,7 @@ defmodule Sentry do
       |> CheckIn.new()
       |> Client.send_check_in(options)
     else
+      # add record dropped event here? Dropped due to....
```
whatyouhide, line 415: Ask on Discord because I don't think we need to report these here but maybe other SDKs do. Remove this comment in the meantime so that we can start shipping this PR.

lib/sentry/client.ex
```diff
@@ -81,16 +82,34 @@ defmodule Sentry.Client do
       :unsampled ->
         # See https://github.com/getsentry/develop/pull/551/files
         Sentry.put_last_event_id_and_source(event.event_id, event.source)
+        Transport.record_discarded_event(:sample_rate, "error")
         :unsampled
 
       :excluded ->
+        # do we need to add this if an event is dropped due to it being a duplicate??
```
whatyouhide, line 89: Ask in Discord

lib/sentry/envelope.ex
```diff
@@ -34,6 +34,41 @@ defmodule Sentry.Envelope do
     }
   end
 
+  @doc """
+  Creates a new envelope containing the client report.
+  since: "10.8.0"
```
whatyouhide, line 39: `@doc since` is an ExDoc feature. I literally meant:

```elixir
@doc since: "10.8.0"
```

after the `@doc` string. There are tons of examples in this repo 🙃

lib/sentry/envelope.ex
whatyouhide: This part of the logic won't change because encoding is simple enough here, so you can go ahead and add tests for this anytime

lib/sentry/envelope.ex
```diff
@@ -34,6 +34,41 @@ defmodule Sentry.Envelope do
...
+  def from_client_report(%ClientReport{} = client_report) do
+    %__MODULE__{
+      event_id: UUID.uuid4_hex(),
+      items: [client_report]
+    }
+  end
+
+  def get_data_category(type) do
+    if is_binary(type) do
+      type
```
whatyouhide, line 51: When will this ever be a string? We talked about functions having a small "type" surface but this function is still accepting `binary | Attachment.t | ...`. Can we make it so this only accepts a struct OR a binary?

Also, can you add typespecs and tests for this?

lib/sentry/envelope.ex
```diff
@@ -34,6 +34,41 @@ defmodule Sentry.Envelope do
...
+          "monitor"
+
+        %ClientReport{} ->
+          "internal"
+
+        %Event{} ->
+          "error"
+
+        _ ->
+          "default"
```
whatyouhide, line 67: This is dangerous because I can call `get_data_category(:completely_banana_value)`. Make sure the input is either a string or a struct (per comm
ent above)

lib/sentry/transport.ex
```diff
@@ -29,18 +29,39 @@ defmodule Sentry.Transport do
...
-  defp post_envelope_with_retries(client, endpoint, headers, payload, retries_left) do
+  def record_discarded_event(reason, event_item) do
+    _ =
+      if is_list(event_item) do
+        Enum.map(event_item, fn event ->
+          ClientReport.add_discarded_event({reason, Envelope.get_data_category(event)})
+        end)
+      else
+        ClientReport.add_discarded_event({reason, Envelope.get_data_category(event_item)})
+      end
```
whatyouhide, line 50: 1. `List.wrap/1`
2. `Enum.map/2` wastes memory here, use `Enum.each/2` since we ignore the result.

```suggestion
    _ =
      event_item
      |> List.wrap()
      |> Enum.each(&ClientReport.add_discarded_event({reason, Envelope.get_data_category(&1)}))
```

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,115 @@
...
+  @send_interval 30_000
+
+  @doc false
+  @spec start_link([]) :: GenServer.on_start()
+  def start_link([]) do
+    GenServer.start_link(__MODULE__, %{}, name: __MODULE__)
+  end
+
+  @spec add_discarded_event(reason(), String.t()) :: :ok
+  def add_discarded_event(reason, category) do
```
whatyouhide, line 50: This is internal for now so `@doc false`

lib/sentry/envelope.ex
```diff
@@ -34,6 +34,42 @@ defmodule Sentry.Envelope do
...
+  def from_client_report(%ClientReport{} = client_report) do
+    %__MODULE__{
+      event_id: UUID.uuid4_hex(),
+      items: [client_report]
+    }
+  end
+
+  @spec get_data_category(Attachment.t() | CheckIn.t() | ClientReport.t() | Event.t()) ::
+          String.t()
+  def get_data_category(type) do
```
whatyouhide, line 51: If you wanna raise, do it like this so that invalid input doesn't even get in the function:

```suggestion
  def get_data_category(%mod{} = type) when mod in [Attachment, CheckIn, ClientReport, Event] do
```

whatyouhide (commented):
There are no `Sentry.Transport` tests for the new functionality

test/envelope_test.exs
```diff
@@ -114,4 +114,38 @@ defmodule Sentry.EnvelopeTest do
       assert decoded_check_in["status"] == "ok"
     end
   end
+
+  test "works with client reports" do
+    put_test_config(environment_name: "test")
+
+    client_report = %ClientReport{
+      timestamp: "2024-10-12T13:21:13",
+      discarded_events: [%{reason: :event_processor, category: "error"}]
```
whatyouhide, line 123: Doesn't this need a quantity as well?

whatyouhide (commented):
We're ready for docs for this and to :shipit:

lib/sentry/client_report.ex
```diff
@@ -0,0 +1,149 @@
...
+        end)
+
+      client_report =
+        %__MODULE__{
+          timestamp: timestamp(),
+          discarded_events: discarded_events
+        }
+
+      IO.inspect(client_report)
+
```
whatyouhide, line 124:
```suggestion
```

whatyouhide (approved):

savhappy:
@whatyouhide should we create a ticket that reminds us to add in spans to the client report when supported?

PR #803, opened 2024-09-26
Add option for generating custom cron monitor slugs

iautom8things:
We've recently adopted the Oban integration with Sentry but have noticed that some of our jobs do not play nice with the assumptions in the library.  We have a pattern of reuse of cron workers where there's an argument that is the key differentiator.  As one example, this might be a `client_id`. 

We would like to be able to monitor these cron jobs separately.

I would be happy to add/improve more testing and/or documentation.  I really just wanted to get this opened as a starting point for discussion.

🙏 Thank you for this consideration.

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -60,10 +60,11 @@ defmodule Sentry.Config do
               """
             ],
             monitor_name_generator: [
-              type: {:fun, 1},
+              type: :mfa,
               doc: """
-              A function that generates a monitor name based on the Oban.Job struct. This can be used
-              to create monitors specific to a Job's argument. *Available since v10.7.1*.
+              A {module, function, arguments} tuple that generates a monitor name based on the Oban.Job struct.
+              This function should only accept a single argument, the Oban.Job struct, and return a string.
+              This can be used to create monitors specific to a Job's argument. *Available since v10.7.1*.
               """
             ]
           ]
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -58,6 +58,13 @@ defmodule Sentry.Config do
               Whether to enable the Oban integration. When enabled, the Sentry SDK will
               capture check-ins for Oban jobs. *Available since v10.2.0*.
               """
+            ],
+            monitor_name_generator: [
+              type: {:fun, 1},
+              doc: """
+              A function that generates a monitor name based on the Oban.Job struct. This can be used
+              create monitors that are specific to a Job's argument. *Available since v10.x.x*.
+              """
             ]
           ]
         ]
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -63,7 +63,7 @@ defmodule Sentry.Config do
               type: {:fun, 1},
               doc: """
               A function that generates a monitor name based on the Oban.Job struct. This can be used
-              create monitors that are specific to a Job's argument. *Available since v10.x.x*.
+              to create monitors specific to a Job's argument. *Available since v10.7.1*.
               """
             ]
           ]
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -59,12 +59,13 @@ defmodule Sentry.Config do
               capture check-ins for Oban jobs. *Available since v10.2.0*.
               """
             ],
-            monitor_name_generator: [
-              type: :mfa,
+            monitor_slug_generator: [
+              type: {:tuple, [:atom, :atom]},
+              type_doc: "`{module(), atom()}`",
               doc: """
-              A {module, function, arguments} tuple that generates a monitor name based on the Oban.Job struct.
-              This function should only accept a single argument, the Oban.Job struct, and return a string.
-              This can be used to create monitors specific to a Job's argument. *Available since v10.7.1*.
+              A `{module, function}` tuple that generates a monitor name based on the `Oban.Job` struct.
+              The function is called with the `Oban.Job` as its arguments and must return a string.
+              This can be used to customize monitor slugs. *Available since v10.8.0*.
               """
             ]
           ]
--- a/lib/sentry/integrations/oban/cron.ex
+++ b/lib/sentry/integrations/oban/cron.ex
@@ -72,16 +72,12 @@ defmodule Sentry.Integrations.Oban.Cron do
     monitor_config_opts = Sentry.Config.integrations()[:monitor_config_defaults]
 
     monitor_slug =
-      case config[:monitor_name_generator] do
+      case config[:monitor_slug_generator] do
         nil ->
           slugify(job.worker)
 
-        {mod, fun, args} when is_atom(mod) and is_atom(fun) and is_list(args) ->
-          if function_exported?(mod, fun, Enum.count(args) + 1) do
-            apply(mod, fun, [job | args]) |> slugify()
-          else
-            slugify(job.worker)
-          end
+        {mod, fun} when is_atom(mod) and is_atom(fun) ->
+          mod |> apply(fun, [job]) |> slugify()
       end
 
     case Keyword.merge(monitor_config_opts, schedule_opts(job)) do
--- a/lib/sentry/integrations/oban/cron.ex
+++ b/lib/sentry/integrations/oban/cron.ex
@@ -8,23 +8,23 @@ defmodule Sentry.Integrations.Oban.Cron do
     [:oban, :job, :exception]
   ]
 
-  @spec attach_telemetry_handler() :: :ok
-  def attach_telemetry_handler do
-    _ = :telemetry.attach_many(__MODULE__, @events, &__MODULE__.handle_event/4, :no_config)
+  @spec attach_telemetry_handler(keyword()) :: :ok
+  def attach_telemetry_handler(config \\ []) do
+    _ = :telemetry.attach_many(__MODULE__, @events, &__MODULE__.handle_event/4, config)
     :ok
   end
 
-  @spec handle_event([atom()], term(), term(), :no_config) :: :ok
-  def handle_event(event, measurements, metadata, _config)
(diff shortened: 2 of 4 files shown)


solnic:
Thanks for the PR. This sounds like a good addition to me. I approved the test run but unfortunately the tests didn't pass, any chance you could take a look and make them pass? 😄

iautom8things:
> Thanks for the PR. This sounds like a good addition to me. I approved the test run but unfortunately the tests didn't pass, any chance you could take a look and make them pass? 😄

That's odd.  In truth, I was testing my tests in isolation with `mix test --only custom_monitor_name_generator`.  For one, just to speed the testing up, but also it seems that the tests for the project are flaky, at least on `master`.

Here's a seed on my branch that my tests pass, but there's another test that fails:
```bash
 ✘ mz@mz-deca  ~/s/sentry-elixir   task/allow-custom-cron-monitor-name-generation ⁝ ✱ 
 [orbstack|default] ❯ mix test --seed 441419                                                   [10:00:16]
Wrote 37 files in 265.39 kb to: _build/test/lib/sentry/priv/sentry.map
...............................

  1) test with Plug sends two errors when a Plug process crashes if cowboy domain is not excluded (Sentry.LoggerHandlerTest)
     test/sentry/logger_handler_test.exs:99
     Assertion with == failed
     code:  assert second_event.original_exception == %RuntimeError{message: "Error"}
     left:  nil
     right: %RuntimeError{message: "Error"}
     stacktrace:
       test/sentry/logger_handler_test.exs:109: (test)

...............................................................................................................................................................................................................................................................
Finished in 2.1 seconds (0.6s async, 1.4s sync)
13 doctests, 274 tests, 1 failure

Randomized with seed 441419

```
---

On `master` the following test fails:

`mix test --seed 30420`

But the following test passes:

`mix test --seed 310035`

---

I just cleaned up the warnings about the unused variable. And I also believe I found the cause for my tests to fail, and fixed them.

But there still appears to be the same flaky test from above (`test/sentry/logger_handler_test.exs:99`) that fails about every 3rd attempt.

lib/sentry/config.ex
savhappy, line 1: @iautom8things can you also update this to say 

`A function that generates a monitor name based on the Oban.Job struct. This can be used to
              create monitors specific to a Job's argument. *Available since v10.7.1*. `

iautom8things: Sure thing! Done 🏁

whatyouhide:
I wont have time to review this this week as I’m on vacay, will review on Monday.

solnic:
> On `master` the following test fails:
> 
> `mix test --seed 30420`

I can't reproduce this. I also left running tests for like over an hour in a loop and they didn't fail :/

iautom8things:
> > On `master` the following test fails:
> > `mix test --seed 30420`
> 
> I can't reproduce this. I also left running tests for like over an hour in a loop and they didn't fail :/

This appears it may be related to the elixir/otp version?  It errors for me on 1.16.2 / 26.2.5 but if I upgrade to 1.17.2 / 27.0.1 it doesn't fail.

lib/sentry/config.ex
```diff
@@ -58,6 +58,13 @@ defmodule Sentry.Config do
               Whether to enable the Oban integration. When enabled, the Sentry SDK will
               capture check-ins for Oban jobs. *Available since v10.2.0*.
               """
+            ],
+            monitor_name_generator: [
+              type: {:fun, 1},
```
whatyouhide, line 63: I think it'd be safer to have the type of this be `{mod, fun}` because captures in config are not generally a good idea (config is used for compilation and yada yada).

lib/sentry/integrations/oban/cron.ex
```diff
@@ -8,23 +8,23 @@ defmodule Sentry.Integrations.Oban.Cron do
     [:oban, :job, :exception]
   ]
 
-  @spec attach_telemetry_handler() :: :ok
-  def attach_telemetry_handler do
-    _ = :telemetry.attach_many(__MODULE__, @events, &__MODULE__.handle_event/4, :no_config)
+  @spec attach_telemetry_handler(keyword()) :: :ok
+  def attach_telemetry_handler(config \\ []) do
```
whatyouhide, line 12: When is this ever called as just `attach_telemetry_handler()`? I don't think it is and if we add opts I don't think it _should_, so let's change this to


```suggestion
  def attach_telemetry_handler(config) when is_list(config) do
```

test/sentry/integrations/oban/cron_test.exs
```diff
@@ -4,111 +4,157 @@ defmodule Sentry.Integrations.Oban.CronTest do
 
   import Sentry.TestHelpers
 
-  setup_all do
-    Sentry.Integrations.Oban.Cron.attach_telemetry_handler()
-  end
-
-  setup do
-    bypass = Bypass.open()
+  describe "default configuration" do
```
whatyouhide, line 7: There seems to be only one additional test from a quick glance; if so, can we not wrap all of this in a `describe` to avoid the huge diff, and only add the new test?

whatyouhide (commented):
Nice job! Left some comments but this is looking like a great start 🙃

test/sentry/integrations/oban/cron_test.exs
iautom8things: So the tests were admittedly where I felt the least comfortable changing; so I'm not surprised you'd like changes.

The problem that I was trying to address is that I assumed we need a different `setup` block in order to show that all of the previous tests remained unchanged.  Before `setup_all` was being used to attach the telemetry handler a single time for all test cases.  The describe blocks were meant to separate the cases where the telemetry handler was attached with a non-default config.

I'd be happy to update this, but I'm not exactly sure how to best proceed.

Thanks :pray:

iautom8things: Well, _rubber duckie_ ... I wasn't aware that tags are a part of the context that is available in the setup function (https://hexdocs.pm/ex_unit/1.12/ExUnit.Case.html#module-tags)

I think I see a way forward.  I'll push something up for your review.

lib/sentry/config.ex
iautom8things: I just want to double check: 

Do you mean `:mfa` from here? https://hexdocs.pm/nimble_options/NimbleOptions.html#module-types

lib/sentry/config.ex
```diff
@@ -58,6 +58,14 @@ defmodule Sentry.Config do
               Whether to enable the Oban integration. When enabled, the Sentry SDK will
               capture check-ins for Oban jobs. *Available since v10.2.0*.
               """
+            ],
+            monitor_name_generator: [
+              type: :mfa,
```
whatyouhide, line 63: Since this is forced to accept only a single argument, make this:

```suggestion
              type: {:tuple, [:atom, :atom]},
              type_doc: "`{module(), atom()}`",
```

so that you can't accidentally call it in a wrong way.

lib/sentry/config.ex
```diff
@@ -58,6 +58,14 @@ defmodule Sentry.Config do
...
+              doc: """
+              A {module, function, arguments} tuple that generates a monitor name based on the Oban.Job struct.
+              This function should only accept a single argument, the Oban.Job struct, and return a string.
+              This can be used to create monitors specific to a Job's argument. *Available since v10.7.1*.
```
whatyouhide, line 67: With suggested change above:

```suggestion
              A `{module, function}` tuple that generates a monitor name based on the `Oban.Job` struct.
              The function is called with the `Oban.Job` as its arguments and must return a string.
              This can be used to customize monitor slugs. *Available since v10.8.0*.
```

lib/sentry/config.ex
```diff
@@ -58,6 +58,14 @@ defmodule Sentry.Config do
               Whether to enable the Oban integration. When enabled, the Sentry SDK will
               capture check-ins for Oban jobs. *Available since v10.2.0*.
               """
+            ],
+            monitor_name_generator: [
```
whatyouhide, line 62: Should this be called `monitor_slug_generator`?

lib/sentry/integrations/oban/cron.ex
```diff
@@ -60,17 +60,30 @@ defmodule Sentry.Integrations.Oban.Cron do
...
+      case config[:monitor_name_generator] do
+        nil ->
+          slugify(job.worker)
+
+        {mod, fun, args} when is_atom(mod) and is_atom(fun) and is_list(args) ->
+          if function_exported?(mod, fun, Enum.count(args) + 1) do
+            apply(mod, fun, [job | args]) |> slugify()
+          else
+            slugify(job.worker)
+          end
```
whatyouhide, line 84:
```suggestion
        {mod, fun} when is_atom(mod) and is_atom(fun) ->
          mod |> apply(fun, [job]) |> slugify()
```

test/sentry/integrations/oban/cron_test.exs
```diff
@@ -4,8 +4,16 @@ defmodule Sentry.Integrations.Oban.CronTest do
...
 
-  setup_all do
-    Sentry.Integrations.Oban.Cron.attach_telemetry_handler()
+  setup context do
+    opts =
+      if context[:custom_monitor_name_generator] do
+        [monitor_name_generator: {__MODULE__, :custom_name_generator, []}]
+      else
+        []
+      end
```
whatyouhide, line 13: You can do

```suggestion
    opts = context[:attach_opts]
```

and then below define the test as

```elixir
@tag attach_opts: [monitor_name_generator: {__MODULE__, :custom_name_generator}]
```

It should be a bit easier to follow and require less code 🙃

iautom8things:
Thank you, @whatyouhide, for all of the suggestions! :bow:

whatyouhide (approved):
@savhappy review and merge when you get the time plz :)

PR #807, opened 2024-10-09
Support non-exception error values from Oban jobs

### LINKED ISSUE #805: Oban ErrorReporter can crash and detach future telemetry events
### Environment

What version are you running? Etc.

* Sentry version: `10.7.1`
* Oban version: `2.18.3`

### Steps to Reproduce

Looking through server logs for disconnected telemetry handlers, I found the log below, which appears to be due to a `:timeout` error.

### Expected Result

I think [this code](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/integrations/oban/error_reporter.ex#L40) should probably account for non-Exceptions somehow.

### Actual Result

In this example, the `exception` is `:timeout` rather than being an `Exception` struct.

```
Handler Sentry.Integrations.Oban.ErrorReporter has failed and has been detached. Class=:error
Reason=:undef
Stacktrace=[
  {:timeout, :__struct__, [], []},
  {Sentry.Integrations.Oban.ErrorReporter, :handle_event, 4,
   [file: ~c"lib/sentry/integrations/oban/error_reporter.ex", line: 40]},
  {:telemetry, :"-execute/3-fun-0-", 4,
   [file: ~c"/tmp/project/deps/telemetry/src/telemetry.erl", line: 167]},
  {:lists, :foreach_1, 2, [file: ~c"lists.erl", line: 1442]},
  {Oban.Queue.Executor, :emit_event, 1,
   [file: ~c"lib/oban/queue/executor.ex", line: 284]},
  {Oban.Backoff, :with_retry, 3, [file: ~c"lib/oban/backoff.ex", line: 108]},
  {Task.Supervised, :invoke_mfa, 2,
   [file: ~c"lib/task/supervised.ex", line: 101]},
  {Task.Supervised, :reply, 4, [file: ~c"lib/task/supervised.ex", line: 36]}
]
```

savhappy:
-added `capture_message` to `handle_event` in `Sentry.Integrations.Oban.ErrorReporter` to account for non-exception args that are passed

-closes #805

### DIFF
--- a/lib/sentry/integrations/oban/error_reporter.ex
+++ b/lib/sentry/integrations/oban/error_reporter.ex
@@ -33,18 +33,22 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
       end
 
     _ =
-      Sentry.capture_exception(exception,
-        stacktrace: stacktrace,
-        tags: %{oban_worker: job.worker, oban_queue: job.queue, oban_state: job.state},
-        fingerprint: [
-          inspect(exception.__struct__),
-          inspect(job.worker),
-          Exception.message(exception)
-        ],
-        extra:
-          Map.take(job, [:args, :attempt, :id, :max_attempts, :meta, :queue, :tags, :worker]),
-        integration_meta: %{oban: %{job: job}}
-      )
+      if is_struct(exception) do
+        Sentry.capture_exception(exception,
+          stacktrace: stacktrace,
+          tags: %{oban_worker: job.worker, oban_queue: job.queue, oban_state: job.state},
+          fingerprint: [
+            inspect(exception.__struct__),
+            inspect(job.worker),
+            Exception.message(exception)
+          ],
+          extra:
+            Map.take(job, [:args, :attempt, :id, :max_attempts, :meta, :queue, :tags, :worker]),
+          integration_meta: %{oban: %{job: job}}
+        )
+      else
+        Sentry.capture_message("Error with %s", interpolation_parameters: [exception])
+      end
 
     :ok
   end
--- a/lib/sentry/integrations/oban/error_reporter.ex
+++ b/lib/sentry/integrations/oban/error_reporter.ex
@@ -49,17 +49,18 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
         integration_meta: %{oban: %{job: job}}
       ]
 
-    if is_exception(reason) do
-      Sentry.capture_exception(
-        reason,
-        opts
-      )
-    else
-      Sentry.capture_message(
-        "Oban job #{job.worker} errored out: %s",
-        opts ++ [interpolation_parameters: [inspect(reason)]]
-      )
-    end
+    _ =
+      if is_exception(reason) do
+        Sentry.capture_exception(
+          reason,
+          opts
+        )
+      else
+        Sentry.capture_message(
+          "Oban job #{job.worker} errored out: %s",
+          opts ++ [interpolation_parameters: [inspect(reason)]]
+        )
+      end
 
     :ok
   end
--- a/lib/sentry/integrations/oban/error_reporter.ex
+++ b/lib/sentry/integrations/oban/error_reporter.ex
@@ -33,17 +33,17 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
       end
 
     fingerprint_opts =
-      (is_exception(reason) && [inspect(reason.__struct__), Exception.message(reason)]) ||
+      if is_exception(reason) do
+        [inspect(reason.__struct__), Exception.message(reason)]
+      else
         [inspect(reason)]
+      end
 
     opts =
       [
         stacktrace: stacktrace,
         tags: %{oban_worker: job.worker, oban_queue: job.queue, oban_state: job.state},
-        fingerprint:
-          [
-            inspect(job.worker)
-          ] ++ fingerprint_opts,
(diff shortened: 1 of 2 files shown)


lib/sentry/integrations/oban/error_reporter.ex
```diff
@@ -33,18 +33,22 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
...
-        fingerprint: [
-          inspect(exception.__struct__),
-          inspect(job.worker),
-          Exception.message(exception)
-        ],
-        extra:
-          Map.take(job, [:args, :attempt, :id, :max_attempts, :meta, :queue, :tags, :worker]),
-        integration_meta: %{oban: %{job: job}}
-      )
+      if is_struct(exception) do
```
GregMefford, line 36: You may want more specifically [`is_exception`](https://hexdocs.pm/elixir/1.12.3/Kernel.html#is_exception/1) here, rather than just `is_struct`, because I think `Exception.message/1` below won't work on just any general struct.

This may not really matter in practice, though, because I think the errors will probably never end up being a struct other than an Exception.

savhappy: Probably not but good to use the right tool for the job. Thanks and updated!

lib/sentry/integrations/oban/error_reporter.ex
```diff
@@ -33,18 +33,22 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
...
+            inspect(job.worker),
+            Exception.message(exception)
+          ],
+          extra:
+            Map.take(job, [:args, :attempt, :id, :max_attempts, :meta, :queue, :tags, :worker]),
+          integration_meta: %{oban: %{job: job}}
+        )
+      else
+        Sentry.capture_message("Error with %s", interpolation_parameters: [exception])
+      end
```
solnic, line 51: It would be more clear to have `reason` instead of `exception` here since it may not be an exception after all :)

savhappy: True! Updated, thanks!

whatyouhide: This is a wildly generic error message we got going on here 😄 I think it'd be really hard to debug as it's just a generic message plus there are no metadata about the job (which you have available).

This needs to be completely restructured to basically build shared options, and then use `Sentry.capture_message` or `Sentry.capture_exception` based on whether `reason` is an exception or not. Also, doesn't `:interpolation_parameters` require string parameters? If so, do `inspect(reason)` here.

whatyouhide: Also let's have the message be something like

```elixir
"Oban job #{job.worker} errored out: %s"
```

Btw the fingerprint option needs to be customized, if it's just reason use `inspect(reason)` insted of `Exception.message/1`

whatyouhide (changes requested):
Nice job, but we need a slightly different approach 🙃

savhappy:
@whatyouhide updated with your feedback! Yeah, I assumed that error message was not up to code 😆 . This restructure feels better

lib/sentry/integrations/oban/error_reporter.ex
```diff
@@ -24,27 +24,43 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
...
       case {apply(Oban.Worker, :from_string, [job.worker]), stacktrace} do
         {{:ok, atom_worker}, []} -> [{atom_worker, :process, 1, []}]
         _ -> stacktrace
       end
 
-    _ =
-      Sentry.capture_exception(exception,
+    fingerprint_opts =
+      (is_exception(reason) && [inspect(reason.__struct__), Exception.message(reason)]) ||
+        [inspect(reason)]
```
whatyouhide, line 37: Please use `if`, this is pretty hard to read and tries to mimic the `? :` operator thing that Elixir doesn't have 🙃

lib/sentry/integrations/oban/error_reporter.ex
```diff
@@ -24,27 +24,43 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
...
         tags: %{oban_worker: job.worker, oban_queue: job.queue, oban_state: job.state},
-        fingerprint: [
-          inspect(exception.__struct__),
-          inspect(job.worker),
-          Exception.message(exception)
-        ],
+        fingerprint:
+          [
+            inspect(job.worker)
+          ] ++ fingerprint_opts,
```
whatyouhide, line 46: Nit but saves some space 😉 

```suggestion
        fingerprint: [inspect(job.worker)] ++ fingerprint_opts,
```

lib/sentry/integrations/oban/error_reporter.ex
```diff
@@ -24,27 +24,43 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
...
         integration_meta: %{oban: %{job: job}}
-      )
+      ]
+
+    _ =
+      if is_exception(reason) do
+        Sentry.capture_exception(
+          reason,
+          opts
+        )
```
whatyouhide, line 57: Saves some space.

```suggestion
        Sentry.capture_exception(reason, opts)
```

whatyouhide (commented):
Last comment and we can ship this.

@solnic yeah that's probably something that could be *configurable* I think.

whatyouhide (approved):

PR #808, opened 2024-10-14
Add 10 min timeout to CI

solnic:
I noticed that one of the test runs got stuck and ran for over half an hour so I figured a timeout would be helpful.


whatyouhide (approved):
Fantastic yep!

PR #814, opened 2024-10-22
Handle bandit's get peer data exception

ardhitama:
As of bandit v1.5.7 during client disconnection it will raise "Unable to obtain transport_info: [reason]" every time it tries to get the transport info. This commit tries to handle that exception by returning nil instead of bubbling the exception.

Context: https://github.com/mtrudel/bandit/blob/b06e43dcee93c5c044b18ed8347838fc2a1728f0/lib/bandit/transport_info.ex#L26

### DIFF
--- a/lib/sentry/plug_context.ex
+++ b/lib/sentry/plug_context.ex
@@ -211,8 +211,18 @@ defmodule Sentry.PlugContext do
 
   defp remote_port(conn) do
     case Plug.Conn.get_req_header(conn, "x-forwarded-for") do
-      [] -> nil
-      [_value | _rest] -> Plug.Conn.get_peer_data(conn).port
+      [] ->
+        nil
+
+      [_value | _rest] ->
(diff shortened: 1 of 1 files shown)


lib/sentry/plug_context.ex
```diff
@@ -211,8 +211,15 @@ defmodule Sentry.PlugContext do
 
   defp remote_port(conn) do
     case Plug.Conn.get_req_header(conn, "x-forwarded-for") do
-      [] -> nil
-      [_value | _rest] -> Plug.Conn.get_peer_data(conn).port
+      [] ->
+        nil
+
+      [_value | _rest] ->
+        try do
```
whatyouhide, line 218: Can we leave a comment here saying that Bandit raises in some cases here and maybe point to the source code like you did in the PR description?

ardhitama: thanks for the comment @whatyouhide , I've updated the commit.

whatyouhide (approved):

PR #815, opened 2024-10-25
Use config_env() instead of Mix.env()

dvic:


whatyouhide (approved):
Great catch yep! Thank you. 💟

PR #817, opened 2024-10-28
Refactor span storage

savhappy:
[ WIP] Needs test adjustment and specs
  Considerations:
  Should we store root spans and child spans in separate tables?
  Storing key/values like this -> parent: `{{:root_span, span_id}, span}`  child: `{parent_span_id, span}`
  Should we implement some sort of sweep in case spans don't get removed `on_end`
  Should we order the child spans in the table? Or have an inserted_at value?

### DIFF
--- a/lib/sentry/opentelemetry/span_storage.ex
+++ b/lib/sentry/opentelemetry/span_storage.ex
@@ -1,78 +1,82 @@
 defmodule Sentry.Opentelemetry.SpanStorage do
   @moduledoc false
 
-  @root_spans_table :sentry_root_spans
-  @child_spans_table :sentry_child_spans
+  @table :span_storage
 
-  def setup do
-    case :ets.whereis(@root_spans_table) do
-      :undefined ->
-        :ets.new(@root_spans_table, [:set, :public, :named_table])
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link(opts \\ []) do
+    name = Keyword.get(opts, :name, __MODULE__)
+    GenServer.start_link(__MODULE__, nil, name: name)
+  end
 
-      _ ->
-        :ok
-    end
+  # start ets table on initialization
+  @impl true
+  def init(nil) do
+    _table =
+      if :ets.whereis(@table) == :undefined do
+        :ets.new(@table, [:named_table, :public, :bag])
+      end
 
-    case :ets.whereis(@child_spans_table) do
-      :undefined ->
-        :ets.new(@child_spans_table, [:bag, :public, :named_table])
+    {:ok, :no_state}
+  end
 
-      _ ->
-        :ok
-    end
-
-    :ok
+  def store_span(span_data) when span_data.parent_span_id == nil do
+    _ = :ets.insert(@table, {{:root_span, span_data.span_id}, span_data})
   end
 
   def store_span(span_data) do
-    if span_data.parent_span_id == nil do
-      :ets.insert(@root_spans_table, {span_data.span_id, span_data})
-    else
-      :ets.insert(@child_spans_table, {span_data.parent_span_id, span_data})
-    end
-
-    :ok
+    _ = :ets.insert(@table, {span_data.parent_span_id, span_data})
   end
 
   def get_root_span(span_id) do
-    case :ets.lookup(@root_spans_table, span_id) do
-      [{^span_id, span}] -> span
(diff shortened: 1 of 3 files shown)


savhappy:
@whatyouhide refactored with the genserver added. Left a comment wondering if we should sweep based on ttl or something else to avoid build up.

@solnic I have two failing test that I'm confused about.. 

``` elixir
  1) test sends captured spans as transactions with child spans (Sentry.Opentelemetry.SpanProcessorTest)
     test/sentry/opentelemetry/span_processor_test.exs:59
     ** (FunctionClauseError) no function clause matching in Calendar.ISO.parse_utc_datetime/2

     The following arguments were given to Calendar.ISO.parse_utc_datetime/2:

         # 1
         nil

         # 2
         :extended

     Attempted function clauses (showing 1 out of 1):

         def parse_utc_datetime(string, format) when is_binary(string) and (format === :basic or format === :extended)

     code: assert_valid_iso8601(transaction.timestamp)
     stacktrace:
       (elixir 1.17.2) lib/calendar/iso.ex:591: Calendar.ISO.parse_utc_datetime/2
       (elixir 1.17.2) lib/calendar/datetime.ex:1233: DateTime.from_iso8601/3
       test/sentry/opentelemetry/span_processor_test.exs:112: Sentry.Opentelemetry.SpanProcessorTest.assert_valid_iso8601/1
       test/sentry/opentelemetry/span_processor_test.exs:67: (test)
       ```
       
Thoughts?

File: lib/sentry/opentelemetry/span_storage.ex [hunk 1]
```diff
@@ -1,78 +1,81 @@
...
+    GenServer.start_link(__MODULE__, nil, name: name)
+  end
 
-      _ ->
-        :ok
-    end
+  @impl true
+  def init(nil) do
+    _table =
+      if :ets.whereis(@table) == :undefined do
```
whatyouhide, line 16: If the table is tied to this GenServer, where would we ever have the table running *without* it that we need this check?

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -1,78 +1,81 @@
 defmodule Sentry.Opentelemetry.SpanStorage do
   @moduledoc false
+  use GenServer
 
-  @root_spans_table :sentry_root_spans
-  @child_spans_table :sentry_child_spans
+  @table :span_storage
```
whatyouhide, line 5: Scope this to Sentry, as the ETS namespace is global. My suggestion is

```suggestion
  @table __MODULE__
```

savhappy: In one of our pairing sessions for `check_in_id_mappings.ex`, we discussed this being bad practice to have both the GenServer and the table name the same. Thoughts?

solnic:
> @solnic I have two failing test that I'm confused about..
> 
> ```elixir
>   1) test sends captured spans as transactions with child spans (Sentry.Opentelemetry.SpanProcessorTest)
>      test/sentry/opentelemetry/span_processor_test.exs:59
>      ** (FunctionClauseError) no function clause matching in Calendar.ISO.parse_utc_datetime/2
> 
>      The following arguments were given to Calendar.ISO.parse_utc_datetime/2:
> 
>          # 1
>          nil
> 
>          # 2
>          :extended
> 
>      Attempted function clauses (showing 1 out of 1):
> 
>          def parse_utc_datetime(string, format) when is_binary(string) and (format === :basic or format === :extended)
> 
>      code: assert_valid_iso8601(transaction.timestamp)
>      stacktrace:
>        (elixir 1.17.2) lib/calendar/iso.ex:591: Calendar.ISO.parse_utc_datetime/2
>        (elixir 1.17.2) lib/calendar/datetime.ex:1233: DateTime.from_iso8601/3
>        test/sentry/opentelemetry/span_processor_test.exs:112: Sentry.Opentelemetry.SpanProcessorTest.assert_valid_iso8601/1
>        test/sentry/opentelemetry/span_processor_test.exs:67: (test)
>      
> ```
> 
> Thoughts?

I'm looking into this finally. Sorry for late reply!

solnic:
@savhappy fixed tests via https://github.com/getsentry/sentry-elixir/pull/817/commits/20534087c012d03ba9bec708221f2af722b35db4

@whatyouhide could you take another look please? would be good to merge this into the main PR branch and continue working there and wrap this up :)

I'm sorry that it took me so long to follow-up here!

PR #826, opened 2024-12-05
Update to LiveView 1.0 🎉

### LINKED ISSUE #825: LiveView has reached 1.0.0
LiveView has reached 1.0.0 🎉 

This project requires ~> 0.20 which is incompatible. Upgrade the LiveView dep?

solnic:
Closes #825


solnic:
@whatyouhide what's the reasoning behind having mix.lock files in git?

mix.exs
```diff
@@ -97,7 +105,7 @@ defmodule Sentry.Mixfile do
       {:hackney, "~> 1.8", optional: true},
       {:jason, "~> 1.1", optional: true},
       {:phoenix, "~> 1.6", optional: true},
-      {:phoenix_live_view, "~> 0.20", optional: true},
+      {:phoenix_live_view, live_view_version(), optional: true},
```
whatyouhide, line 108: I think instead of basing on the Elixir version, we should do `~> 0.20 or ~> 1.0` here. Thoughts?

test_integrations/phoenix_app/mix.exs
```diff
@@ -35,12 +35,11 @@ defmodule PhoenixApp.MixProject do
       {:nimble_options, "~> 1.0"},
       {:nimble_ownership, "~> 0.3.0 or ~> 1.0"},
 
-      {:phoenix, "~> 1.7.14"},
       {:postgrex, ">= 0.0.0"},
+      {:phoenix, "~> 1.7.14"},
       {:phoenix_html, "~> 4.1"},
+      {:phoenix_live_view, "~> 1.0", override: true},
```
whatyouhide, line 41: You don't need the override anymore:

```suggestion
      {:phoenix_live_view, "~> 1.0"},
```

whatyouhide (commented):
In general is so that you get reproducible `mix deps.get` every time instead of making it depend on what versions are up on Hex. For libraries like this one, dependants don't read `mix.lock`, it's only for local use when developing the library. Does it make sense?

mix.exs
solnic: @whatyouhide despite the fact that LV 1.0.0 does not work with older Elixir, it did happily install 1.0.0 on Elixir 1.13 - so I thought it has to be explicitly restricted by the version requirement.

solnic:
> In general is so that you get reproducible mix deps.get every time instead of making it depend on what versions are up on Hex. For libraries like this one, dependants don't read mix.lock, it's only for local use when developing the library. Does it make sense?

@whatyouhide yeah it does but the downside is that we're locking ourselves to specific versions, and then people release new versions of stuff we depend on and we're missing running our test suite against these updated versions. My approach for years was to always let the CI install whatever is the latest. This also helps catch installation issues early on. There are of course special cases like explicitly testing against a very specific set of versions for some crucial deps. In our case that would be of course Phoenix and LiveView.

I'm not saying we should remove the locks though, just raising this as a concern. We could solve this in various ways - like having dependabot configured for our needs, or a separate CI run that does mix deps.update --all.

test_integrations/phoenix_app/mix.exs
solnic: @whatyouhide seems like this broke CI :/

whatyouhide:
Yeah we shouldn't remove the locks. Dependabot is probably what we want here.

test_integrations/phoenix_app/mix.exs
whatyouhide: Seems like a caching issue, purge the cache with the `gh` CLI and you should be good to go.

marpo60:
👋 👋 👋 

Not sure if it was on your radar but we built [blend](https://github.com/mimiquate/blend) to solve this issue. You can test an hex package against different version of its dependencies.

[This is an example](https://github.com/surface-ui/surface/pull/744) of surface-ui testing against phoenix_live_view 0.20.x and 0.19.x

phinnaeus:
@solnic it's best practice not just with elixir but in pretty much every language that supports it to commit your lockfiles. that's the whole point of the lockfile concept, for reproducibility. note this is not the same as vendoring your dependencies.

mix.exs
whatyouhide: Ok, but we still can't bump the version to `~> 1.0` for Sentry (when Elixir >= 1.14) because people might be depending on older LV versions and Sentry doens't need the latest.

whatyouhide (commented):
Can you also rebase on master please? 🙃

mix.exs
phinnaeus: Why not just change it to `:phoenix_live_view, ">= 0.20", optional: true`

whatyouhide: `>= 20.0` will have the same issue as `~> 0.20 or ~> 1.0`

phinnaeus: Maybe I'm being dumb, but wouldn't `>= 0.20` open up for consumers to use LiveView 1.0 where the current `~> 0.20` is equivalent to `>= 0.20 and < 0.21`.

To be clear I'm currently blocked on upgrading to LiveView 1.0 because of this:

```
% mix deps.get
Resolving Hex dependencies...
Resolution completed in 0.167s
Because sentry >= 10.5.0 depends on phoenix_live_view ~> 0.20 and your app depends on phoenix_live_view ~> 1.0, sentry >= 10.5.0 is forbidden.
So, because your app depends on sentry ~> 10.7, version solving failed.
** (Mix) Hex dependency resolution failed
```

solnic:
@whatyouhide I'm not sure how to make this work with 1.13 and locks checked in. Any ideas? Do we need a separate lock for 1.13? When installing deps on CI we can't have locks that point to LV 1.0 because that will make 1.13 install it and crash. Having "~> 0.20 or ~> 1.0" doesn't make 1.13 install older LV (unsurprisingly).

...unless I'm missing something obvious?

whatyouhide:
For now patch CI so that you run `mix deps.unlock phoenix_live_view` if the Elixir version is older. This is temporary anyway since it'll go away once we drop support for older Elixir versions.

solnic:
> @solnic it's best practice not just with elixir but in pretty much every language that supports it to commit your lockfiles. that's the whole point of the lockfile concept, for reproducibility. note this is not the same as vendoring your dependencies.

In my experience, it is FAR simpler to not have them checked in for OSS projects when you care about staying on top of the deps (as in, test against latest versions). Locks checked in make perfect sense for apps, but not for OSS libs. The whole point of running CI for OSS libs is to ensure that things don't break and your lib can be installed, it is so easy to get behind by an accident. I understand that you can have locks checked in and use other tools to inform you about outdated deps, but from a pragmatic/time-saver POV it is just extra work. I've found it to be just simpler and less time consuming to not have locks in OSS lib repos.

This is my personal opinion and experience - our situation here is different :)

whatyouhide:
Peter, I have never experienced issues with checked-in lockfiles or had my time consumed 🙃 We have a path forward here, so let's just take that and merge this PR 💯

phinnaeus:
@whatyouhide I'm not trying to argue or hold anything up, just for my own curiosity -- what's the problem with using `>= 20.0`? It seems like it would allow consumers to use either `1.0` or `0.20` unlike `~> 0.20` is equivalent to `>= 0.20 and < 0.21`.

Munksgaard:
@phinnaeus: I'm pretty sure [Mix supports `or` clauses in the version requirements](https://hexdocs.pm/elixir/Version.html#module-requirements). Perhaps `"~> 0.20 or ~> 1.0"` is more reasonable?

solnic:
With "or" it still installs the latest version because that's what we have in the lock, unless I'm doing something wrong here :(

solnic:
> @whatyouhide I'm not trying to argue or hold anything up, just for my own curiosity -- what's the problem with using >= 20.0?

Open-ended dep requirements are problematic because 2.0.0 may no longer be compatible, we could have ">= 1.0 and < 2.0" though.

solnic:
@whatyouhide @phinnaeus looks like I've made it 🙃 the trick was to skip doing deps.get in test/integrations since we don't even run these tests under 1.13, and also skip doing --check-locked under 1.13

phinnaeus (approved):

whatyouhide (approved):

PR #829, opened 2024-12-06
Add Owner Field to Monitor Config

### LINKED ISSUE #824: Oban Cron: provide a way to set owner of monitor
Hi there and thanks for the great library!

I'm wondering wether it would be possible to provide a way to set the `owner` of a cron monitor via the automatic integration?

We have a setup where we know the owning Sentry team in our codebase and if we could somehow tell the integration to set it correctly it would ease our debugging flow a lot.

e.g. by specifying a callback or some kind of mapping. You probably know best how to achieve it! 

Thanks ❤

savhappy:
Small addition that adds the `:owner` field to a check-in's monitor_config. Closes #824

### DIFF
--- a/lib/sentry/check_in.ex
+++ b/lib/sentry/check_in.ex
@@ -58,7 +58,8 @@ defmodule Sentry.CheckIn do
                 optional(:max_runtime) => number(),
                 optional(:failure_issue_threshold) => number(),
                 optional(:recovery_threshold) => number(),
-                optional(:timezone) => String.t()
+                optional(:timezone) => String.t(),
+                optional(:owner) => String.t()
               },
           contexts: Interfaces.context()
         }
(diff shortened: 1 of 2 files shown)


solnic (approved):

lib/sentry/check_in.ex
```diff
@@ -58,7 +58,8 @@ defmodule Sentry.CheckIn do
                 optional(:max_runtime) => number(),
                 optional(:failure_issue_threshold) => number(),
                 optional(:recovery_threshold) => number(),
-                optional(:timezone) => String.t()
+                optional(:timezone) => String.t(),
+                optional(:owner) => String.t()
```
whatyouhide, line 62: **Blocking**: we need to mention that this is a new field available from the next version.

whatyouhide (changes requested):
I’m not sure this will be very useful without a way to set this per job on the Oban and Quantum integrations. Any ideas for the API?

Miradorn:
> @savhappy any chance we can get this back into shape? 🙃

i think this was solved already in https://github.com/getsentry/sentry-elixir/pull/879 , right?

PR #832, opened 2024-12-11
Add support for :skip_retries config option to Oban.ErrorReporter

solnic:
This is inspired by this thread on Bsky started by @katafrakt:

https://bsky.app/profile/katafrakt.bsky.social/post/3lcuccw6zik2x

Long story short - there are cases where reporting every job failure as an error is not desired. When you expect that some jobs just fail sometimes due to some external factors, you want to be notified only if all attempts failed.

To solve this, we can introduce a config option to instruct Oban.ErrorReporter what to do when a job failed but it was not the final attempt. This PR proposes such an addition.


whatyouhide:
@solnic what's the status on this? Something we want to pursue?

brendon9x:
Hi all – this would be a very useful feature. A common case is AI workloads: I am finding Anthropic fails about 1/250 times so I'm getting Sentries every hour or so. But these are just transient so all Oban jobs are succeeding on the 2nd try.

@solnic – this looks quite close. I would be happy to pick this up if the effort of handing over your vision for this PR isn't going to be more complicated than just doing it yourself.

lib/sentry/config.ex
```diff
@@ -78,6 +78,25 @@ defmodule Sentry.Config do
           with `oban_tags.` and with a value of `true`. *Available since 12.0.0*.
           """
         ],
+        skip_error_report_callback: [
+          type: {:custom, __MODULE__, :__validate_skip_error_report_callback__, []},
```
whatyouhide: Don't think we need custom validation?

```suggestion
          type: {:or, [nil, {:fun, 2}]},
```

lib/sentry/integrations/oban/error_reporter.ex
```diff
@@ -31,13 +31,50 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
...
       report(job, kind, reason, stacktrace, config)
     else
       :ok
     end
   end
 
+  defp should_report?(job, config) do
+    case Keyword.get(config, :skip_error_report_callback) do
+      callback when is_function(callback, 2) ->
+        not call_skip_error_report_callback(callback, job)
```
whatyouhide: This `not` right here makes me think that it might be easier to rephrase the option name as `:should_report_error`? Might make it clearer for the end user?

sl0thentr0py (approved):

Issue #830, opened 2024-12-10
Add `timezone` support for Oban cron monitoring


Flo0807:
Sentry supports monitoring scheduled Oban jobs.  I wanted to use this feature in my application to monitor Oban cron jobs, but Sentry always threw `A missed check-in was detected` errors. I looked into the problem and noticed in the web interface that the configured timezone for all cron monitors was set to `UTC` while I configured the Oban cron jobs to use the `Europe/Berlin` timezone. Therefore, I needed to change the timezone of the created cron monitors to `Europe/Berlin` manually to be in sync with my jobs. 

I figured out that you can set the timezone with a [Manual Setup](https://docs.sentry.io/platforms/elixir/crons/#manual-setup), but I'd like to have a `timezone` setting for my config that allows me to set the timezone for the oban cron integrations. I would rather not do a manual setup only to set the `timezone` value.

This is my setup:

```elixir
# config/config.exs
config :sentry,
  ...,
  integrations: [
    oban: [
      cron: [enabled: true],
      capture_errors: true
    ]
  ]
```

```elixir
# config/config.exs
config :my_app, Oban,
  repo: MyApp.Repo,
  queues: [default: 10],
  plugins: [
    {Oban.Plugins.Pruner, max_age: :timer.hours(24) * 90},
    {Oban.Plugins.Lifeline, rescue_after: :timer.minutes(5)},
    {Oban.Plugins.Cron,
     crontab: [
       {"0 8 * * 1", MyApp.Workers.ExampleWorker}
     ],
     timezone: "Europe/Berlin"}
  ]
```

whatyouhide:
@savhappy @solnic I think it's time we consider a per-job configuration when necessary.

One option is to do it via a behaviour that jobs (not just Oban jobs) can implement. The behaviour can be `Sentry.CheckIn` too.

```elixir
defmodule Sentry.CheckIn do
  @callback sentry_check_in_configuration() :: (options_to_merge :: keyword())

  # ...
end
```

Then, an Oban worker could implement it like this:

```elixir
defmodule MyApp.Worker do
  use Oban.Worker
 
  @behaviour Sentry.CheckIn

  @impl Oban.Worker
  def process(...)

  @impl Sentry.CheckIn
  def sentry_check_in_configuration do
    [timezone: "Europe/Berlin"]
  end
end
```

Then, integrations like the Oban one can essentially do:

```elixir
config =
[...]

PR #833, opened 2024-12-11
Support per-module custom options for check ins

savhappy:
Following the conversations from issues #830 & #824, the decision to add a public API to be able to set custom config options to the monitor is in-progress.


whatyouhide:
Taking this over.

whatyouhide (approved):

PR #839, opened 2024-12-17
Fix duplicate @compile attribute

ruslandoga:
Or should it maybe be removed completely?

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -665,7 +665,7 @@ defmodule Sentry.Config do
       """
   end
 
-  @compile {:inline, fetch!: 1}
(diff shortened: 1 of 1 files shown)


lib/sentry/config.ex
```diff
@@ -665,7 +665,7 @@ defmodule Sentry.Config do
       """
   end
 
-  @compile {:inline, fetch!: 1}
+  @compile {:inline, get: 1}
```
ruslandoga, line 668: `@compile {:inline, fetch!: 1}` is already defined above, on line 656

whatyouhide (approved):

whatyouhide:
Thank you! 💟

PR #840, opened 2024-12-17
feat: add `tags_from_metadata` option to LoggerHandler

### LINKED ISSUE #827: Add an ability to set tags from logger metadata
I believe `Sentry.LoggerHandler` would benefit from the option to set Sentry tags based on the log metadata as opposed to using it only as extra context. 

We have a multi-tenant application, where there is a supervision sub-tree per tenant. Each process already correctly tracks their belonging to a tenant using `Logger.metadata` calls, and this tenant ID would be extremely useful to have as a Sentry tag for filtering and grouping. I understand the option of using [`set_tags_context/1`](https://hexdocs.pm/sentry/Sentry.Context.html#set_tags_context/1), but I would rather avoid code duplication given we already set that metadata field, and `LoggerHandler` already reads those metadata keys correctly. 

I think allowing something like `tags: metadata_keys()` option would be enough to cover this use-case.

Thanks a lot!

icehaunter:
Added a new configuration option 'tags_from_metadata' to both `LoggerHandler` that allows specifying which metadata fields should be included as tags in Sentry events. 

Closes #827

### DIFF
--- a/lib/sentry/logger_backend.ex
+++ b/lib/sentry/logger_backend.ex
@@ -150,7 +150,14 @@ defmodule Sentry.LoggerBackend do
     # The context in the Logger backend process is not the same as the one in the process
     # that did the logging. This behavior is different than the one in Sentry.LoggerHandler,
     # since Logger handlers run in the caller process.
-    opts = LoggerUtils.build_sentry_options(level, sentry_context, Map.new(meta), state.metadata)
+    opts =
+      LoggerUtils.build_sentry_options(
+        level,
+        sentry_context,
+        Map.new(meta),
+        state.metadata,
+        _tags_from_metadata = []
+      )
 
     case meta[:crash_reason] do
       # If the crash reason is an exception, we want to report the exception itself
--- a/lib/sentry/logger_handler.ex
+++ b/lib/sentry/logger_handler.ex
@@ -44,6 +44,14 @@ defmodule Sentry.LoggerHandler do
       If set to `:all`, all metadata will be included.
       """
     ],
+    tags_from_metadata: [
+      type: {:list, :atom},
+      default: [],
+      doc: """
+      Use this to include logger metadata as tags in reports. Metadata under the specified keys
+      in those keys will be added as tags to the event.
+      """
+    ],
     capture_log_messages: [
       type: :boolean,
       default: false,
@@ -220,6 +228,7 @@ defmodule Sentry.LoggerHandler do
     :level,
     :excluded_domains,
     :metadata,
+    :tags_from_metadata,
     :capture_log_messages,
     :rate_limiting,
     :sync_threshold
@@ -321,7 +330,8 @@ defmodule Sentry.LoggerHandler do
             log_level,
             log_meta[:sentry],
             log_meta,
-            config.metadata
+            config.metadata,
+            config.tags_from_metadata
           )
 
         log_unfiltered(log_event, sentry_opts, config)
--- a/lib/sentry/logger_handler.ex
+++ b/lib/sentry/logger_handler.ex
@@ -49,7 +49,7 @@ defmodule Sentry.LoggerHandler do
       default: [],
       doc: """
       Use this to include logger metadata as tags in reports. Metadata under the specified keys
-      in those keys will be added as tags to the event.
+      in those keys will be added as tags to the event. *Available since v10.9.0*.
       """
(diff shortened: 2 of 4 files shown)


ruslandoga:
Sorry for a drive-by comment, but what happens if metadata values are maps or lists? Wouldn't tags need to be scalar?

lib/sentry/logger_backend.ex
```diff
@@ -141,7 +148,10 @@ defmodule Sentry.LoggerBackend do
 
     sentry_context =
       if sentry_context_from_meta || sentry_context_from_sentry do
-        Map.merge(sentry_context_from_meta || %{}, sentry_context_from_sentry || %{})
+        Map.merge(
+          Map.new(sentry_context_from_meta || %{}),
+          Map.new(sentry_context_from_sentry || %{})
+        )
```
whatyouhide, line 154: Unnecessary change?

lib/sentry/logger_backend.ex
```diff
@@ -42,6 +42,10 @@ defmodule Sentry.LoggerBackend do
   to `[]`. If set to `:all`, all metadata will be included. `:all` is available
   since v9.0.0 of this library.
 
+  * `:tags` - To include Logger metadata as tags in reports, the `:tags` key can be set to a list of keys.
```
whatyouhide, line 45: Should we call this `:tags_from_metadata`?

lib/sentry/logger_handler.ex
```diff
@@ -44,6 +44,15 @@ defmodule Sentry.LoggerHandler do
       If set to `:all`, all metadata will be included.
       """
     ],
+    tags: [
+      type: {:or, [{:list, :atom}, {:in, [:all]}]},
+      default: [],
+      type_doc: "list of `t:atom/0`, or `:all`",
+      doc: """
+      Use this to include logger metadata as tags in reports. If it's a list of keys, metadata
+      in those keys will be added as tags to the event. If set to `:all`, all metadata will be included as tags.
```
whatyouhide, line 53: I don't think we can include **all** metadata as tags because some metadata would be really worthless as tags (`:line` as an example). We should discourage that behavior from users I think. Thoughts?

lib/sentry/logger_backend.ex
whatyouhide: Also, sorry just seeing now that this is `Sentry.LoggerBackend`: let's not add this option there and make it only available for the Logger handler. We will eventually deprecate and then remove the Logger backend, so better to not add new features to it.

whatyouhide (changes requested):
Some changes but looking great.

lib/sentry/logger_handler.ex
icehaunter: Yeah, I agree. I'll remove the `:all`

lib/sentry/logger_backend.ex
icehaunter: Not that it matters given your request to revert changes to the backend, but without this,
`Logger.error("Testing error", sentry: [tags: %{explicit: "tag"}])` code fails even without other changes, as `sentry_context_from_meta` is a list, not a map, so `Map.merge` fails - it's just not covered by existing tests it seems

lib/sentry/logger_handler.ex
icehaunter: What are your thoughts on `tags` vs `tags_from_metadata` here? I'm OK with either given it lives in `runtime.exs` and isn't repeated often

lib/sentry/logger_backend.ex
whatyouhide: Aaaah I see sorry, my eyes mixed and I thought this was just a formatting change, I now see that `Map.new/1` is called on both lines. But this is still a fix even if you'll revert the rest of the changes here, won't it? Maybe it would be nice to revert it here and open in as a separate PR with a related regression test case.

lib/sentry/logger_handler.ex
whatyouhide: `tags_from_metadata` is more explicit I think.

lib/sentry/logger_backend.ex
icehaunter: Reverted

whatyouhide (approved):
Fantastic! 🍬

PR #842, opened 2024-12-27
Support for transactions

solnic:
This was extracted from #784

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -285,7 +285,7 @@ defmodule Sentry.Client do
   @spec render_transaction(%Transaction{}) :: map()
   def render_transaction(%Transaction{} = transaction) do
     transaction
-    |> Transaction.to_map()
+    |> Transaction.to_payload()
     |> Map.merge(%{
       platform: "elixir",
       sdk: %{
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -100,8 +100,7 @@ defmodule Sentry.Client do
     client = Config.client()
 
     # This is a "private" option, only really used in testing.
-    request_retries =
-      Application.get_env(:sentry, :request_retries, Transport.default_retries())
+    request_retries = Application.get_env(:sentry, :request_retries, Transport.default_retries())
 
     client_report
     |> Envelope.from_client_report()
@@ -115,6 +114,7 @@ defmodule Sentry.Client do
     client = Keyword.get_lazy(opts, :client, &Config.client/0)
     sample_rate = Keyword.get_lazy(opts, :sample_rate, &Config.sample_rate/0)
     before_send = Keyword.get_lazy(opts, :before_send, &Config.before_send/0)
+    after_send_event = Keyword.get_lazy(opts, :after_send_event, &Config.after_send_event/0)
 
     request_retries =
       Keyword.get_lazy(opts, :request_retries, fn ->
@@ -123,13 +123,9 @@ defmodule Sentry.Client do
 
     with :ok <- sample_event(sample_rate),
          {:ok, %Transaction{} = transaction} <- maybe_call_before_send(transaction, before_send) do
-      case encode_and_send(transaction, result_type, client, request_retries) do
-        {:ok, id} ->
-          {:ok, id}
-
-        {:error, %ClientError{} = error} ->
-          {:error, error}
-      end
+      send_result = encode_and_send(transaction, result_type, client, request_retries)
+      _ignored = maybe_call_after_send(transaction, send_result, after_send_event)
+      send_result
     else
       :unsampled ->
         ClientReport.Sender.record_discarded_events(:sample_rate, [transaction])
@@ -194,7 +190,7 @@ defmodule Sentry.Client do
     """
   end
 
-  defp maybe_call_after_send(%Event{} = event, result, callback) do
+  defp maybe_call_after_send(event, result, callback) do
     message = ":after_send_event must be an anonymous function or a {module, function} tuple"
 
     case callback do
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -113,18 +113,25 @@ defmodule Sentry.Client do
 
     result_type = Keyword.get_lazy(opts, :result, &Config.send_result/0)
     client = Keyword.get_lazy(opts, :client, &Config.client/0)
+    sample_rate = Keyword.get_lazy(opts, :sample_rate, &Config.sample_rate/0)
 
     request_retries =
       Keyword.get_lazy(opts, :request_retries, fn ->
         Application.get_env(:sentry, :request_retries, Transport.default_retries())
       end)
 
-    case encode_and_send(transaction, result_type, client, request_retries) do
-      {:ok, id} ->
-        {:ok, id}
+    with :ok <- sample_event(sample_rate) do
+      case encode_and_send(transaction, result_type, client, request_retries) do
+        {:ok, id} ->
+          {:ok, id}
 
-      {:error, %ClientError{} = error} ->
-        {:error, error}
+        {:error, %ClientError{} = error} ->
+          {:error, error}
+      end
+    else
+      :unsampled ->
+        ClientReport.Sender.record_discarded_events(:sample_rate, [transaction])
+        :unsampled
     end
   end
 
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -114,13 +114,15 @@ defmodule Sentry.Client do
     result_type = Keyword.get_lazy(opts, :result, &Config.send_result/0)
     client = Keyword.get_lazy(opts, :client, &Config.client/0)
     sample_rate = Keyword.get_lazy(opts, :sample_rate, &Config.sample_rate/0)
+    before_send = Keyword.get_lazy(opts, :before_send, &Config.before_send/0)
 
     request_retries =
       Keyword.get_lazy(opts, :request_retries, fn ->
         Application.get_env(:sentry, :request_retries, Transport.default_retries())
       end)
 
-    with :ok <- sample_event(sample_rate) do
+    with :ok <- sample_event(sample_rate),
+         {:ok, %Transaction{} = transaction} <- maybe_call_before_send(transaction, before_send) do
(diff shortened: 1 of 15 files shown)


lib/sentry/envelope.ex
```diff
@@ -2,11 +2,21 @@ defmodule Sentry.Envelope do
...
+    Event,
+    Transaction,
+    UUID
+  }
 
   @type t() :: %__MODULE__{
           event_id: UUID.t(),
-          items: [Event.t() | Attachment.t() | CheckIn.t() | ClientReport.t(), ...]
+          items: [
+            Attachment.t() | CheckIn.t() | ClientReport.t() | Event.t() | Transaction.t()
```
whatyouhide, line 18: The `...` means a non-empty list, not sure why it was removed?

```suggestion
            Attachment.t() | CheckIn.t() | ClientReport.t() | Event.t() | Transaction.t(), ...
```

lib/sentry/envelope.ex
```diff
@@ -46,13 +56,31 @@ defmodule Sentry.Envelope do
     }
   end
 
+  @doc """
+  Creates a new envelope containing a transaction with spans.
+  """
+  @spec from_transaction(Sentry.Transaction.t()) :: t()
```
whatyouhide, line 62: You aliased it 🙃 

```suggestion
  @spec from_transaction(Transaction.t()) :: t()
```

lib/sentry/envelope.ex
```diff
@@ -126,4 +154,15 @@ defmodule Sentry.Envelope do
         throw(error)
     end
   end
+
+  defp item_to_binary(json_library, %Transaction{} = transaction) do
+    case transaction |> Sentry.Client.render_transaction() |> json_library.encode() do
+      {:ok, encoded_transaction} ->
+        header = ~s({"type": "transaction", "length": #{byte_size(encoded_transaction)}})
```
whatyouhide, line 161: Less bytes for free 🙃 

```suggestion
        header = ~s({"type":"transaction","length":#{byte_size(encoded_transaction)}})
```

lib/sentry/options.ex
```diff
@@ -191,6 +194,10 @@ defmodule Sentry.Options do
 
   @create_event_opts_schema NimbleOptions.new!(@create_event_opts_schema_as_keyword)
 
+  @send_transaction_opts_schema_as_keyword Keyword.merge(@common_opts_schema_as_keyword, [])
+
+  @send_transaction_opts_schema NimbleOptions.new!(@send_transaction_opts_schema_as_keyword)
```
whatyouhide, line 199: We don't need both and the `merge(_, [])` I think. Just go with

```suggestion
  @send_transaction_opts_schema NimbleOptions.new!(@@common_opts_schema_as_keyword)
```

lib/sentry/options.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Sentry.Options do
   @moduledoc false
 
-  @send_event_opts_schema_as_keyword [
+  @common_opts_schema_as_keyword [
```
whatyouhide, line 4: Can you link to the docs that say that sample rate and before_send callbacks don't applly to transactions?

lib/sentry/transaction.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Sentry.Transaction do
```
whatyouhide, line 1: This needs docs

lib/sentry/transaction.ex
```diff
@@ -0,0 +1,75 @@
...
+  @type t() :: %__MODULE__{}
+
+  alias Sentry.{Config, UUID}
+
+  @enforce_keys ~w(event_id span_id spans)a
+
+  defstruct [
+    :event_id,
+    :environment,
+    :span_id,
```
whatyouhide, line 11:
```suggestion
  defstruct @enforce_keys ++ [
```

lib/sentry/transaction.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Sentry.Transaction do
+  @type t() :: %__MODULE__{}
+
+  alias Sentry.{Config, UUID}
+
+  @enforce_keys ~w(event_id span_id spans)a
```
whatyouhide, line 6: Same number of characters and easier for readers less familiar with Elixir. Not a long list, so I don't think `~w` is worth it 🙃 

```suggestion
  @enforce_keys [:event_id, :span_id, :spans]
```

lib/sentry/transaction.ex
```diff
@@ -0,0 +1,75 @@
...
+    :span_id,
+    :transaction,
+    :transaction_info,
+    :contexts,
+    :measurements,
+    :spans,
+    type: "transaction"
+  ]
+
+  def new(attrs) do
```
whatyouhide, line 20: Needs to be `@doc false` too I think?

lib/sentry/transaction.ex
```diff
@@ -0,0 +1,75 @@
...
+
+    root_span
+    |> Sentry.Span.to_map()
+    |> Map.put(:spans, Enum.map(child_spans, &Sentry.Span.to_map/1))
+    |> Map.drop([:description])
+    |> Map.merge(transaction_attrs)
+  end
+end
+
+defmodule Sentry.Span do
```
whatyouhide, line 53: Should this go under `Sentry.Interfaces`?

whatyouhide: Also, docs

lib/sentry/transaction.ex
```diff
@@ -0,0 +1,75 @@
...
+        :event_id,
+        :environment,
+        :transaction,
+        :transaction_info,
+        :contexts,
+        :measurements,
+        :type
+      ])
+
+    {[root_span], child_spans} = Enum.split_with(transaction.spans, &is_nil(&1.parent_span_id))
```
whatyouhide, line 43: I’m not super comfy with this. I think we should make illegal states unrepresentable if possible, so possibly have a `:root_span` and `:child_spans` fields maybe?

whatyouhide (changes requested):
Lots of comments, sorry 😄

lib/sentry/transaction.ex
solnic: @whatyouhide sorry but I'm not entirely sure what you mean here. My understanding is that the span without parent_span_id becomes the transaction with extra information and the rest of spans become its child spans. No?

whatyouhide: I mean that a `transaction` struct with a `span` field where no spans have `parent_span_id == nil` *or* more than one spans have `parent_span_id == nil` is an **invalid transaction**, and we should never get to that state, right? If that is correct, then what I was suggesting here is that we store the root span and the child spans separately.

solnic: @whatyouhide oh so I need to double check this but IIRC a transaction with no child spans is accepted by Sentry

solnic: @whatyouhide I ended up reworking this struct via ef59c6f4598036377950b9dc53b1ee07a7df8e4d

lib/sentry/options.ex
whatyouhide: @solnic this is still pending.

lib/sentry/transaction.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  @moduledoc since: "11.0.0"
+
+  alias Sentry.{Config, UUID, Interfaces.Span, Interfaces.SDK}
+
+  @typedoc since: "11.0.0"
+  @type t() ::
+          %__MODULE__{
+            # Required
+            event_id: <<_::256>>,
```
whatyouhide, line 16:
```suggestion
            event_id: UUID.t(),
```

lib/sentry/transaction.ex
```diff
@@ -0,0 +1,78 @@
...
+            environment: String.t(),
+            transaction: String.t(),
+            transaction_info: map(),
+            measurements: map(),
+            type: String.t(),
+            tags: map(),
+            data: map(),
+
+            # Interfaces
+            spans: [Span.t()],
```
whatyouhide, line 41: A transaction without spans is ok?

lib/sentry/client.ex
```diff
@@ -225,6 +282,19 @@ defmodule Sentry.Client do
...
+  def render_transaction(%Transaction{} = transaction) do
+    transaction
+    |> Transaction.to_payload()
+    |> Map.merge(%{
+      platform: "elixir",
+      sdk: %{
+        name: "sentry.elixir",
+        version: Application.spec(:sentry, :vsn)
+      }
+    })
```
whatyouhide, line 295: Can you elaborate on why this is necessary?

whatyouhide (commented):
Few more comments.

lib/sentry/client.ex
solnic: @whatyouhide I just followed what we do for events, and since a transaction is an event + span info, then I thought it's a good idea to add platform and sdk information too. I'm not sure if it's strictly required though - I'll double check it.

lib/sentry/transaction.ex
solnic: @whatyouhide yes! the only really required piece of span information is meant to go to context.trace which is why I defined it as a required map in the type spec. spans list is only needed when you have child spans, and that's not always the case.

lib/sentry/options.ex
solnic: @whatyouhide sorry but link to what docs? Are you asking to write these docs and link to them?

whatyouhide: No, I’m asking where you found the information that `sample_rate` and `before_send` callbacks don't apply to transactions. As in, docs on the Sentry website that I can go read 😄

solnic: @whatyouhide oh, so I just didn't initially add support for these options. Transaction sampling [is a thing](https://develop.sentry.dev/application-architecture/dynamic-sampling/fidelity-and-biases/#transaction-sampling) in Sentry, so we should support it; and for consistency and feature-completion we should also support callbacks.

I just added support for all three options. One thing we might want to consider is deprecating after_send_event in favor of after_send though.

whatyouhide (approved):

Issue #538, opened 2023-03-10
Investigate quickest path to otel + performance tracing


sl0thentr0py:
@krainboltgreene so there are several things that would need to be done to get performance tracing working with `sentry-elixir`. This SDK was so far handled by the community so it is not on par with what we call our [Unified API](https://develop.sentry.dev/sdk/unified-api/) for making common abstractions across all Sentry SDKs.

I will make 2 lists below, one of the bare minimum that would lead to quick opentelemetry based performance and the other more 'ideal' list where we make the SDK feature compatible in terms of performance with other SDKs.

## Required
* Generalize the `Envelope` abstraction to also be able to send items of [type `transaction`](https://develop.sentry.dev/sdk/envelopes/#transaction).
* Add `Transaction` and `Span` abstractions.
* Add a [`SpanProcessor` and `Propagator`](https://develop.sentry.dev/sdk/performance/opentelemetry/) to hook into opentelemetry.

## Optional
* The elixir SDK is further also missing `Hub` (for concurrency) and `Scope` abstractions but technically they can be ignored if a quick path to opentelemetry support is desired.

I will make a new issue out of this to track it in `sentry-elixir`. Unfortunately, this is a non-trivial amount of development work, so I can't give you very clear cut instructions on how you can contribute, but feel free to try stuff out and make a PR if you're interested and we can collaborate.

If there is sufficient interest from the community, we can also potentially prioritize me working on this as well next quarter.

_Originally posted by @sl0thentr0py in https://github.com/getsentry/sentry/discussions/40712#discussioncomment-5268072_

tsloughter:
What does "Hub" for concurrency in Elixir look like? It looks to be a context store? Like how OpenTelemetry we use the process dictionary?

Does this mean Sentry does no intend to adopt the OpenTelemetry API?

Also, what is "opentelemetry based performance" mean here? Based on what I read of the SpanProcessor model it doesn't read like you'd be relying on the Otel SDK for span operations, so if you found performance better with OpenTelemetry I wouldn't be sure thats the case when combined with the SpanProcessor. But I don't yet fully understand what it is doing, so I may be wrong there :)

sl0thentr0py:
> What does "Hub" for concurrency in Elixir look like? It looks to be a context store? Like how OpenTelemetry we use the process dictionary?

something like that yea, basically the `Hub` needs to be cloned per concurrency unit, so we'll need to spec out what that looks like for elixir.

[...]

PR #843, opened 2024-12-27
OpenTelemetry integration

solnic:
This is replacing #784 and is built on top of #842 

Once #842 is merged I'm gonna update this one and we can shift focus to OT integration again


solnic:
Closing in favor of #875 (which is the initial integration, there will be follow-up issues added under #538)

PR #844, opened 2024-12-30
Allow the usage of JSON for Elixir 1.18+

altjohndev:
Sadly, the [JSON](https://hexdocs.pm/elixir/JSON.html) kernel module does not provide the `encode/1` function, which impacts the usage of this library.

In order to match the kernel implementation, this PR changes the interface from `decode/1` and `encode/1` to `decode!/1` and `encode!/1`.

This PR does not change the default library (keeping it as `Jason`).

It will use `JSON` for tests if the Elixir version is 1.18+. The CI tests both libraries.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,8 +1,16 @@
 import Config
 
+json_library =
+  if Version.compare(System.version(), "1.18.0") == :lt do
+    Jason
+  else
+    JSON
+  end
+
 if config_env() == :test do
   config :sentry,
     environment_name: :test,
+    json_library: json_library,
     tags: %{},
     enable_source_code_context: true,
     root_source_code_paths: [File.cwd!()],
@@ -15,4 +23,4 @@ if config_env() == :test do
   config :logger, backends: []
 end
 
-config :phoenix, :json_library, Jason
+config :phoenix, :json_library, json_library
--- a/config/config.exs
+++ b/config/config.exs
@@ -15,4 +15,4 @@ if config_env() == :test do
   config :logger, backends: []
 end
 
-config :phoenix, :json_library, Jason
+config :phoenix, :json_library, if(Code.ensure_loaded?(JSON), do: JSON, else: Jason)
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,16 +1,8 @@
 import Config
 
-json_library =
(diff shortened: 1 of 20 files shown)


config/config.exs
```diff
@@ -1,8 +1,16 @@
 import Config
 
+json_library =
+  if Version.compare(System.version(), "1.18.0") == :lt do
+    Jason
+  else
+    JSON
+  end
```
whatyouhide, line 8:
```suggestion
json_library =
  if Code.ensure_loaded?(JSON) do
    JSON
  else
    Jason
  end
```

whatyouhide (commented):
I’m thinking maybe a better approach for this would be:

1. Introduce a private module, `Sentry.JSON`, which would act as a temporary middleman (until we will depend on Elixir 1.18+).
2. Check the Elixir version when compiling Sentry and
   1. Use `JSON` if on Elixir 1.18+, no configuration (after all, we'll eventually drop `:json_library` now that JSON is in stdlib) + warn if `:json_library` is configured. This also means we don't need to switch to `encode!` and `decode!`, we know which lib we're using.
   2. Use the configured `:json_library` if on earlier versions.

Thoughts?

altjohndev:
Hi @whatyouhide , applied the suggested changes.

whatyouhide:
> It would be wise to still let developers define custom libraries during this transitioning period.

Why? I can only think of custom protocols and stuff like that, but maybe that's already enough.

altjohndev:
> Why? I can only think of custom protocols and stuff like that, but maybe that's already enough.

Just like this library was not ready to use the `JSON` module, many other libraries and services will require adjustments.

For example, Some services/libraries require a transition for structs that derive from `Jason` (or another encoder) to derive to `JSON` instead.

whatyouhide:
> many other libraries and services will require adjustments

True, but using `JSON` *alongside* other JSON library would totally be fine. Either way, I get the point: let's keep the configuration option then, but change the default to `JSON`. This is a bit of a breaking change but one we can deal with IMO, we just advise people to explicitly set the `:json_library` option if they're on 1.18 but don't want to use `JSON`.

altjohndev:
Thank you for the advice. 🙇 

Let me know if I need to change or improve something.

whatyouhide:
Closing in favor of #845.

PR #758, opened 2024-07-29
Sav/switch hackney to finch

savhappy:
-Switch from Hackney to Finch
-Alter tests to pass with Finch

PR #846, opened 2025-01-03
Add support for Req http client (draft)

krns:
Hello, I would be happy if Req would be supported as http client lib. Here is a working proof of concept.

Since Hackney is the only HTTP client that is supported so far, I had to bend a little in some places. A few comments on this below.

One general question:

Because Req is not known in the Sentry library I get a few warning like:

```
warning: Req.new/1 is undefined (module Req is not available or is yet to be defined). Make sure the module name is correct and has been specified in full (or that an alias has been defined)
warning: Req.merge/2 is undefined (module Req is not available or is yet to be defined). Make sure the module name is correct and has been specified in full (or that an alias has been defined)
warning: Req.post/1 is undefined (module Req is not available or is yet to be defined). Make sure the module name is correct and has been specified in full (or that an alias has been defined)
```

Should I make use of `Code.ensure_loaded?(Req)` and `Application.ensure_all_started(:req)`?

### DIFF
--- a/lib/mix/tasks/sentry.send_test_event.ex
+++ b/lib/mix/tasks/sentry.send_test_event.ex
@@ -26,6 +26,8 @@ defmodule Mix.Tasks.Sentry.SendTestEvent do
     type: :string
   ]
 
+  @requirements ["app.start"]
+
   @impl true
   def run(args) when is_list(args) do
     {opts, args} = OptionParser.parse!(args, strict: @switches)
--- a/lib/sentry/req_client.ex
+++ b/lib/sentry/req_client.ex
@@ -1,0 +1,32 @@
+defmodule Sentry.ReqClient do
+  @behaviour Sentry.HTTPClient
+
+  @moduledoc """
+  The built-in HTTP client.
+
+  This client implements the `Sentry.HTTPClient` behaviour.
+
+  It's based on the [Req](https://github.com/wojtekmach/req) HTTP client,
+  which is an *optional dependency* of this library. If you wish to use another
+  HTTP client, you'll have to implement your own `Sentry.HTTPClient`. See the
+  documentation for `Sentry.HTTPClient` for more information.
+  """
+
+  @impl true
+  def post(url, headers, body) do
+    opts =
+      Sentry.Config.req_opts()
+      |> Keyword.put(:decode_body, false)
+      |> Keyword.put(:url, url)
+      |> Keyword.put(:body, body)
+
+    req =
+      Req.new(opts)
+      |> Req.merge(headers: headers)
+
+    case Req.post(req) do
+      {:ok, %{status: status, headers: headers, body: body}} = result -> {:ok, status, headers, body}
+      {:error, _reason} = error -> error
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


lib/mix/tasks/sentry.send_test_event.ex
```diff
@@ -26,6 +26,8 @@ defmodule Mix.Tasks.Sentry.SendTestEvent do
     type: :string
   ]
 
+  @requirements ["app.start"]
+
```
krns, line 30: I don't like this Req-specific change inside the test command but I left it like this for now because I wanted to get some general feedback first.

Details why this is needed: https://elixirforum.com/t/argumenterror-unknown-registry-req-finch-on-a-simple-http-request-using-req-library/64638/9

lib/sentry/req_client.ex
```diff
@@ -0,0 +1,32 @@
...
+      |> Keyword.put(:decode_body, false)
+      |> Keyword.put(:url, url)
+      |> Keyword.put(:body, body)
+
+    req =
+      Req.new(opts)
+      |> Req.merge(headers: headers)
+
+    case Req.post(req) do
+      {:ok, %{status: status, headers: headers, body: body}} = result -> {:ok, status, headers, body}
```
krns, line 28: We should pattern match when Req is available.

```suggestion
      {:ok, %Req.Response{status: status, headers: headers, body: body}} = result -> {:ok, status, headers, body}
```

I transform the result into the Hackney-inspired form so everything else works.

lib/sentry/transport.ex
```diff
@@ -125,7 +125,7 @@ defmodule Sentry.Transport do
   defp client_post_and_validate_return_value(client, endpoint, headers, body) do
     case client.post(endpoint, headers, body) do
       {:ok, status, resp_headers, resp_body}
-      when is_integer(status) and status in 200..599 and is_list(resp_headers) and
```
krns, line 128: I removed this because it's highly Hackney-specific (Req response readers are a Map).

lib/sentry/req_client.ex
```diff
@@ -0,0 +1,32 @@
...
+  documentation for `Sentry.HTTPClient` for more information.
+  """
+
+  @impl true
+  def post(url, headers, body) do
+    opts =
+      Sentry.Config.req_opts()
+      |> Keyword.put(:decode_body, false)
+      |> Keyword.put(:url, url)
+      |> Keyword.put(:body, body)
```
krns, line 21: Req decodes responses automatically. I choose to disable body decoding so the users can make use of the `json_library` option.

lib/sentry/transport.ex
```diff
@@ -125,7 +125,7 @@ defmodule Sentry.Transport do
...
+      when is_integer(status) and status in 200..599 and
```
krns, line 128: I removed the `is_list` guard because it's highly Hackney-specific (Req response readers are a Map).

whatyouhide:
@krns thanks for the PR, but I don't think this is the right approach. This SDK already supports the [`Sentry.HTTPClient`](https://hexdocs.pm/sentry/10.8.1/Sentry.HTTPClient.html) behaviour. With that, you can use any custom HTTP client (which includes Req) provided you will write a translation layer for it. For example, [here's how to use Finch](https://claudio-ortolina.org/posts/using-finch-with-sentry/)—a blog post from 4 years ago.

We are already looking at replacing the *default* Hackney-based client with one based on Finch (#758), but I’m not convinced this is worth it if Erlang/Elixir will ever get a stdlib HTTP client. We might opt for Mint instead if they don't, since it will be the lowest-level and hopefully most-widely-adopted-as-base client.

Hopefully this all makes sense. Cheers!

PR #853, opened 2025-01-27
OpenTelemetry Support

solnic:
This is a small rework of #784 on top of #842 (that got merged).

---

### Tasks

- [ ] Figure out why tests are failing here
- [ ] Finalize support for Phoenix
- [ ] Finalize support for Oban
- [ ] Finalize support for Ecto
- [ ] Figure out how parent/children spans should be passed as Transactions when span processor captures only a single root span
- [ ] Test it out in a real app


solnic:
@whatyouhide here we go again - do we want to narrow down the scope of this to ship it in smaller chunks, or do we want an all-in approach? :)

solnic:
@stephanie-anderson @whatyouhide @sl0thentr0py hey hey, gonna pick it up again. I added a (very high-level) todo. Please chime in if you'd like to reduce the scope here.

whatyouhide:
I would 100% go with smaller chunks here. We need to PoC this to see how well it scales and all, and I don't think going all in is going to make that easy,

solnic:
Closing this in favor of #875

PR #858, opened 2025-02-07
Keep Newest Breadcrumbs Instead of Oldest

dajinchu:
Currently, exceeding max_breadcrumbs causes only the oldest breadcrumbs to be retained, instead of the newest, as I would have expected. I believe this is a bug?

It's caused by the following code, which inserts the newest breadcrumb at the front of the list, but then trims the list from the back.
https://github.com/getsentry/sentry-elixir/blob/8ee657f110f589a8008f56a957972c391529f879/lib/sentry/context.ex#L426-L427

Similar code exists here but is actually correct because `Context.get_all` returns the breadcrumb list reversed.
https://github.com/getsentry/sentry-elixir/blob/8ee657f110f589a8008f56a957972c391529f879/lib/sentry/event.ex#L216

### DIFF
--- a/lib/sentry/context.ex
+++ b/lib/sentry/context.ex
@@ -424,7 +424,7 @@ defmodule Sentry.Context do
       get_sentry_context()
       |> Map.update(@breadcrumbs_key, [map], fn breadcrumbs ->
         breadcrumbs = [map | breadcrumbs]
-        Enum.take(breadcrumbs, -1 * Sentry.Config.max_breadcrumbs())
+        Enum.take(breadcrumbs, 1 * Sentry.Config.max_breadcrumbs())
       end)
 
     :logger.update_process_metadata(%{@logger_metadata_key => sentry_metadata})
--- a/lib/sentry/context.ex
+++ b/lib/sentry/context.ex
@@ -423,8 +423,7 @@ defmodule Sentry.Context do
     sentry_metadata =
       get_sentry_context()
       |> Map.update(@breadcrumbs_key, [map], fn breadcrumbs ->
-        breadcrumbs = [map | breadcrumbs]
-        Enum.take(breadcrumbs, 1 * Sentry.Config.max_breadcrumbs())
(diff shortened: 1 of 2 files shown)


test/context_test.exs
```diff
@@ -39,6 +39,16 @@ defmodule Sentry.ContextTest do
     assert first_breadcrumb.timestamp <= second_breadcrumb.timestamp
   end
 
+  test "retains most recent breadcrumbs when exceeding max" do
```
dajinchu, line 42: This test fails on master.

whatyouhide:
@dajinchu thanks for the PR! What do other SDKs do, like the Python or Ruby ones? 🙃

dajinchu:
Javascript lib adds crumbs to the end of the list, and slices from the end of the list, thus retaining the most recent.

https://github.com/getsentry/sentry-javascript/blob/22f841ebc38ac0e7da4ae8f6750b43c410070996/packages/core/src/scope.ts#L482-L484

lib/sentry/context.ex
```diff
@@ -424,7 +424,7 @@ defmodule Sentry.Context do
       get_sentry_context()
       |> Map.update(@breadcrumbs_key, [map], fn breadcrumbs ->
         breadcrumbs = [map | breadcrumbs]
-        Enum.take(breadcrumbs, -1 * Sentry.Config.max_breadcrumbs())
+        Enum.take(breadcrumbs, 1 * Sentry.Config.max_breadcrumbs())
```
whatyouhide, line 427: Can you maybe rewrite this code to

```elixir
Enum.take([map | breadcrumbs], Sentry.Config.max_breadcrumbs())
```

?

dajinchu: yeah haha what was I even thinking

whatyouhide (approved):

whatyouhide:
Thank you @dajinchu 💟

PR #866, opened 2025-02-25
Add global `:extra` config

### LINKED ISSUE #863: Support setting global `extra` context
Hi !

We would like to include the time at which OTP started in our Sentry issues. We don't really care about it being searchable, so tags aren't really a good fit, but the `:extra` context seem to be the way to go for that!
Setting this per-process is hard, if possible really, and the information can be known at startup time, so, having the possibility to set it right in the configuration would be great 🙌 

As far as I see, the implementation wouldn't be so hard, it's mostly adding a config key, then using it [here](https://github.com/getsentry/sentry-elixir/blob/3b3ff64280596f48550cdd7c660cddbc82fb532f/lib/sentry/event.ex#L203). I'm wondering if such change would be acceptable 😄

Gladear:
This PR aims to add a new `:extra` configuration option, to be able to add global `:extra` context, passed in to every event.

Resolves #863

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -197,6 +197,13 @@ defmodule Sentry.Config do
       A map of tags to be sent with every event.
       """
     ],
+    extra: [
+      type: {:map, :any, :any},
+      default: %{},
+      doc: """
+      A map of extra data to be sent with every event.
+      """
+    ],
     max_breadcrumbs: [
       type: :non_neg_integer,
       default: 100,
@@ -545,6 +552,9 @@ defmodule Sentry.Config do
 
   @spec tags() :: map()
   def tags, do: fetch!(:tags)
+
+  @spec extra() :: map()
+  def extra, do: fetch!(:extra)
 
   @spec release() :: String.t() | nil
   def release, do: get(:release)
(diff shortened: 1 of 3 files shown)


lib/sentry/config.ex
```diff
@@ -197,6 +197,13 @@ defmodule Sentry.Config do
       A map of tags to be sent with every event.
       """
     ],
+    extra: [
+      type: {:map, :any, :any},
```
whatyouhide, line 201: Couple touches

```suggestion
      type: {:map, :atom, :any},
      type_doc: "`t:Sentry.Context.extra/0`",
```

whatyouhide (approved):

lib/sentry/config.ex
```diff
@@ -197,6 +197,14 @@ defmodule Sentry.Config do
       A map of tags to be sent with every event.
       """
     ],
+    extra: [
+      type: {:map, :atom, :any},
```
whatyouhide, line 201: @Gladear this breaks a bunch of tests but we want extra to be tagged as atoms. Can you fix the tests?

Gladear: Done!

PR #867, opened 2025-02-27
Fix invalid JSON in `:message`

whatyouhide:
We've seen this in some of our internal logs. Easy to break the Sentry reporting by using messages (or interpolation paramaters) that are not JSON-encodable strings.


solnic (approved):
LGTM

PR #872, opened 2025-03-05
Improve Oban error reporting

whatyouhide:
- We don't report `{:discard` or `{:cancel` return values
- We unwrap `Oban.PerformError`s if the return value is an exception
- Moar tests


savhappy (approved):

savhappy (approved):

whatyouhide:
@mbaleczny wait, I am not sure what it is you want to do. Start reporting those again, or **stop** reporting them?

mbaleczny:
@whatyouhide I want to stop reporting them. But I want the crashed and timed out oban jobs being reported. That worked while errors were wrapped with `Oban.PerformError`.

whatyouhide:
That's how things should be working now: Only reports crashes, but not `:cancel`s and `:discard`s. I'll need a bit more info about this as this is the first report of this not working as expected. Could you please open up an issue?

PR #875, opened 2025-03-14
Add SpanProcessor for OpenTelemetry

solnic:
To make this work you need to add opentelemetry deps:

```elixir
      {:opentelemetry, "~> 1.3"},
      {:opentelemetry_api, "~> 1.2"},
      {:opentelemetry_exporter, "~> 1.6"},
      {:opentelemetry_phoenix, "~> 2.0"},
      {:opentelemetry_ecto, "~> 1.2"},
      {:opentelemetry_bandit, "~> 0.2"}
```

Then configure our span processor:

```elixir
config :opentelemetry, span_processor: {Sentry.OpenTelemetry.SpanProcessor, []}
```

Things should start looking more or less like this:

Screenshot 2025-03-19 at 11 32 54
Screenshot 2025-03-19 at 11 31 53
Screenshot 2025-03-19 at 11 33 03
Screenshot 2025-03-19 at 11 29 03


solnic:
```
1) test loads in_app_module_allow_list (Public.Sentry.ConfigTest)
Error:      apps/public/test/public_test.exs:4
     ** (RuntimeError) the Sentry configuration seems to be not available (while trying to fetch :in_app_module_allow_list). This is likely because the :sentry application has not been started yet. Make sure that you start the :sentry application before using any of its functions.

     code: assert Sentry.Config.in_app_module_allow_list() |> Enum.sort() ==
     stacktrace:
       (sentry 10.8.1) lib/sentry/config.ex:698: Sentry.Config.in_app_module_allow_list/0
       test/public_test.exs:5: (test)


Finished in 0.00 seconds (0.00s async, 0.00s sync)
1 test, 1 failure
==> admin
Running ExUnit with seed: 25[95](https://github.com/getsentry/sentry-elixir/actions/runs/13856201920/job/38773483717?pr=875#step:10:96)1, max_cases: 8



  1) test loads in_app_module_allow_list (Admin.Sentry.ConfigTest)
Error:      apps/admin/test/admin_test.exs:4
     ** (RuntimeError) the Sentry configuration seems to be not available (while trying to fetch :in_app_module_allow_list). This is likely because the :sentry application has not been started yet. Make sure that you start the :sentry application before using any of its functions.

     code: assert Sentry.Config.in_app_module_allow_list() |> Enum.sort() ==
     stacktrace:
       (sentry 10.8.1) lib/sentry/config.ex:6[98](https://github.com/getsentry/sentry-elixir/actions/runs/13856201920/job/38773483717?pr=875#step:10:99): Sentry.Config.in_app_module_allow_list/0
       test/admin_test.exs:5: (test)
```

@whatyouhide ☝🏻 I'm having trouble figuring out these failures. Could you tell me how I could debug what exactly causes the :sentry app startup failure? We start it manually in the umbrella test helpers and for whatever reason it started failing when opentelemetry is included.

whatyouhide:
Adding just the OTel deps without any other changes leads to the same issue?

sl0thentr0py:
please write some minimal description in the PR about '
* what functionality this adds 
* instructions to add this to a Phoenix app with opentelemetry instrumentation and test it out

solnic:
> Adding just the OTel deps without any other changes leads to the same issue?

@whatyouhide @sl0thentr0py I narrowed it down to `opentelemetry` dep. When it's included, the app doesn't start in umbrella integration test. We need to make otel deps optional anyway, so it's time to address this. How do we want to approach it though? Should I create sentry-opentelemetry package in this repo?

solnic:
@whatyouhide for the time being I addressed it by using optional deps for otel packages via 6fdf12175af343f0e1dbe92819756d9991ab7566 but then one of the tests in event_test.exs started to fail so I fixed it via d04c5d3925b01964b0296c9f0f93ccf8269abf95 even though I don't understand what's going on there 🙃

whatyouhide:
@solnic is this ready for review? It's still a draft

solnic:
> @solnic is this ready for review? It's still a draft

Not yet. I got it working but phoenix + bandit spans are not processed in a way that would make sense for Sentry for some reason. I've been investigating how to fix. It seems like phoenix spans are not coming in as children of bandit spans so there's a disconnect here. I'll figure it out 🙂

sl0thentr0py:
> We need to make otel deps optional anyway, so it's time to address this. 

If we're just shipping a `SpanProcessor` first, having them as peer dependencies is more than fine. We just document well how to get otel up and running in parallel to sentry and hook them up properly.

We will revisit packaging later once we have a proper working prototype.

solnic:
@sl0thentr0py @whatyouhide this is now open for reviews. I got it *deployed to production* already and it's working well (see screenshots in the description).

sl0thentr0py:
okay I'm taking a preliminary look now

one thing we definitely need is that instead of the boolean `Config.tracing` we need
* a [`traces_sample_rate`](https://develop.sentry.dev/sdk/telemetry/traces/#tracessamplerate)
* also ideally, a [`traces_sampler` callback](https://develop.sentry.dev/sdk/telemetry/traces/#tracessampler)

without these, it is very hard for people to control quota spend so this is a hard requirement

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,199 @@
...
+    require OpenTelemetry.SemConv.Incubating.MessagingAttributes, as: MessagingAttributes
+    @behaviour :otel_span_processor
+
+    require Logger
+
+    alias Sentry.{Transaction, OpenTelemetry.SpanStorage, OpenTelemetry.SpanRecord}
+    alias Sentry.Interfaces.Span
+
+    @impl true
+    def on_start(_ctx, otel_span, _config) do
```
sl0thentr0py, line 18: The way you've implemented it, we can do nothing in `on_start` since the `on_end` logic is sufficient to build the span tree and transaction.

This will also simplify your `SpanStorage.update_span` where you can remove the unnecessary deletes.
(This is also how we've done it in the new python and JS span processors.)

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+          :ets.insert(@table, {span_data.parent_span_id, {span_data, stored_at}})
+        end
+      end)
+    end
+
+    :ok
+  end
+
+  def remove_span(span_id) do
+    case get_root_span(span_id) do
```
sl0thentr0py, line 81: Is this only for removing root spans? If yes, let's rename it to `remove_root_span`.
Otherwise, shouldn't it also have a branch for removing a child and all its children?

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+    |> Enum.each(fn {{:root_span, span_id}, {_span, stored_at}} ->
+      if stored_at < cutoff_time do
+        remove_span(span_id)
+      end
+    end)
+
+    :ets.match_object(@table, {:_, {:_, :_}})
+    |> Enum.each(fn {parent_id, {_span, stored_at}} = object ->
+      cond do
+        not is_nil(get_root_span(parent_id)) and stored_at < cutoff_time ->
```
sl0thentr0py, line 121: don't follow this logic, you're deleting it for both cases so why not just a single
```elixir
        stored_at < cutoff_time ->
          :ets.delete_object(@table, object)

```

sl0thentr0py (changes requested):
some suggestions, looks very good otherwise!

lib/sentry/opentelemetry/span_record.ex
```diff
@@ -0,0 +1,72 @@
+if Code.ensure_loaded?(OpenTelemetry) do
+  defmodule Sentry.OpenTelemetry.SpanRecord do
```
whatyouhide, line 2: This is public (without `@moduledoc false`), I don't think it should be?

lib/sentry/opentelemetry/span_record.ex
```diff
@@ -0,0 +1,72 @@
...
+    require Record
+    require OpenTelemetry
+
+    @fields Record.extract(:span, from_lib: "opentelemetry/include/otel_span.hrl")
+    Record.defrecordp(:span, @fields)
+
+    defstruct @fields ++ [:origin]
+
+    def new(otel_span) do
```
whatyouhide, line 11: Can we add specs to all public functions like we do in the rest of the codebase?

whatyouhide: Also, you can add an additional check here

```suggestion
    def new(span() = otel_span) do
```

lib/sentry/opentelemetry/span_record.ex
```diff
@@ -0,0 +1,72 @@
+if Code.ensure_loaded?(OpenTelemetry) do
+  defmodule Sentry.OpenTelemetry.SpanRecord do
+    require Record
+    require OpenTelemetry
+
+    @fields Record.extract(:span, from_lib: "opentelemetry/include/otel_span.hrl")
+    Record.defrecordp(:span, @fields)
```
whatyouhide, line 7: Easier to read code with this, IMO?

```suggestion
    Record.defrecordp(:otel_span, @fields)
```

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
+defmodule Sentry.OpenTelemetry.SpanStorage do
+  @moduledoc false
+  use GenServer
+
+  @table :span_storage
+  @cleanup_interval :timer.minutes(5)
+  @span_ttl :timer.minutes(30)
+
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link(opts \\ []) do
```
whatyouhide, line 10: We never use the 0-arity version:

```suggestion
  def start_link(opts) when is_list(opts) do
```

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
+defmodule Sentry.OpenTelemetry.SpanStorage do
+  @moduledoc false
+  use GenServer
+
+  @table :span_storage
```
whatyouhide, line 5: This is confusing to me. If we allow the server to be named whatever, but then we enforce the name for the ETS table, then we allow multiple servers to interact with the same table, which is not what we want I think.

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+    name = Keyword.get(opts, :name, __MODULE__)
+    GenServer.start_link(__MODULE__, opts, name: name)
+  end
+
+  @impl true
+  def init(opts) do
+    _table =
+      if :ets.whereis(@table) == :undefined do
+        :ets.new(@table, [:named_table, :public, :bag])
+      end
```
whatyouhide, line 20: If you tie the ETS table to the server, you won't need to check for its existence here.

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+
+    cleanup_interval = Keyword.get(opts, :cleanup_interval, @cleanup_interval)
+    schedule_cleanup(cleanup_interval)
+
+    {:ok, %{cleanup_interval: cleanup_interval}}
```
whatyouhide, line 25: Can you make the state into a struct? It's a pretty common pattern in Elixir and avoids nasty typos and stuff.

```elixir
defstruct [:cleanup_interval]

# and here:
{:ok, %__MODULE__{cleanup_interval: cleanup_interval}}
```

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  def store_span(span_data) when span_data.parent_span_id == nil do
```
whatyouhide, line 28: Nit, but the (*very* strong) convention in Elixir is to group public API functions first, and then callbacks. Can you move `init/1` down together with `handle_info/2`?

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+    stored_at = System.system_time(:second)
+
+    case :ets.lookup(@table, {:root_span, span_data.span_id}) do
+      [] -> :ets.insert(@table, {{:root_span, span_data.span_id}, {span_data, stored_at}})
```
whatyouhide, line 32: We can "flatten" the tuple to simplify, so that it's `{key, span_data, stored_at}`:

```suggestion
      [] -> :ets.insert(@table, {{:root_span, span_data.span_id}, span_data, stored_at})
```

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+
+    if span_data.parent_span_id == nil do
+      case :ets.lookup(@table, {:root_span, span_data.span_id}) do
+        [] ->
+          :ets.insert(@table, {{:root_span, span_data.span_id}, {span_data, stored_at}})
+
+        _ ->
+          :ets.delete(@table, {:root_span, span_data.span_id})
+          :ets.insert(@table, {{:root_span, span_data.span_id}, {span_data, stored_at}})
+      end
```
whatyouhide, line 65: This is all subject to race conditions. You'll want to use `ets:update_element/3` (and leave a note to change it to `ets:update_element/4` when we depend exclusively on OTP 27+).

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+    else
+      existing_spans = :ets.lookup(@table, span_data.parent_span_id)
+
+      Enum.each(existing_spans, fn {parent_id, {span, stored_at}} ->
+        if span.span_id == span_data.span_id do
+          :ets.delete_object(@table, {parent_id, {span, stored_at}})
+          :ets.insert(@table, {span_data.parent_span_id, {span_data, stored_at}})
+        end
+      end)
```
whatyouhide, line 74: Same here.

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+    end
+
+    :ok
+  end
+
+  def remove_span(span_id) do
```
whatyouhide, line 80: Please use `ets:select_delete/2` here, it's going to be more efficient and do it atomically.

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,132 @@
...
+    cleanup_stale_spans()
+    schedule_cleanup(state.cleanup_interval)
+    {:noreply, state}
+  end
+
+  defp schedule_cleanup(interval) do
+    Process.send_after(self(), :cleanup_stale_spans, interval)
+  end
+
+  defp cleanup_stale_spans do
```
whatyouhide, line 107: Use `ets:select_delete/2` here too please.

whatyouhide, line 107: Also, if a root span is expired, should we delete all its children?

test/sentry/opentelemetry/sampler_test.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Sentry.Opentelemetry.SamplerTest do
+  use Sentry.Case, async: false
```
whatyouhide, line 2: Why can't this be async?

test/sentry/opentelemetry/span_processor_test.exs
```diff
@@ -0,0 +1,126 @@
+defmodule Sentry.Opentelemetry.SpanProcessorTest do
+  use Sentry.Case, async: false
+
+  import Sentry.TestHelpers
+
+  alias Sentry.OpenTelemetry.SpanStorage
+
+  setup do
```
whatyouhide, line 8: If you start a server per test (`start_supervised` + not requiring a name) then this can be async and you won't need to do table cleanup.

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,199 @@
...
+
+      SpanStorage.update_span(span_record)
+
+      if span_record.parent_span_id == nil do
+        root_span_record = SpanStorage.get_root_span(span_record.span_id)
+        child_span_records = SpanStorage.get_child_spans(span_record.span_id)
+        transaction = build_transaction(root_span_record, child_span_records)
+
+        result =
+          case Sentry.send_transaction(transaction) do
```
whatyouhide, line 38: Maybe we can do this in a later PR, but I’m pretty scared of sending an HTTP request *every time a span ends*. Don't implementations usually buffer these?

whatyouhide (changes requested):
Looking like a good start but still a bunch of work left to do. Let me know if any of the comments are not clear!

lib/sentry/opentelemetry/span_record.ex
solnic: This breaks:

```
     ** (ArgumentError) expected argument to be a literal atom, literal keyword or a :otel_span record, got runtime: {:span, 135612037030898638558808201682891567440, 9466375428462860189, [], :undefined, "instrumented_function", :internal, -576460748421781665, :undefined, {:attributes, 128, :infinity, 0, %{}}, {:events, 128, 128, :infinity, 0, []}, {:links, 128, 128, :infinity, 0, []}, :undefined, 1, true, :undefined}
```

whatyouhide: This works just fine for me:

```
diff --git lib/sentry/opentelemetry/span_record.ex lib/sentry/opentelemetry/span_record.ex
index 0bfdae3..361f961 100644
--- lib/sentry/opentelemetry/span_record.ex
+++ lib/sentry/opentelemetry/span_record.ex
@@ -4,12 +4,12 @@ if Code.ensure_loaded?(OpenTelemetry) do
     require OpenTelemetry
 
     @fields Record.extract(:span, from_lib: "opentelemetry/include/otel_span.hrl")
-    Record.defrecordp(:span, @fields)
+    Record.defrecordp(:otel_span, @fields)
 
     defstruct @fields ++ [:origin]
 
     def new(otel_span) do
-      otel_attrs = span(otel_span)
+      otel_attrs = otel_span(otel_span)
 
       {:attributes, _, _, _, attributes} = otel_attrs[:attributes]
 ```

File: lib/sentry/opentelemetry/span_storage.ex (hunk 8 above)
Comment (solnic) [reply lib/sentry/opentelemetry/span_storage.ex:5]: @whatyouhide right, I forgot that this was the original idea :/ could you give me some hints how to best implement configurable table names though?

File: lib/sentry/opentelemetry/span_processor.ex (hunk 1 above)
Comment (solnic) [reply lib/sentry/opentelemetry/span_processor.ex:18]: @sl0thentr0py I'll dig into this more but when I make on_start a no-op the tests are no longer passing 🙃

File: lib/sentry/opentelemetry/span_storage.ex (hunk 16 above)
Comment (solnic) [reply lib/sentry/opentelemetry/span_storage.ex:115]: @whatyouhide this is what happens via remove_root_span which by default removes children too

File: lib/sentry/opentelemetry/span_processor.ex (hunk 19 above)
Comment (solnic) [reply lib/sentry/opentelemetry/span_processor.ex:38]: @whatyouhide I actually hit rate limits in my app on prod because of this so it's definitely something we want to address. I'll do a follow-up PR after this gets merged so that we're not lost in the comments and commits.

**Comment by solnic** [comment]
> @solnic I’m assuming you're still working on this so re-request my review if this gets ready again.

@whatyouhide yes, wrapping it up today, still a couple of things to address 🙃

**Comment by solnic** [comment]
@whatyouhide @sl0thentr0py ready for another round of reviews 🤞🏻

**Comment by whatyouhide** [comment]
I'll review tomorrow.

File: lib/sentry/config.ex [hunk 20]
```diff
@@ -125,6 +125,18 @@ defmodule Sentry.Config do
       be used as the value for this option.
       """
     ],
+    traces_sample_rate: [
+      type: {:custom, __MODULE__, :__validate_traces_sample_rate__, []},
+      default: 0.0,
+      doc: """
+      The sample rate for transaction events. A value between 0.0 and 1.0 (inclusive).
+      A value of 0.0 means no transactions will be sampled, while 1.0 means all transactions
+      will be sampled. This value is also used to determine if tracing is enabled - if it's
```
whatyouhide, line 134:
```suggestion
      will be sampled. This value is also used to determine if tracing is enabled: if it's
```

lib/sentry/config.ex
```diff
@@ -125,6 +125,18 @@ defmodule Sentry.Config do
...
+      greater than 0, tracing is enabled.
+
+      This feature requires `opentelemetry` package and you can optionally use integrations with Bandit, Phoenix or Ecto.
```
whatyouhide, line 137: Remove references to libs and maybe point to some general OTel docs that we will have.

```suggestion
      This feature requires `opentelemetry` package.
```

lib/sentry/config.ex
```diff
@@ -125,6 +125,18 @@ defmodule Sentry.Config do
       be used as the value for this option.
       """
     ],
+    traces_sample_rate: [
+      type: {:custom, __MODULE__, :__validate_traces_sample_rate__, []},
+      default: 0.0,
+      doc: """
+      The sample rate for transaction events. A value between 0.0 and 1.0 (inclusive).
```
whatyouhide, line 132: Please backtick around all numbers so that they're rendered as code in Markdown

lib/sentry/opentelemetry/sampler.ex
```diff
@@ -0,0 +1,31 @@
+if Code.ensure_loaded?(:otel_sampler) do
+  defmodule Sentry.OpenTelemetry.Sampler do
```
whatyouhide, line 2: Can you elaborate on why we need a Sentry-specific sampler?

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,200 @@
...
+            {:error, error} ->
+              Logger.error("Failed to send transaction to Sentry: #{inspect(error)}")
+              {:error, :invalid_span}
+          end
+
+        :ok = SpanStorage.remove_root_span(span_record.span_id)
+
+        result
+      else
+        true
```
whatyouhide, line 55: Do we need this?

lib/sentry/opentelemetry/span_record.ex
whatyouhide: I don't think this conversation is resolved?

solnic: @whatyouhide it does not work:

```
  1) test removes span records from storage after sending a transaction (Sentry.Opentelemetry.SpanProcessorTest)
     test/sentry/opentelemetry/span_processor_test.exs:86
     ** (ArgumentError) expected argument to be a literal atom, literal keyword or a :otel_span record, got runtime: {:span, 33858799209017318098040011039887091530, 12090113571782940245, [], :undefined, "instrumented_function", :internal, -576460750127872124, :undefined, {:attributes, 128, :infinity, 0, %{}}, {:events, 128, 128, :infinity, 0, []}, {:links, 128, 128, :infinity, 0, []}, :undefined, 1, true, :undefined}
     code: TestEndpoint.instrumented_function()
     stacktrace:
       (elixir 1.18.1) lib/record.ex:516: Record.__keyword__/3
       (sentry 10.9.0) lib/sentry/opentelemetry/span_record.ex:14: Sentry.OpenTelemetry.SpanRecord.new/1
       (sentry 10.9.0) lib/sentry/opentelemetry/span_processor.ex:20: Sentry.OpenTelemetry.SpanProcessor.on_start/3
       (stdlib 6.1.2) lists.erl:2146: :lists.foldl/3
       (opentelemetry 1.5.0) /usr/local/src/sentry-elixir/deps/opentelemetry/src/otel_span_ets.erl:64: :otel_span_ets.start_span/7
       (opentelemetry 1.5.0) /usr/local/src/sentry-elixir/deps/opentelemetry/src/otel_tracer_default.erl:37: :otel_tracer_default.start_span/4
       (opentelemetry 1.5.0) /usr/local/src/sentry-elixir/deps/opentelemetry/src/otel_tracer_default.erl:43: :otel_tracer_default.with_span/5
       test/sentry/opentelemetry/span_processor_test.exs:91: (test)
```

solnic:
> This feels so close! How far away do we think it is from release?

@jwaldrip yes just waiting for @whatyouhide for the final review :) then I'll follow up with another PR with support for sampler (initial work is already done) and we should be ready to release it 🙂

jwaldrip:
@solnic I am pulling the branch directly and testing with our self hosted instance of sentry. Is there anything we need to do in our API configuration to have these show up in sentry under performance?

solnic:
> @solnic I am pulling the branch directly and testing with our self hosted instance of sentry. Is there anything we need to do in our API configuration to have these show up in sentry under performance?

@jwaldrip I'm sorry but I'm not familiar with self hosted instances. I need to confirm if anything extra is needed. For now please try to replicate setup that I added to our test_integrations/phoenix_app test app 🙂 Thanks for helping with testing!

lib/sentry/opentelemetry/span_record.ex
whatyouhide: Can you push up a branch where this fails? I’m suspecting there's somewhere else in the codebase where you might be calling `span(...)` in a way that passes `{:span, ...}` where it's supposed to be `:otel_span`.

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,147 @@
+defmodule Sentry.OpenTelemetry.SpanStorage do
```
whatyouhide, line 1: Can you `@spec` all the public API functions in this module?

lib/sentry/opentelemetry/sampler.ex
whatyouhide: Can you elaborate on this with links to docs?

test/sentry/opentelemetry/span_processor_test.exs
```diff
@@ -0,0 +1,123 @@
...
+    def child_instrumented_function(name) do
+      Tracer.with_span "child_instrumented_function_#{name}" do
+        :timer.sleep(140)
+      end
+    end
+  end
+
+  @tag span_storage: true
+  test "sends captured root spans as transactions" do
+    put_test_config(environment_name: "test", traces_sample_rate: 1.0)
```
whatyouhide, line 29: The config we use is global, so you cannot use `async: true` for the test. Do we do that in any other test?

test/support/case.ex
```diff
@@ -7,11 +7,28 @@ defmodule Sentry.Case do
...
+
+    case context[:span_storage] do
+      nil -> :ok
+      true -> setup_span_storage([])
+      opts when is_list(opts) -> setup_span_storage(opts)
+    end
+  end
+
+  defp setup_span_storage(opts) do
+    uid = :rand.uniform(1000000)
```
whatyouhide, line 25: Is this supposed to be unique?

test_integrations/phoenix_app/mix.exs
```diff
@@ -55,7 +55,12 @@ defmodule PhoenixApp.MixProject do
       {:bypass, "~> 2.1", only: :test},
       {:hackney, "~> 1.18", only: :test},
 
-      {:sentry, path: "../.."}
+      {:sentry, path: "../.."},
+
+      {:opentelemetry, "~> 1.5", optional: true},
+      {:opentelemetry_api, "~> 1.4", optional: true},
+      {:opentelemetry_exporter, "~> 1.0", optional: true},
+      {:opentelemetry_semantic_conventions, "~> 1.27", optional: true}
```
whatyouhide, line 63: These shouldn't be optional in the test Phoenix app.

whatyouhide (changes requested):
This is not really ready for review, there are quite a few comments that have been resolved without being addressed or comments that have not been addressed in general.

This is a big, complex PR, so I would in general suggest splitting it up in "less controversial" pieces. For example, adding a `:traces_sample` configuration is very straightforward, and if it were its own PR, then it could be a piece we merge and not think about. Same goes for adding tests to the Phoenix app, for example.

In general, for this kind of workflow I suggest something like [stacked PRs](https://axolo.co/blog/p/managing-stacked-pr) if you're familiar with the concept.

solnic:
> This is not really ready for review, there are quite a few comments that have been resolved without being addressed or comments that have not been addressed in general.

Oops, seems like I've got a commit from April 4th that I forgot to push. I used GH PR ext in VSCode to mark things as done but then my branch wasn't synced. Dooh! Sorry about that. The docs were update according to your requests via f88309459326226415d22681d124d1a4be44ab41

> In general, for this kind of workflow I suggest something like [stacked PRs](https://axolo.co/blog/p/managing-stacked-pr) if you're familiar with the concept.

Yes I'm familiar with this approach. Should I start doing it here?

test/support/case.ex
solnic: @whatyouhide yes

lib/sentry/opentelemetry/span_record.ex
solnic: @whatyouhide this comes from otel tracer that we don't control, it passes `{:span, ....}` to `on_start` so it crashes when the tuple is defined as `{:otel_span, ...}`.

solnic: Here's exception from one of the tests:

```
  1) test sends captured root spans as transactions (Sentry.Opentelemetry.SpanProcessorTest)
     test/sentry/opentelemetry/span_processor_test.exs:28
     ** (FunctionClauseError) no function clause matching in Sentry.OpenTelemetry.SpanRecord.new/1

     The following arguments were given to Sentry.OpenTelemetry.SpanRecord.new/1:

         # 1
         {:span, 6348120493686974701910169917510003045, 12343669577700150675, [], :undefined, "child_instrumented_function_one", :internal, -576460750972794583, :undefined, {:attributes, 128, :infinity, 0, %{}}, {:events, 128, 128, :infinity, 0, []}, {:links, 128, 128, :infinity, 0, []}, :undefined, 1, true, :undefined}

     Attempted function clauses (showing 1 out of 1):

         def new({:otel_span, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _} = otel_span)

     code: TestEndpoint.child_instrumented_function("one")
     stacktrace:
       (sentry 10.9.0) lib/sentry/opentelemetry/span_record.ex:13: Sentry.OpenTelemetry.SpanRecord.new/1
       (sentry 10.9.0) lib/sentry/opentelemetry/span_processor.ex:20: Sentry.OpenTelemetry.SpanProcessor.on_start/3
       (stdlib 6.1.2) lists.erl:2146: :lists.foldl/3
       (opentelemetry 1.5.0) /usr/local/src/sentry-elixir/deps/opentelemetry/src/otel_span_ets.erl:64: :otel_span_ets.start_span/7
       (opentelemetry 1.5.0) /usr/local/src/sentry-elixir/deps/opentelemetry/src/otel_tracer_default.erl:37: :otel_tracer_default.start_span/4
       (opentelemetry 1.5.0) /usr/local/src/sentry-elixir/deps/opentelemetry/src/otel_tracer_default.erl:43: :otel_tracer_default.with_span/5
       test/sentry/opentelemetry/span_processor_test.exs:33: (test)
```

lib/sentry/opentelemetry/span_processor.ex
solnic: @whatyouhide yes but not until now I managed to figure out how to do it to make this work, I just pushed 9a5c550cd77c6b00990f39f74c87a93c6601b740 that simplified things

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,200 @@
...
+
+      {"db", db_query_text}
+    end
+
+    defp get_op_description(%{
+           attributes:
+             %{unquote(to_string(MessagingAttributes.messaging_system())) => :oban} = attributes
+         }) do
+      {"queue.process", attributes["oban.job.worker"]}
+    end
```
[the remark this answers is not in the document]
solnic: @whatyouhide `queue.process` is a Sentry concept, that's why we have to handle it ourselves here. From Sentry docs:

**queue.process**

Based on the knowledge sources, queue.process is a span operation used in Sentry's tracing instrumentation for queue consumers. This operation is used to track when messages are delivered to or processed by a consumer in a messaging system.

**Consumer Instrumentation**

To implement queue consumer instrumentation with Sentry, you need to:
Create a span with the operation `queue.process`
Include specific attributes to enrich your consumer spans with `queue` metrics

lib/sentry/opentelemetry/span_record.ex
solnic: > Can you push up a branch where this fails? I’m suspecting there's somewhere else in the codebase where you might be calling `span(...)` in a way that passes `{:span, ...}` where it's supposed to be `:otel_span`.

I pushed a branch called 874-add-spanprocessor-for-otel-record-rename with this commit  177415a5691c2d851c8fd22a3cca233925fafbdf

lib/sentry/opentelemetry/sampler.ex
yordis: I would expect the Sampler to be my application-specific concern; this module seems unnecessary to me

test/support/case.ex
whatyouhide: Then `:rand.uniform/1` won't guarantee that in any way. Use `System.unique_integer([:positive])` instead.

jwaldrip:
@solnic are these the final parts that need to be reviewed by @whatyouhide? Hoping we can see this released this month. 🥺

sonny3690:
@solnic this is a critical feature for us that would make us use Sentry! Thank you so much for your hard work here <3

sergiotapia:
Thank you @solnic very excited for this release. I need this badly! Appreciate all your hard work!

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,151 @@
...
+  @moduledoc false
+  use GenServer
+
+  defstruct [:cleanup_interval, :table_name]
+
+  alias Sentry.OpenTelemetry.SpanRecord
+
+  @cleanup_interval :timer.minutes(5)
+  @span_ttl :timer.minutes(30)
```
whatyouhide, line 10: Can we refactor these and the options below to use NimbleOptions like the rest of the library? The schema would look something like:

```elixir
@opts [
  cleanup_interval: [
    type: :pos_integer,
    default: :timer.minutes(5)
  ],
  span_ttl: [
    type: :timeout,
    default: :timer.minutes(30)
  ]
]
```

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,151 @@
...
+
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link(opts) when is_list(opts) do
+    name = Keyword.get(opts, :name, __MODULE__)
+    GenServer.start_link(__MODULE__, opts, name: name)
+  end
+
+  @impl true
+  def init(opts) do
+    table_name = Keyword.get(opts, :table_name, default_table_name())
```
whatyouhide, line 20: If the name of this server is customizable, I don't think we need to make this customizable too. We can just always do `Module.concat(name, ETSTable)`.

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,151 @@
...
+    {:ok, %__MODULE__{cleanup_interval: cleanup_interval, table_name: table_name}}
+  end
+
+  @impl true
+  def handle_info(:cleanup_stale_spans, state) do
+    cleanup_stale_spans(state.table_name)
+    schedule_cleanup(state.cleanup_interval)
+
+    {:noreply, state}
+  end
```
whatyouhide, line 36: Nit, but can we group public API and callbacks together? Usual flow of a GenServer is

1. `start_link`
2. Public API (`store_span/2` and friends)
3. State (`defstruct`)
4. Callbacks
5.

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,151 @@
...
+      [] -> nil
+    end
+  end
+
+  @spec get_child_spans(String.t(), keyword()) :: [SpanRecord.t()]
+  def get_child_spans(parent_span_id, opts \\ []) do
+    table_name = Keyword.get(opts, :table_name, default_table_name())
+
+    :ets.match_object(table_name, {{:child_span, parent_span_id, :_}, :_, :_})
+    |> Enum.map(fn {_key, span_data, _stored_at} -> span_data end)
```
whatyouhide, line 72: You can change the match spec to only return `span_data` here and not have to do the `Enum.map/2`.

whatyouhide (commented):
Did another round of review.

lib/sentry/opentelemetry/sampler.ex
sl0thentr0py: Controlling sampling in an easy way is very central to our performance offering business wise since that is what we charge users. We have a unified way we expose sampling on our SDKs (`traces_sample_rate/traces_sampler`) and we will generally not diverge in the end-user experience here.

sergiotapia: As an end-user, the reason why we went with Sentry vs AppSignal is that Sentry allows us to sample our traces and stuff to avoid getting on beefier plans for some projects.

whatyouhide: @sl0thentr0py but OTel has built-in sampling, and this integration is all based on OTel. Do we really need a separate way to do this, that is not how you'd do it with just OTel?

sl0thentr0py: its not just about functionality, its about a unified product and developer experience

whatyouhide: Right, I don't think this is the right move—the Elixir SDK is unique in it being (AFAIK) the only Sentry SDK that is solely based on OTel. Either way, your choice of course. @solnic let's split this up into yet its own separate PR please.

sl0thentr0py:
hey everyone, I will take one final look at this now and we are going to prioritize shipping a first version. Other nitpicks can be handled later and we will iterate on it. Thanks for all the reviews and discussions but this has been open for a fairly long time so now we should wrap it up. 

@solnic told me he's been running it in his production app for a while so that's also good enough as a signal for me to get this out.

lib/sentry/client.ex
```diff
@@ -117,7 +117,7 @@ defmodule Sentry.Client do
 
     result_type = Keyword.get_lazy(opts, :result, &Config.send_result/0)
     client = Keyword.get_lazy(opts, :client, &Config.client/0)
-    sample_rate = Keyword.get_lazy(opts, :sample_rate, &Config.sample_rate/0)
+    sample_rate = Keyword.get_lazy(opts, :sample_rate, &Config.traces_sample_rate/0)
```
sl0thentr0py, line 120: the errors `sample_rate` and transactions `traces_sample_rate` are two different options and should be kept separate and used for the relevant sampling decision, so please don't replace the original one.

sl0thentr0py (changes requested):

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,196 @@
...
+        result =
+          case Sentry.send_transaction(transaction) do
+            {:ok, _id} ->
+              true
+
+            :ignored ->
+              true
+
+            {:error, error} ->
+              Logger.error("Failed to send transaction to Sentry: #{inspect(error)}")
```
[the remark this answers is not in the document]
solnic: Yes good catch. I changed it to `warning`.

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,196 @@
...
+    end
+
+    @impl true
+    def on_end(otel_span, _config) do
+      span_record = SpanRecord.new(otel_span)
+
+      SpanStorage.store_span(span_record)
+
+      if span_record.parent_span_id == nil do
+        child_span_records = SpanStorage.get_child_spans(span_record.span_id)
```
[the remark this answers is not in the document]
solnic: @whatyouhide OTel guarantees that `on_end` is not called for the same span more than once because it removes the span from its storage before sending a span to processors's `on_end`, so even if something may caused another call to `on_end` for the same span, that span will not be found and processor callbacks won't be called again.

lib/sentry/config.ex
```diff
@@ -627,6 +642,9 @@ defmodule Sentry.Config do
   @spec integrations() :: keyword()
   def integrations, do: fetch!(:integrations)
 
+  @spec tracing?() :: boolean()
+  def tracing?, do: fetch!(:traces_sample_rate) > 0.0
```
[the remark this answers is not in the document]
solnic: @whatyouhide actually, that's how this works across the SDKs, so we need to stay consistent over here

lib/sentry/client.ex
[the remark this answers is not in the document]
solnic: @whatyouhide I'm gonna do a follow up PR with improved sampler which is meant to use this config option. @sl0thentr0py told me we need to sample sooner than in `send_transaction` and that the sampler is the correct place so we'd be dropping spans sooner rather than aggregating them in our storage to potentially drop later.

lib/sentry/config.ex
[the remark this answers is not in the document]
whatyouhide: yuck 😞

lib/sentry/client.ex
whatyouhide: This comment is not correct. It's not replacing the original one, it's switching from `sample_rate` to `traces_sample_rate` in this function, which reports *transactions*. We didn't have `traces_sample_rate` before, so we used `sample_rate` as a stopgap.

sl0thentr0py: I'm saying we need both

test/sentry/opentelemetry/span_processor_test.exs
```diff
@@ -0,0 +1,123 @@
...
+        flunk("Invalid ISO8601 timestamp: #{timestamp}, reason: #{inspect(reason)}")
+    end
+  end
+
+  defp assert_valid_trace_id(trace_id) do
+    assert is_binary(trace_id), "Expected trace_id to be a string"
+    assert String.length(trace_id) == 32, "Expected trace_id to be 32 characters long #{trace_id}"
+
+    assert String.match?(trace_id, ~r/^[a-f0-9]{32}$/),
+           "Expected trace_id to be a lowercase hex string"
```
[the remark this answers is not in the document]
solnic: No this actually does not pass

whatyouhide:
As discussed in Discord, I won't re-review this. Thanks for all the work @solnic!

lib/sentry/opentelemetry/span_storage.ex
```diff
@@ -0,0 +1,151 @@
...
+    :ok
+  end
+
+  defp schedule_cleanup(interval) do
+    Process.send_after(self(), :cleanup_stale_spans, interval)
+  end
+
+  defp cleanup_stale_spans(table_name) do
+    now = System.system_time(:second)
+    cutoff_time = now - @span_ttl
```
sl0thentr0py, line 129: iiuc, `now` is in seconds whereas `@span_ttl` is in milliseconds?

lib/sentry/opentelemetry/span_processor.ex
```diff
@@ -0,0 +1,196 @@
...
+        span_id: span_record.span_id,
+        parent_span_id: span_record.parent_span_id,
+        op: op,
+        description: description,
+        origin: span_record.origin,
+        data: span_record.attributes
+      }
+    end
+
+    defp build_otel_context(span_record), do: span_record.attributes
```
sl0thentr0py, line 101: on other sdks we actually populate this with the `resource` attributes
https://opentelemetry.io/docs/concepts/resources/

if it's not possible to access those attributes, I'd just leave this out since it's redundant

sl0thentr0py (changes requested):
just 2 small comments, then good to go!

lib/sentry/opentelemetry/span_processor.ex
solnic: @sl0thentr0py huh good catch! I removed it for now

sl0thentr0py (approved):
lgtm!! :shipit: 
and thank you for the feature

jwaldrip:
@solnic Why did this get reverted? Is there no OpenTelemetry span support now?

solnic:
@jwaldrip we had to revert to push 10.10.0 with OTP 28 fixes. It'll be restored soon with improved Sampler that's the missing piece for finally releasing it.

PR #876, opened 2025-03-19
improvement: add igniter installer

zachdaniel:
with this change, the installation instructions can be as simple as

`mix igniter.install sentry --dsn "your_dsn"`

For the release task change, a notice is emitted instructing the user to add the relevant task to their release process.

All good if you'd rather not add this change, but the installation steps for sentry were a great candidate for an installer.

My next steps would be:

1. either to contribute this package the necessary telemetry listeners to trace Ash applications automatically
2. create an `ash_sentry` package that composes this installer and adds the necessary integration.

We can talk about Ash integration in a separate issue, just thought that I'd explain where I'm looking to go.


zachdaniel:
Oh, and you can test it yourself creating a new app:

```elixir
mix igniter.new my_app --with phx.new --install sentry@github:zachdaniel/sentry-elixir
```

lib/mix/tasks/sentry.install.ex
```diff
@@ -0,0 +1,188 @@
...
+        "prod.exs",
+        app_name,
+        [:enable_source_code_context],
+        true
+      )
+      |> Igniter.Project.Config.configure(
+        "prod.exs",
+        app_name,
+        [:root_source_code_paths],
+        {:code, cwd_code}
```
whatyouhide, line 87: Can we trim the `quote`d vars above a bit?

```suggestion
        {:code, quote(do: [File.cwd!()]}
```

lib/mix/tasks/sentry.install.ex
```diff
@@ -0,0 +1,188 @@
...
+        "prod.exs",
+        app_name,
+        [:dsn],
+        igniter.args.options[:dsn]
+      )
+      |> Igniter.Project.Config.configure(
+        "prod.exs",
+        app_name,
+        [:environment_name],
+        {:code, mix_env_code}
```
whatyouhide, line 75:
```suggestion
        {:code, quote(do: Mix.env())}
```

lib/mix/tasks/sentry.install.ex
```diff
@@ -0,0 +1,188 @@
...
+        [:root_source_code_paths],
+        {:code, cwd_code}
+      )
+      |> configure_phoenix()
+      |> add_logger_handler()
+      |> Igniter.add_notice("""
+      Sentry: 
+
+        Add a call to mix sentry.package_source_code in your release script to 
+        make sure the stacktraces you receive are complete.
```
whatyouhide, line 95:
```suggestion
        Add a call to
        
          mix sentry.package_source_code
          
        in your release script to make sure the stacktraces you receive
        are correctly categorized.
```

lib/mix/tasks/sentry.install.ex
```diff
@@ -0,0 +1,188 @@
...
+
+    defp add_logger_handler(igniter) do
+      app_module = Igniter.Project.Application.app_module(igniter)
+
+      Igniter.Project.Module.find_and_update_module(igniter, app_module, fn zipper ->
+        with {:ok, zipper} <- Igniter.Code.Function.move_to_def(zipper, :start, 2),
+             {:ok, zipper} <- Igniter.Code.Common.move_to_do_block(zipper) do
+          code =
+            """
+            :logger.add_handler(:my_sentry_handler, Sentry.LoggerHandler, %{
```
whatyouhide, line 107:
```suggestion
            :logger.add_handler(:sentry_handler, Sentry.LoggerHandler, %{
```

lib/mix/tasks/sentry.install.ex
```diff
@@ -0,0 +1,188 @@
...
+
+    ## Example
+
+    ```bash
+    #{example()}
+    ```
+
+    ## Options
+
+    * `--dsn` - Your sentry `dsn`
```
whatyouhide, line 24: Nits

```suggestion
      * `--dsn` - Your Sentry DSN.

```

lib/mix/tasks/sentry.install.ex
```diff
@@ -0,0 +1,188 @@
+defmodule Mix.Tasks.Sentry.Install.Docs do
+  @moduledoc false
+
+  def short_doc do
+    "Installs sentry. Requires igniter to be installed."
```
whatyouhide, line 5:
```suggestion
    "Configures Sentry. Requires igniter to be installed."
```

lib/mix/tasks/sentry.install.ex
```diff
@@ -0,0 +1,188 @@
...
+  end
+
+  def example do
+    "mix sentry.install --dsn <your_dsn>"
+  end
+
+  def long_doc do
+    """
+    #{short_doc()}
```
whatyouhide, line 14: Can we add a description of what Igniter is here?

lib/mix/tasks/sentry.install.ex
```diff
@@ -0,0 +1,188 @@
+defmodule Mix.Tasks.Sentry.Install.Docs do
```
whatyouhide, line 1: This is not installing Sentry, it's configuring it, right? Can we reflect that in the name?

whatyouhide (commented):
I like the direction!

lib/mix/tasks/sentry.install.ex
zachdaniel: So the way igniter works is that when you run `mix igniter.install package`, we look for a task defined by that package called `<package>.install` that defines an `igniter/2` function, so the task actually *must* be called `Mix.Tasks.Sentry.Install` for that to work. In igniter we call these tasks an installer even in the cases that its already present in your mix.exs.

And from the end user's perspective, they don't typically call this task directly, they run `mix igniter.install sentry` which will download and install the package.

For a preview of what I mean:

```
mix archive.install hex igniter_new

mix igniter.new --with phx.new my_app
cd my_app
mix igniter.install sentry

# or one shot it with
mix igniter.new --with phx.new my_app --install sentry
```

zachdaniel:
Something also to keep in mind is that for users who don't want to use igniter or can't for whatever reason, this shouldn't affect them at all given that it is an optional dependency. What we typically do is add a tabset to our installation guides, one for "with igniter (recommended)" and one for "manual". https://hexdocs.pm/ash/get-started.html#create-a-new-project

whatyouhide (approved):
Nope, this looks good now. @sl0thentr0py you Sentry folks ok with this? Should help conversion ahah! 😄

zachdaniel:
Any updates on how we're feeling about this?

solnic:
Thanks @zachdaniel - I like the idea behind igniter. The only concern here is that having it officially in the SDK adds a little bit of extra maintenance burden for us. I can see that igniter has 0.5M downloads already, so I guess it's becoming popular quickly though (congrats!).

Is the idea here that each lib / framework is meant to provide their igniter integration? I'm asking because it's also common for tools like this to ship *some* integrations built-in instead.

zachdaniel:
In general yes the idea is that the installer lives directly inside the thing being installed. This ensures that it's versioned alongside the thing that is being installed, allowing hex version resolution to be the driver of which installer you get.

We've talked about adding ways to set up one shot installers that remove themselves after installation and don't live in the package being installed, but they lose a lot of the properties that make an installer good (like the above stuff).

I wouldn't personally maintain the sentry installer as something built in to igniter(we don't have any built in installers), although I'm happy support any igniter related issues and to be in the loop for the work done in this PR, i.e I'll support it but the buck stops w/ the project maintainers not me.

No pressure from me to accept the PR at the end of the day 😄 you could also just leave it open and wait to see if it becomes a truly ubiquitous standard?

sl0thentr0py:
i'm not familiar enough with the elixir ecosystem to tell how popular this is and so on, but if it's just an optional thing I don't mind.
Final call is yours @solnic :)

whatyouhide:
> how popular this is and so on

Not very popular yet, but this is not a lot to maintain (and it might become popular). I would just really strongly advise to add this as a dependency to Sentry itself ever in the future, but I don't think we would need that.

solnic:
> > Not very popular yet, but this is not a lot to maintain (and it might become popular).
> 
> I'm not being defensive I promise, but it depends a bit on how you define popular. Oban, live debugger, error tracker, all of Ash, inertia, tidewave, mishka chelekom, beacon, oidcc etc. all have igniter installers. It's not ubiquitous because many packages have no need of an installer. Again, its fine to say its not popular, just pointing out that a fair few current important packages are using igniter for their onboarding 😄
> 
> Something we can also do is add your CI to our test suite: https://github.com/ash-project/igniter/actions/runs/15287244884
> 
> So if we can run your tests with `--only igniter`, we can ensure that your installer doesn't break w/ new versions of igniter.
> 
> @solnic happy to help if any issues arise.

@zachdaniel alright so let's get this rebased/updated + fix CI and we could merge it in 🙂

zachdaniel:
Looks like some issues come from building for pre 1.15 versions of Elixir, due to spitfire. This wouldn't cause problems for existing users on older versions of sentry, as igniter is an optional dependency. So we'd need to do something like only run CI for 1.15+ versions of Elixir 🥶

zachdaniel:
I'll update the tests not to require phx_new as an optional dependency. That is a utility some projects choose to use, but its not necessary 😄

zachdaniel:
Ready for another run 😄

mise.toml
```diff
@@ -0,0 +1,2 @@
+[tools]
+elixir = "1.18.4"
```
solnic, line 2: Please remove this, pretty sure it was added by a mistake.

mix.lock
```diff
@@ -1,52 +1,64 @@
 %{
-  "bandit": {:hex, :bandit, "1.6.11", "2fbadd60c95310eefb4ba7f1e58810aa8956e18c664a3b2029d57edb7d28d410", [:mix], [{:hpax, "~> 1.0", [hex: :hpax, repo: "hexpm", optional: false]}, {:plug, "~> 1.14", [hex: :plug, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:thousand_island, "~> 1.0", [hex: :thousand_island, repo: "hexpm", optional: false]}, {:websock, "~> 0.5", [hex: :websock, repo: "hexpm", optional: false]}], "hexpm", "543f3f06b4721619a1220bed743aa77bf7ecc9c093ba9fab9229ff6b99eacc65"},
+  "bandit": {:hex, :bandit, "1.6.1", "9e01b93d72ddc21d8c576a704949e86ee6cde7d11270a1d3073787876527a48f", [:mix], [{:hpax, "~> 1.0", [hex: :hpax, repo: "hexpm", optional: false]}, {:plug, "~> 1.14", [hex: :plug, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:thousand_island, "~> 1.0", [hex: :thousand_island, repo: "hexpm", optional: false]}, {:websock, "~> 0.5", [hex: :websock, repo: "hexpm", optional: false]}], "hexpm", "5a904bf010ea24b67979835e0507688e31ac873d4ffc8ed0e5413e8d77455031"},
```
solnic, line 2: In #894 we bumped bandit and here we go back to 1.6.1 - could you set it again to 1.6.11 like we have in master?

solnic (changes requested):

mise.toml
zachdaniel: Yes, not sure how that happened, thanks.

zachdaniel:
PR feedback/failures should be resolved, last major issue is the 1.13 issue. I'm going to try and make a PR to spitfire that lets it at least compile on 1.13. Our usage of it in igniter doesn't care if it blows up when we try to use it, as what we use it for only works on recent Elixir versions anyway.

solnic:
> I'm assuming yanking 1.13 from CI is off the table 😅

Yes we need to support it. Let's just move igniter tests to test_integration - we only run them on Elixir > 1.16.

zachdaniel:
I'm not sure that will work, because it's the dependency failing to compile. We'd have to make the dependency only appear conditionally using some kind of flag. I'll see about getting spitfire compiling on 1.13 😄

zachdaniel:
https://github.com/elixir-tools/spitfire/pull/62

Opened a PR to make spitfire compile on 1.13

zachdaniel:
Updated spitfire

solnic (approved):

PR #879, opened 2025-04-04
Add `owner` option for check-ins

Miradorn:
resolves https://github.com/getsentry/sentry-elixir/issues/824 for real :D

### DIFF
--- a/lib/sentry/check_in.ex
+++ b/lib/sentry/check_in.ex
@@ -58,7 +58,8 @@ defmodule Sentry.CheckIn do
                 optional(:max_runtime) => number(),
                 optional(:failure_issue_threshold) => number(),
                 optional(:recovery_threshold) => number(),
-                optional(:timezone) => String.t()
+                optional(:timezone) => String.t(),
+                optional(:owner) => String.t()
               },
           contexts: Interfaces.context()
         }
@@ -127,6 +128,7 @@ defmodule Sentry.CheckIn do
         failure_issue_threshold: number_schema_opts,
         recovery_threshold: number_schema_opts,
         timezone: [type: :string],
+        owner: [type: :string],
         schedule: [
           type:
             {:or,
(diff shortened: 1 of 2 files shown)


lib/sentry/check_in.ex
```diff
@@ -127,6 +128,7 @@ defmodule Sentry.CheckIn do
         failure_issue_threshold: number_schema_opts,
         recovery_threshold: number_schema_opts,
         timezone: [type: :string],
+        owner: [type: :string],
```
whatyouhide, line 131: Can we say something like this in the docs for this type:

```
The type for the check-in struct.

`:owner` is available since v10.10.0.
```

Miradorn: good call. I extracted the config into its own type to make it clearer

lib/sentry/check_in.ex
```diff
@@ -40,8 +40,25 @@ defmodule Sentry.CheckIn do
               unit: :year | :month | :week | :day | :hour | :minute
             }
 
+  @typedoc """
+  The type for the monitor_config that can be part of the checking
+
+  - `:owner` is available since v10.10.0.
+  """
```
whatyouhide, line 47:
```suggestion
  """
  @typedoc since: "10.10.0"
```

lib/sentry/check_in.ex
```diff
@@ -40,8 +40,25 @@ defmodule Sentry.CheckIn do
...
+          optional(:failure_issue_threshold) => number(),
+          optional(:recovery_threshold) => number(),
+          optional(:timezone) => String.t(),
+          optional(:owner) => String.t()
+        }
+
   @typedoc """
   The type for the check-in struct.
+
+  - `monitor_config.owner` is available since v10.10.0.
```
whatyouhide, line 61:
```suggestion

  `:owner` within `:monitor_config` is available since v10.10.0.
```

whatyouhide (approved):

whatyouhide:
Thanks @Miradorn 💟

PR #882, opened 2025-04-18
Update CI

solnic:
This just bumps OSes and Elixir 1.13 to OTP 25.3. Previously used versions are no longer supported by GH and https://github.com/erlef/setup-beam

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,10 +33,11 @@ jobs:
             otp: '26.2'
             os: ubuntu-22.04
 
-          # Oldest supported Elixir/Erlang pair.
-          - elixir: '1.13.4-otp-22'
-            otp: '22.3.4'
-            os: ubuntu-20.04
+          # Oldest supported Elixir/Erlang pair that's tested;
+          # However, official support goes back to OTP 22.x
+          - elixir: '1.13'
+            otp: '25.3'
+            os: ubuntu-22.04
 
     steps:
     - name: Check out this repository
@@ -61,13 +62,13 @@ jobs:
     - name: Download Mix dependencies
       if: steps.mix-downloaded-deps-cache.outputs.cache-hit != 'true'
       run: |
-        if [ ${{ matrix.elixir }} != '1.13.4-otp-22' ]; then
+        if [ ${{ matrix.elixir }} != '1.13' ]; then
           mix deps.get --check-locked
         else
           mix deps.get
         fi
 
-        if [ ${{ matrix.elixir }} != '1.13.4-otp-22' ]; then
+        if [ ${{ matrix.elixir }} != '1.13' ]; then
           cd test_integrations/phoenix_app
 
           if [ ${{ matrix.lint }} == 'true' ]; then


whatyouhide:
LMK when this is ready for review

solnic:
> LMK when this is ready for review

@whatyouhide now 🙃

.github/workflows/main.yml
```diff
@@ -22,109 +22,109 @@ jobs:
         # https://hexdocs.pm/elixir/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp
         include:
           # Newest supported Elixir/Erlang pair.
-          - elixir: '1.18'
-            otp: '27.2'
+          - elixir: "1.18"
+            otp: "27.2"
```
whatyouhide, line 26: Was this change necessary?

.github/workflows/main.yml
```diff
@@ -22,109 +22,109 @@ jobs:
...
+          - elixir: "1.17"
+            otp: "26.2"
             os: ubuntu-22.04
 
           # Oldest supported Elixir/Erlang pair.
-          - elixir: '1.13.4-otp-22'
-            otp: '22.3.4'
-            os: ubuntu-20.04
+          - elixir: "1.13"
+            otp: "25.3"
```
whatyouhide, line 38: We need to leave a comment here saying that this is the lowest *tested* OTP version but technically we support back to OTP 22.

.github/workflows/main.yml
```diff
@@ -22,109 +22,109 @@ jobs:
...
+          path: |
+            deps
+            test_integrations/phoenix_app/deps
+          key: |
+            ${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-mix-deps-${{ hashFiles('**/mix.lock') }}
+
+      - name: Download Mix dependencies
+        if: steps.mix-downloaded-deps-cache.outputs.cache-hit != 'true'
+        run: |
+          if [ ${{ matrix.elixir }} != '1.13.4-otp-22' ]; then
```
whatyouhide, line 64: All these checks against `1.13.4-otp-22` are broken now, as there's no such `matrix.elixir` value?

.github/workflows/main.yml
```diff
@@ -22,109 +22,109 @@ jobs:
...
-        else
-          mix deps.get
-        fi
-
-        if [ ${{ matrix.elixir }} != '1.13.4-otp-22' ]; then
-          cd test_integrations/phoenix_app
-
-          if [ ${{ matrix.lint }} == 'true' ]; then
+      - name: Check out this repository
+        uses: actions/checkout@v4
```
whatyouhide, line 43: The indentation here is not a necessary change and makes the diff harder to review ☹️ Maybe we can split it up into a separate PR, because it's going to pollute the blame if someone needs to debug this...

whatyouhide (changes requested):
There are a ton of unnecessary changes here, left a couple more comments as well.

solnic:
@whatyouhide I made the changes minimalistic now. Previously my auto-formatter kicked in 🙃

whatyouhide (approved):

PR #884, opened 2025-04-28
Fix ignored Plug exception in the default filter

mbaleczny:
I noticed the `Plug.Parsers.RequestTooLarge` atom does not exists in the `Plug` repo.


whatyouhide:
Oh, great catch! 💟

whatyouhide (approved):

PR #885, opened 2025-05-02
Remain compatible with new translation format to be introduced in Elixir 1.19

martosaur:
Hi 👋 

Elixir 1.19 will [change](https://github.com/elixir-lang/elixir/pull/14380) how log translation works. Instead of replacing structured reports elixir will inject string translation into the report and wrap the `report_cb` callback. This enables logging libraries like Sentry to access report data directly instead of parsing iodata like they do now. But it also breaks the parsing. 

This PR introduces a bare minimum change to make Sentry work with the new translations – whenever we detect a report with `elixir_translation` key, we replace the report with the translation, effectively restoring pre-1.19 behaviour.

Feel free to close this PR and replace it with a more sophisticated solution that takes advantage of the new structure 🙏

### DIFF
--- a/test/sentry/logger_handler_test.exs
+++ b/test/sentry/logger_handler_test.exs
@@ -154,7 +154,7 @@ defmodule Sentry.LoggerHandlerTest do
       assert %KeyError{} = event.original_exception
       assert [exception] = event.exception
       assert exception.type == "KeyError"
-      assert exception.value == "key :foo not found in: []"
+      assert exception.value =~ ~r"key :foo not found in:\s+\[\]"
       assert Enum.find(exception.stacktrace.frames, &(&1.function =~ "Keyword.fetch!/2"))
     end
   end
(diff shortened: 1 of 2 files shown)


test/sentry/logger_handler_test.exs
```diff
@@ -154,7 +154,7 @@ defmodule Sentry.LoggerHandlerTest do
       assert %KeyError{} = event.original_exception
       assert [exception] = event.exception
       assert exception.type == "KeyError"
-      assert exception.value == "key :foo not found in: []"
+      assert exception.value == "key :foo not found in:\n\n    []"
```
martosaur, line 157: That's unrelated to the translation: https://github.com/elixir-lang/elixir/pull/14269

whatyouhide:
I think this is totally fine for the time being. Great catch, thank you @martosaur 💟

PR #887, opened 2025-05-15
Add Telemetry integration

whatyouhide:
I think this would be a really helpful addition!

[Corresponding docs PR](https://github.com/getsentry/sentry-docs/pull/13725).

### DIFF
--- a/lib/sentry/integrations/telemetry.ex
+++ b/lib/sentry/integrations/telemetry.ex
@@ -32,11 +32,12 @@ defmodule Sentry.Integrations.Telemetry do
         ) :: :ok
   def handle_event(@failure_event, _measurements, %{} = metadata, :no_config) do
     stacktrace = metadata[:stacktrace] || []
+    handler_id = stringified_handler_id(metadata[:handler_id])
 
     options = [
       stacktrace: stacktrace,
       tags: %{
-        telemetry_handler_id: stringified_handler_id(metadata[:handler_id]),
+        telemetry_handler_id: handler_id,
         event_name: inspect(metadata[:event_name])
       }
     ]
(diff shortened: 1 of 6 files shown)


lib/sentry/integrations/telemetry.ex
```diff
@@ -0,0 +1,65 @@
...
+      case {metadata[:kind], metadata[:reason]} do
+        {:error, reason} ->
+          exception = Exception.normalize(:error, reason, stacktrace)
+          Sentry.capture_exception(exception, options)
+
+        {kind, reason} ->
+          options =
+            Keyword.merge(options,
+              extra: %{kind: inspect(kind), reason: inspect(reason)},
+              interpolation_parameters: [inspect(metadata[:handler_id])]
```
solnic, line 54: This could reuse the already stringified handler_id because it gets set as `options.tasgs.telemetry_handler_id` above.

lib/sentry/integrations/telemetry.ex
```diff
@@ -0,0 +1,65 @@
...
+            )
+
+          Sentry.capture_message("Telemetry handler %s failed", options)
```
solnic, line 57: Shouldn't this be a warning or even an error?

solnic (approved):
This is indeed a very useful thing to have! Left a couple of comments, otherwise LGTM 🚀

lib/sentry/integrations/telemetry.ex
whatyouhide: This is already an error that gets reported to Sentry. Not sure what you mean here?

solnic: @whatyouhide oh I just assumed that `capture_message` doesn't send an error by default, and sends an info-level message instead. IIRC Sentry displays and manages these messages differently, especially when it comes to notifications.

whatyouhide: I think you're thinking about logging. These get displayed as errors.

solnic: @whatyouhide messages in Sentry do have levels too, and our SDK supports `level` option in `capture_message`. I just checked and luckily the default level is `error`.

One thing I noticed is that the UI displays "(no error message)" which may be a little confusing for the users.

image

PR #889, opened 2025-05-21
Fix regexes on Erlang/OTP 28

josevalim:
Regexes cannot be part of module bodies in Erlang/OTP 28, so this PR adjusts the repository accordingly.

A new release with this fix would be appreciated, since Erlang/OTP 28 has just been released. :)


grzuy:
Thanks @josevalim for the fix!

It still not possible to set this option now on `config/*.exs` module bodies to a list of regex, right?

PR #891, opened 2025-05-22
Fix for OTP 28

mayel:
see https://elixirforum.com/t/elixir-v1-18-4-released/70965

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -495,7 +495,7 @@ defmodule Sentry.Config do
 
   @valid_keys Keyword.keys(@raw_opts_schema)
 
-  defp opts_schema do: NimbleOptions.new!(@raw_opts_schema)
+  defp opts_schema, do: NimbleOptions.new!(@raw_opts_schema)
 
   @spec validate!() :: keyword()
   def validate! do


solnic:
Thank you for the PR - why do we need this change though? I thought only regexps no longer work as module attributes.

mayel:
I think functions in general are no longer supported, and regexes get compiled to functions?

Ah, actually the compile error makes it clear: `The supported values are: lists, tuples, maps, atoms, numbers, bitstrings, PIDs and remote functions in the format &Mod.fun/arity`

solnic:
> Ah, actually the compile error makes it clear: `The supported values are: lists, tuples, maps, atoms, numbers, bitstrings, PIDs and remote functions in the format &Mod.fun/arity`

@mayel are you saying this is the error that you get when you try to compile on OTP 28?

mayel:
Yeah the full error looked similar to the one here: https://github.com/jonasschmidt/ex_json_schema/issues/93

lib/sentry/config.ex
```diff
@@ -493,9 +493,10 @@ defmodule Sentry.Config do
                      hook_opts_schema
                    ])
 
-  @opts_schema NimbleOptions.new!(@raw_opts_schema)
   @valid_keys Keyword.keys(@raw_opts_schema)
 
+  defp opts_schema do: NimbleOptions.new!(@raw_opts_schema)
+
```
solnic, line 499: @mayel this is a syntax error

solnic (changes requested):

lib/sentry/config.ex
mayel: Oops sorry, did that quickly using github's editor. Can fix in a bit.

mayel: done

whatyouhide:
If this is not the same issue, then we should solve it like this: https://github.com/getsentry/sentry-elixir/pull/889

solnic:
> If this is not the same issue, then we should solve it like this: #889

@whatyouhide in #889 a regexp caused the issue, here it seems like having a custom `NimbleOptions` struct as a module attribute causes compilation errors too

whatyouhide:
It shouldn't. The issue is likely some value in that struct, which we should approach like we did for #889.

mayel:
Here's what I'm getting, using `mise` to install these versions:
```
erlang    28
elixir    1.18.4-otp-27
```


```
==> sentry
Compiling 41 files (.ex)

== Compilation error in file lib/sentry/config.ex ==
** (ArgumentError) cannot inject attribute @opts_schema into function/macro because cannot escape #Reference<0.3116512704.2745827333.161870>. The supported values are: lists, tuples, maps, atoms, numbers, bitstrings, PIDs and remote functions in the format &Mod.fun/arity
    (elixir 1.18.4) lib/kernel.ex:3777: Kernel.do_at_escape/2
    (elixir 1.18.4) expanding macro: Kernel.@/1
    (sentry 10.9.0) lib/sentry/config.ex:484: Sentry.Config.validate!/1
could not compile dependency :sentry, "mix compile" failed. Errors may have been logged above. You can recompile this dependency with "mix deps.compile sentry --force", update it with "mix deps.update sentry" or clean it with "mix deps.clean sentry"

**Comment by solnic** [comment]
:( I'm not sure what `#Reference<0.3116512704.2745827333.161870>` might be. @whatyouhide maybe you could shed some light?

**Comment by whatyouhide** [comment]
I wont have the time to look at this now. Suggestions:

1. Make sure nimble_options itself is fine (clone it and `mix test`)
2. Take away all the options in the schema and see if it compiles—if it does, bisect by commenting out halves until you find the option that breaks it

**Comment by wojtekmach** [comment]
Sentry compiles just fine here on OTP 28:

```
~$ iex
Erlang/OTP 28 [erts-16.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Interactive Elixir (1.18.4) - press Ctrl+C to exit (type h() ENTER for help)
iex> Mix.install([{:sentry, github: "getsentry/sentry-elixir"}, :hackney])
# (...)
:ok
```

**Comment by mayel** [comment]
@wojtekmach with the latest version from hex as well?

**Comment by wojtekmach** [comment]
The latest version on Hex fails for me:

```
== Compilation error in file lib/sentry/config.ex ==
** (ArgumentError) cannot inject attribute @opts_schema into function/macro because cannot escape #Reference<0.3761450473.4023517185.203366>. The supported values are: lists, tuples, maps, atoms, numbers, bitstrings, PIDs and remote functions in the format &Mod.fun/arity
    (elixir 1.19.0-dev) lib/kernel.ex:3796: Kernel.do_at_escape/2
    (elixir 1.19.0-dev) expanding macro: Kernel.@/1
    lib/sentry/config.ex:484: Sentry.Config.validate!/1
could not compile dependency :sentry, "mix compile" failed. Errors may have been logged above. You can recompile this dependency with "mix deps.compile sentry --force", update it with "mix deps.update sentry" or clean it with "mix deps.clean sentry"
** (exit) shutdown: 1
    (mix 1.19.0-dev) lib/mix/tasks/compile.all.ex:107: Mix.Tasks.Compile.All.compile/4
    (mix 1.19.0-dev) lib/mix/tasks/compile.all.ex:75: Mix.Tasks.Compile.All.do_run/2
    (mix 1.19.0-dev) lib/mix/task.ex:499: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.19.0-dev) lib/mix/tasks/compile.ex:141: Mix.Tasks.Compile.run/1
    (mix 1.19.0-dev) lib/mix/task.ex:499: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.19.0-dev) lib/mix/tasks/deps.compile.ex:216: anonymous fn/2 in Mix.Tasks.Deps.Compile.do_mix/2
    (mix 1.19.0-dev) lib/mix/project.ex:480: Mix.Project.in_project/4
```

but my understanding is it does not contain the regex fix: https://github.com/getsentry/sentry-elixir/pull/889.

solnic:
@wojtekmach could you try with latest stuff from master?

wojtekmach:
Sorry I wasn’t clear, what I meant was Sentry master compiles just fine on OTP 28 for me. In other words, I don’t think this PR is necessary.

mayel:
Ah yes it compiles with latest master, thanks and sorry for the noise! (noting though that I had to add `opentelemetry_semantic_conventions` to my deps for it to successfully compile)

PR #894, opened 2025-05-23
Bump Bandit version and fix incomplete test assertions

martosaur:
Just a bunch of very minor fixes I ran into while investigating something:
* Bumping Bandit also bumped Plug, which deprecated `Plug.Test`.
* Couple of tests related to duplicated reporting only asserted one event instead of 2


solnic (approved):
Thank you 😄

martosaur:
rebased 🫡

PR #899, opened 2025-05-27
Add Elixir main and OTP master to CI matrix

solnic:
#skip-changelog


whatyouhide:
I don't think this is worth the hassle—generally, we should likely strive to run tests when RCs come out so that we can address issues.

Issue #895, opened 2025-05-23
Limit role of `Sentry.PlugCapture`?


martosaur:
### Problem Statement

The existence of `Sentry.PlugCapture` has been bothering me for a while, so I made a deep dive. Why does Sentry need a special Plug to report errors happening in Plug specifically? 

The answer seems to be that back in the day those errors did not have a `crash_reason` metadata and thus Sentry's LoggerBackend [could not report them](https://github.com/getsentry/sentry-elixir/issues/353#issuecomment-510260828) off of a log event. When this was fixed in 2019 `Sentry.PlugCapture` was an established module and the solution to the double reporting problem was to ignore log messages with `:cowboy` domain 🙄 

However, there was and is actually a valid reason for keeping it! Cowboy executes plug in a separate process and that means that its own log event is missing Sentry context. `conn` could still be sourced from the metadata, but anything set by `Sentry.PlugContext` is gone.

Since then, Bandit released and became the default choice for Phoenix webserver. It received the same treatment as Cowboy: Bandit tags all errors with `:bandit` domain and Sentry [ignores them](https://github.com/getsentry/sentry-elixir/pull/739). However, Bandit runs code in the same process and does not suffer from the same problem as Cowboy. If you compare errors captured by `Sentry.PlugCapture` and handler/backend, they are pretty much identical.

### Solution Brainstorm

All of the above makes me believe that `Sentry.PlugCapture` can be soft-deprecated and only recommended for Cowboy applications. As a user, I expect Sentry errors to be the same I see in logs and `Sentry.PlugCapture` side steps logging. Most of the time, this isn't noticeable, but as soon as you start tinkering with logs you realize that what is logged is not what is reported to Sentry. I imagine laying this all out will also help with the upcoming structured logging feature.

whatyouhide:
I think this is a fantastic idea (and a great deep dive, thanks @martosaur!). Would you be open to draft a soft-deprecation docs update? We'd need to do it in both this repo and https://github.com/getsentry/sentry-docs.

martosaur:
I should have probably cc @mtrudel right away, but better late then never. Just a heads up, IIRC there is a class of Bandit errors that can happen outside of Plug and thus previously would had been both _not_ reported by `PlugCapture` and ignored by the logger handler. After this change goes live, they will surface to Sentry and there might be a wave of users who'll see them for the first time. That should be okay though, as Bandit's logging is a lot more robust and configurable now.

PR #900, opened 2025-05-27
Remove bandit from default excluded domains and recommend PlugCapture only for Cowboy

martosaur:
Closes https://github.com/getsentry/sentry-elixir/issues/895

See the issue for a write up. The PR here mostly updates the docs. The important change is removing `:bandit` from the list of excluded domains. Not sure if this counts as a breaking change or not. Users who use default `excluded_domains` with `PlugCapture` will see duplicated errors after upgrade 🤔 Let me know what you think!


lib/sentry/plug_capture.ex
```diff
@@ -1,11 +1,17 @@
...
 
   It is intended for usage with `Sentry.PlugContext`, which adds relevant request
   metadata to the Sentry context before errors are captured.
 
+  > #### Cowboy only {: .info}
+  >
+  > Note that `Sentry.PlugCapture` is only recommended for Cowboy applications.
+  For applications running on Bandit, which is the default in Phoenix,
+  `Sentry.PlugContext` should be enough, and using `Sentry.PlugCapture` might
+  result in duplicate errors.
```
whatyouhide, line 13: Nits!

```suggestion
  > #### Only for Cowboy {: .info}
  >
  > `Sentry.PlugCapture` is only recommended for Cowboy applications.
  > For applications running on Bandit, which is the most recent default webserver
  > in Phoenix, `Sentry.PlugContext` should be enough, and using `Sentry.PlugCapture`
  > might result in duplicate errors.
```

lib/sentry/plug_capture.ex
```diff
@@ -69,6 +75,7 @@ defmodule Sentry.PlugCapture do
       * scrubs sensitive body params just like `Sentry.PlugContext.default_body_scrubber/1`
 
   """
+  require Logger
```
whatyouhide, line 78: Why this change?

lib/sentry/plug_context.ex
```diff
@@ -12,9 +12,9 @@ defmodule Sentry.PlugContext do
 
       plug Sentry.PlugContext
 
-  However, this module is generally intended to be used with `Sentry.PlugCapture`:
-  this plug will add context metadata to the request, while `Sentry.PlugCapture` will
-  capture raised exceptions and errors and report them to Sentry with the added metadata.
+  This plug will add context metadata to the request, which will be added to
+  reported errors that happen during plug execution. Note that for Cowboy
+  applications, you will also need to use `Sentry.PlugCapture`.
```
whatyouhide, line 17: Nit for trimming some words:

```suggestion
  This plug will add context metadata to the request, which will be added to
  reported errors that happen during plug execution. For Cowboy
  applications, you will also need to use `Sentry.PlugCapture`.
```

whatyouhide (commented):
My gut feeling is that removing `:bandit` from the default is not really necessary. It's mostly going to be ignored, right?

martosaur:
> My gut feeling is that removing :bandit from the default is not really necessary. It's mostly going to be ignored, right?

Currently logger handler ignores them to let `PlugCapture` log them directly. If Bandit users remove `PlugCapture` from their apps, logger handler (or backend) has to step in!

Check out this test modification. If we both exclude `:bandit` and comment out `Sentry.PlugCapture` in `Sentry.ExamplePlugApplication` there's no path to log this error:

```elixir
    @tag handler_config: %{excluded_domains: [:bandit]}
    test "sends two errors when a Plug process crashes if PlugCapture is used and :bandit not excluded",
         %{sender_ref: ref} do
      start_supervised!({Sentry.ExamplePlugApplication, server: :bandit}, restart: :temporary)

      :hackney.get("http://127.0.0.1:8003/error_route", [], "", [])

      assert_receive {^ref, _event}, 1000 # This assertion fails
      # assert_receive {^ref, _event}, 1000
    end
```

whatyouhide (approved):
You are totally correct, sorry! 💯

whatyouhide:
Thank you @martosaur! Please tag me in there because I don't have notifications on for that.

PR #903, opened 2025-06-04
Better span sampler

solnic:
This updates OpenTelemetry.Sampler so that it uses `traces_sample_rate` for sampling spans before they get sent to the span processor.

This way we're not processing spans when they are
dropped during sampling, which was previously the
case as the Client was responsible for sampling
just before attempting to send a transaction.

This is based on:

- https://opentelemetry.io/docs/concepts/sampling/
- https://github.com/getsentry/sentry-python/blob/07ba0f0f46a6729251bd94984ad203befa4737b0/sentry_sdk%2Fopentelemetry%2Fsampler.py

#skip-changelog

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -622,7 +622,7 @@ defmodule Sentry.Config do
   @spec sample_rate() :: float()
   def sample_rate, do: fetch!(:sample_rate)
 
-  @spec traces_sample_rate() :: float()
+  @spec traces_sample_rate() :: nil | float()
   def traces_sample_rate, do: fetch!(:traces_sample_rate)
 
   @spec hackney_opts() :: keyword()


lib/sentry/opentelemetry/sampler.ex
```diff
@@ -2,30 +2,138 @@ if Code.ensure_loaded?(:otel_sampler) do
...
+          {:record_and_sample, [], tracestate}
+
+        true ->
+          random_value = :rand.uniform()
+          sampled = random_value < sample_rate
+
+          tracestate = build_tracestate(sample_rate, random_value, sampled)
+          decision = if sampled, do: :record_and_sample, else: :drop
+          {decision, [], tracestate}
+      end
```
whatyouhide, line 100: Refactoring suggestion:

```elixir
{sampled, random_value} =
  cond do
    sample_rate == 0.0 -> {false, 1.0}
    sample_rate == 1.0 -> {true, 0.0}
    true -> random_value = :rand.uniform(); {random_value < sample_rate, random_value}
  end

tracestate = build_tracestate(sample_rate, random_value, sampled)
decision = if sampled, do: :record_and_sample, else: :drop
{decision, [], tracestate}
```

whatyouhide (approved):

sl0thentr0py (changes requested):
just one comment about the client report, rest we can postpone when we implement propagation.

If you want, you can already set the default `traces_sample_rate` to `nil` and tweak the logic a bit to account for that

lib/sentry/opentelemetry/sampler.ex
```diff
@@ -2,30 +2,125 @@ if Code.ensure_loaded?(:otel_sampler) do
...
+              {:inherit, false, tracestate}
+
+            nil ->
+              :no_trace
+          end
+      end
+    end
+
+    defp make_sampling_decision(sample_rate) do
+      {sampled, random_value} =
```
sl0thentr0py, line 84: this needs a `nil` branch right?

PR #911, opened 2025-06-11
Don't use local function capture for Telemetry handler

RudolfMan:
When the integration enabled on app start it throws the message:

> The function passed as a handler with ID "Sentry.Integrations.Telemetry-telemetry-failures" is a local function.
This means that it is either an anonymous function or a capture of a function without a module specified. That may cause a performance penalty when calling that handler. For more details see the note in `telemetry:attach/4` documentation.


whatyouhide (approved):
Ooops, good catch!

PR #922, opened 2025-07-09
Improve Event callback docs

brain-geek:
Changing a few things there:
- typo in function name (`after_send` is not a function)
- Adding a note about how to stop Sentry from reporting event. `Filtering Exceptions` section says: 

> If you would like to prevent Sentry from sending certain exceptions, you can use the `:before_send` configuration option. See the [*Event Callbacks* section](#module-event-callbacks) below.

But in the Event Callbacks section it does not say a word on how to filter those.

- Adding a type spec for a bit easier investigation into how to test it. I had to look into the code of this actual repo to understand this and previous point. Now it should be a bit more obvious.

#skip-changelog

### DIFF
--- a/lib/sentry.ex
+++ b/lib/sentry.ex
@@ -124,16 +124,18 @@ defmodule Sentry do
 
       config :sentry,
         before_send: {MyModule, :before_send},
-        after_send_event: {MyModule, :after_send}
+        after_send_event: {MyModule, :after_send_event}
 
   `MyModule` could look like this:
 
       defmodule MyModule do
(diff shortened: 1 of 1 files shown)


lib/sentry.ex
```diff
@@ -124,16 +124,18 @@ defmodule Sentry do
 
       config :sentry,
         before_send: {MyModule, :before_send},
-        after_send_event: {MyModule, :after_send}
+        after_send_event: {MyModule, :after_send_event}
```
whatyouhide, line 127: Can we instead change the function name in the code example to be `:after_send`?

PR #923, opened 2025-07-12
traceparent support

spunkedy:
when the traceparent header is passed on the requests to phoenix, it is currently not supported to be able to pull in the parent span so that we can support having distributed tracing until the full otel support is implemented

### DIFF
--- a/lib/sentry/opentelemetry/span_processor.ex
+++ b/lib/sentry/opentelemetry/span_processor.ex
@@ -15,19 +15,22 @@ if Code.ensure_loaded?(OpenTelemetry) do
     alias Sentry.{Transaction, OpenTelemetry.SpanStorage, OpenTelemetry.SpanRecord}
     alias Sentry.Interfaces.Span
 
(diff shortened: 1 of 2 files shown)


spunkedy:
```
      {:sentry, git: "https://github.com/spunkedy/sentry-elixir", branch: "feature-enable-traceparent"},
```

has been tested with passing the headers

solnic:
Thank you for the PR! 💜 We actually need to introduce an opentelemetry propagator as we have an architectural design for this functionality that is followed by other SDKs and we need the Elixir SDK to follow it as well 🙂

If this is something you'd be interested in doing - you could port what we do in the Ruby SDK: https://github.com/getsentry/sentry-ruby/blob/master/sentry-opentelemetry/lib/sentry/opentelemetry/propagator.rb

venkatd:
There's an older implementation, but I tested it and it doesn't work with the latest sentry:
https://github.com/scripbox/opentelemetry_sentry/blob/v0.1.1/lib/opentelemetry_sentry/propagator.ex

I believe the main issue is that baggage is not being propagated

PR #931, opened 2025-08-01
Relax version reqs on opentelemetry deps

### LINKED ISSUE #928: Cannot upgrade to Sentry 11.0 (w/o tracing) because of OTel dependencies
### Description

We cannot upgrade Sentry to 11.0, even though we **do not** use tracing, because it requires `opentelemetry_semantic_conventions` 1.x but we're on 0.2.0. This is quite annoying as there's no reason for us to not be able to update given we don't use tracing. This is one of the biggest reasons for warranting a separate package (`sentry_tracing`). cc @solnic

solnic:
This enables installing `sentry` in apps with old
opentelemetry libs, that are not compatible with
tracing, in which case tracing is simply not
loaded and cannot be used as a feature.

Closes #928


sl0thentr0py (approved):
ty! :shipit:

PR #950, opened 2025-10-28
feat(oban): add option to include job tags into reported exception tags

adriencohen:
### Description

Added an `oban_tags` option to the `oban` configuration to include a job's tags in the tags of the Sentry issue created. The goal is to have ownership rules based on the tags of an Oban job.

This is likely not ready to merge as is at least more thought needs to be put into the configuration option. Should we consider changing `capture_errors` into a keyword list, similar to `cron`? That would be a breaking change.

I'm opening this PR mainly to discuss whether you’d consider such a feature.

Thank you!


adriencohen:
Definitely makes sense. I’ve committed the changes, and I think we’re getting closer to something that could be released. Let me know what you think

whatyouhide (approved):

PR #969, opened 2026-01-02
feat: support for Structured Logs

### LINKED ISSUE #906: Sentry logger handler for Structured Logging
We already have `Sentry.LoggerHandler` but it's used to send regular events, not logs. That's why we need to add a new handler dedicated to sending log events.

The new handler should be enabled only when `config.enabled_logs` is set to `true`.

### LINKED ISSUE #907: Support for default attributes in Structured Logging
### Problem Statement

TBD

### Solution Brainstorm

TBD

### LINKED ISSUE #908: Support for parameterized log payloads
### Problem Statement

TBD

### Solution Brainstorm

TBD

solnic:
Initial work on support for Logs.

This follows requirements described in https://develop.sentry.dev/sdk/telemetry/logs/

---

Closes #906 
Closes #907 
Closes #908

### DIFF
--- a/lib/sentry/log_event_buffer.ex
+++ b/lib/sentry/log_event_buffer.ex
@@ -132,6 +132,7 @@ defmodule Sentry.LogEventBuffer do
         # Flush immediately if we've reached max_events
         send_events(events)
         cancel_timer(state.timer_ref)
+        flush_stale_timeout_message()
         {:noreply, %{state | events: [], count: 0, timer_ref: schedule_flush()}}
       else
         {:noreply, %{state | events: events, count: new_count}}
@@ -188,7 +189,9 @@ defmodule Sentry.LogEventBuffer do
   # This can happen if the timer fires while we're in the middle of a flush.
   defp flush_stale_timeout_message do
     receive do
-      :flush_timeout -> :ok
+      :flush_timeout ->
+        log_debug("Flushed stale timeout message")
+        :ok
     after
       0 -> :ok
     end


lib/sentry/client.ex
```diff
@@ -134,6 +135,36 @@ defmodule Sentry.Client do
...
+
+  Log events are sent asynchronously and do not support callbacks or sampling.
+  They are buffered and sent in batches according to the Sentry Logs Protocol.
+
+  Returns `{:ok, envelope_id}` on success or `{:error, reason}` on failure.
+  """
+  @doc since: "12.0.0"
+  @spec send_log_events([LogEvent.t()]) ::
+          {:ok, envelope_id :: String.t()} | {:error, ClientError.t()}
+  def send_log_events([]), do: {:ok, ""}
```
whatyouhide: If we have a "log batch" struct, would it make sense to have `send_log_batch` instead of this?

lib/sentry/envelope.ex
```diff
@@ -8,14 +8,22 @@ defmodule Sentry.Envelope do
...
   @type t() :: %__MODULE__{
           event_id: UUID.t(),
           items: [
-            Attachment.t() | CheckIn.t() | ClientReport.t() | Event.t() | Transaction.t(),
+            Attachment.t()
+            | CheckIn.t()
+            | ClientReport.t()
+            | Event.t()
+            | LogBatch.t()
+            | LogEvent.t()
```
whatyouhide: When do we have an envelope with log events in it instead of a log batch?

lib/sentry/log_batch.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Sentry.LogBatch do
+  @moduledoc """
+  A batch of log events to be sent in a single envelope item.
+
+  According to the Sentry Logs Protocol, log events are sent in batches
+  within a single envelope item with content_type "application/vnd.sentry.items.log+json".
```
whatyouhide: Same nit as above:


```suggestion
  within a single envelope item with content type `application/vnd.sentry.items.log+json`.
```

lib/sentry/log_event.ex
```diff
@@ -0,0 +1,326 @@
...
+      iex> log_event = %{level: :info, msg: {:string, "Hello %s"}, meta: %{}}
+      iex> Sentry.LogEvent.from_logger_event(log_event, %{}, ["Jane"])
+      %Sentry.LogEvent{body: "Hello Jane", template: "Hello %s", parameters: ["Jane"], ...}
+
+      # With named parameters
+      iex> log_event = %{level: :info, msg: {:string, "Hello %{name}"}, meta: %{}}
+      iex> Sentry.LogEvent.from_logger_event(log_event, %{}, %{name: "Jane"})
+      %Sentry.LogEvent{body: "Hello Jane", template: "Hello %{name}", parameters: ["Jane"], ...}
+  """
+  @spec from_logger_event(map(), map(), list() | map() | nil) :: t()
```
whatyouhide: We can be more precise?

```suggestion
  @spec from_logger_event(:logger.log_event(), map(), [String.t()] | %{optional(String.t()) => String.t()} | nil) :: t()
```

lib/sentry/log_event_buffer.ex
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+  ## Public API
+
+  @doc """
+  Starts the log event buffer process.
+  """
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
```
whatyouhide: Cannot be tested this way.

lib/sentry/log_event_buffer.ex
```diff
@@ -0,0 +1,206 @@
...
+      timer_ref: schedule_flush()
+    }
+
+    {:ok, state}
+  end
+
+  @impl GenServer
+  def handle_cast({:add_event, event}, state) do
+    # Check if queue is at max capacity
+    if length(state.events) >= @max_queue_size do
```
whatyouhide: This is going to be called constantly. Can we keep track of the current number of events instead of recomputing `length(state.events)` every time?

lib/sentry/log_event_buffer.ex
```diff
@@ -0,0 +1,206 @@
...
+      else
+        {:noreply, %{state | events: events}}
+      end
+    end
+  end
+
+  @impl GenServer
+  def handle_call(:flush, _from, state) do
+    send_events(state.events)
+    cancel_timer(state.timer_ref)
```
whatyouhide: There's an (unlikely) race condition here where the timer fired while we're in this function and `cancel_timer/1` has no effect, leading to the `:flush` message being in the message queue and handled right after flushing here.

Two solutions:

  * Switch to `gen_statem`, which makes all this pretty easy with its timeout handling.
  * Flush the `:flush_timeout` message by using a `receive do :flush_timeout -> :ok after 0 -> :ok end` or something.

lib/sentry/log_event.ex
```diff
@@ -0,0 +1,326 @@
...
+
+  @type level :: :trace | :debug | :info | :warn | :error | :fatal
+
+  @typedoc """
+  A log event struct.
+  """
+  @type t :: %__MODULE__{
+          level: level(),
+          body: String.t(),
+          timestamp: float(),
```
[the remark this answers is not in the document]
solnic: Yeah it's a float see https://develop.sentry.dev/sdk/telemetry/logs/#log-envelope-item-payload

lib/sentry/logs_handler.ex
```diff
@@ -0,0 +1,247 @@
+defmodule Sentry.LogsHandler do
```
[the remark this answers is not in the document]
whatyouhide: Yeah this is a tough one. Renaming `Sentry.LoggerHandler` is annoying because it will require users who upgrade to do it manually, and there's no "nice" way to deprecate the name itself. I think keeping only `Sentry.LoggerHandler` and making it configurable is messier on the SDK side but might be a better experience for the end user, provided it:

* validates configuration as much as possible automatically so that it's easy to make sure you're doing the right config
* supports the new Structured Logs approach as **opt in**—for example, we use the logger handler for catching log errors and crashes but we do not use the Sentry Logs product at all (and wouldn't want to)

Thoughts?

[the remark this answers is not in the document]
solnic: Doing extra work on SDK side to improve DX is our principle guideline thus collapsing into one `LoggerHandler` sounds like the way to go even if it means more work for us.

sl0thentr0py (changes requested):
2 comments about schema changes

PR #973, opened 2026-01-06
Support for filtering logs via `before_send_log`

### LINKED ISSUE #909: Support for filtering logs via `before_send_log`
### Problem Statement

In order to be able to filter out certain log events using arbitrary, user-provided logic, we need to support `before_send_log` configuration option as described here https://develop.sentry.dev/sdk/telemetry/logs/#initialization-options

### Solution Brainstorm

TBD

solnic:
I'll update changelog for the entire feature when preparing changelog for `12.0.0`.

#skip-changelog

Closes #909

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -160,7 +160,7 @@ defmodule Sentry.Client do
       :not_collecting ->
         log_events
         |> run_before_send_log_callbacks()
-        |> do_send_log_events()
+        |> send_log_events()
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/sentry/client.ex
```diff
@@ -154,17 +158,49 @@ defmodule Sentry.Client do
...
 
+  defp call_before_send_log(log_event, function) when is_function(function, 1) do
+    function.(log_event)
+  end
+
+  defp call_before_send_log(log_event, {mod, fun}) do
+    apply(mod, fun, [log_event])
+  end
+
+  defp do_send_log_events([]), do: {:ok, ""}
```
whatyouhide: `do_` has essentially disappeared from the conventions of the Elixir community. Can you rename this to `send_log_events`?

PR #999, opened 2026-02-25
fix: drain rate-limited items from buffer and record client reports

solnic:
When a rate limit is active, the scheduler now drains buffered items and discards them instead of retaining them for later retry. Client reports with :ratelimit_backoff reason are recorded for each discarded item so Sentry knows data was dropped.

This aligns with the Sentry SDK spec which states that rate-limited data must be discarded during the rate limit period, and client reports must be recorded when items are dropped.

Ref: https://github.com/getsentry/sentry-elixir/pull/996#discussion_r2847458121

---

#skip-changelog

### DIFF
--- a/test/sentry/telemetry_processor_integration_test.exs
+++ b/test/sentry/telemetry_processor_integration_test.exs
@@ -342,7 +342,7 @@ defmodule Sentry.TelemetryProcessorIntegrationTest do
       :ok
     end
 
-    test "rate-limited HTTP response causes subsequent items to be drained and client report sent",
+    test "rate-limited HTTP response causes subsequent events to be dropped with client report",
          ctx do
       test_pid = self()
       ref = make_ref()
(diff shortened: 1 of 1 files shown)


solnic:
> One next level that you could still implement, but it's totally optional, is: before you add items to the buffer, you could also check if they're rate limited, and then you could already drop them and again record client reports. I would do this in a follow-up PR. This is the first requirement here: https://develop.sentry.dev/sdk/foundations/processing/telemetry-processor/#telemetry-buffer:

@philipphofmann I already did that on Friday :) See:

```elixir
  defp add_to_buffer(processor, category, item) when is_atom(processor) do
    data_category = Category.data_category(category)

    if RateLimiter.rate_limited?(data_category) do
      {:ok, {:rate_limited, data_category}}
    else
      Buffer.add(buffer_name(processor, category), item)
      Scheduler.signal(scheduler_name(processor))
      :ok
    end
  end
```

sl0thentr0py (approved):

PR #1030, opened 2026-04-02
refactor(test): rework Sentry.Test to use Bypass-based HTTP testing

solnic:
Replaces the in-memory event interception approach in `Sentry.Test` with real HTTP-level testing via Bypass. Events now flow through the full SDK pipeline—encoding, envelope construction, and HTTP transport—before being captured by a local Bypass server and decoded back into structs for assertions.

### Motivation

The previous `Sentry.Test` implementation used `NimbleOwnership` and an in-process collection mechanism that short-circuited the `Client.encode_and_send/4` path. This meant test assertions ran against events that never passed through the actual transport layer, masking potential serialization or envelope-packaging bugs.

Furthermore, there were various issues with config isolation causing test flakiness.

### Key changes

**`Sentry.Test` rewrite** — The module now spins up a per-test Bypass instance via `setup_sentry/1`, wires `before_send` / `before_send_log` callbacks to capture event structs, and points the DSN at the local endpoint. Collection state is stored in an isolated ETS table per test process.

**`Sentry.Test.Registry` (new)** — A GenServer started in the OTP supervision tree when `test_mode: true`. It manages the ETS table for collectors and boots a global default Bypass instance so that tests *without* `setup_sentry/1` still get a valid DSN (preserving the old `{:ok, ""}` return value).

**`Sentry.Client` simplified** — All `Sentry.Test.maybe_collect/1` branches removed from `encode_and_send/4`. Events always go through the normal transport path; test capture happens at the HTTP boundary instead.

**`nimble_ownership` dependency removed** — No longer needed since collection is ETS-based rather than ownership-based. ETS is simpler and avoids the complexity of ownership transfer/allowance bookkeeping across process boundaries; since events are now captured at the HTTP layer by Bypass (which already handles multi-process requests), a shared ETS table keyed by test PID provides natural isolation without extra machinery.

**Backward compatibility** — `start_collecting_sentry_reports/0`, `pop_sentry_reports/0`, `pop_sentry_transactions/0`, and `pop_sentry_logs/0` continue to work. The `owner_pid` parameter on these functions is deprecated (no longer meaningful with ETS-based collection). A dedicated `test_backward_compat_test.exs` validates the legacy workflow.

### Follow-up

A separate PR will be opened later on with convenient helpers and improved tests that cover more ground giving us more confidence with the full stack exercised during tests.

It will also make it much nicer to use Sentry test mode in user's test suites 🎉


lib/sentry.ex
```diff
@@ -380,10 +380,6 @@ defmodule Sentry do
         ClientReport.Sender.record_discarded_events(:event_processor, [event])
         :ignored
 
-      # If we're in test mode, let's send the event down the pipeline anyway.
-      Config.test_mode?() ->
-        Client.send_event(event, options)
-
       !Config.dsn() ->
         # We still validate options even if we're not sending the event. This aims at catching
         # configuration issues during development instead of only when deploying to production.
```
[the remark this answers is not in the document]
solnic: It's actually correct now since the system respects real logic that would be used in prod environments and we can test it properly.

lib/sentry/test.ex
```diff
@@ -417,38 +237,15 @@ defmodule Sentry.Test do
...
-          nil -> {:not_collecting, []}
-          events when is_list(events) -> {events, []}
-        end)
-      catch
-        :exit, {:noproc, _} ->
-          raise ArgumentError, "not collecting reported events from #{inspect(owner_pid)}"
-      end
-
-    case result do
-      {:ok, :not_collecting} ->
```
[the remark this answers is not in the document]
solnic: This is used only in tests so this would be a truly paranoid defensives, not worth the complexity.

sl0thentr0py (approved):

lib/sentry/logger_handler.ex
```diff
@@ -335,9 +335,15 @@ defmodule Sentry.LoggerHandler do
...
-    # Dispatch to all configured backends
+    # Dispatch to all configured backends.
+    # Rescue any exceptions to prevent OTP from removing the handler entirely
+    # when a single event fails to process.
     Enum.each(backends, fn backend ->
-      backend.handle_event(log_event, config, handler_id)
+      try do
+        backend.handle_event(log_event, config, handler_id)
+      rescue
+        _ -> :ok
```
sl0thentr0py, line 345: so this will swallow internal problems completely silently?

solnic: @sl0thentr0py so, sometimes logger handler would crash when running tests, and I thought we don't want that given the general contract of making sure our tooling doesn't crash people's code, but this is for sure out of scope here, so I reverted it, and taking a note instead.

sl0thentr0py: we should not crash people's code indeed, but if there's an underlying problem with the handler we need to fix that first.

whatyouhide:
This change makes me really sad.

`Sentry.Test` is **public API**, that we offer to users to make their testing experience better, faster, and concurrency safe. This change is likely going to have performance implications on people's test suites.

> Furthermore, there were various issues with config isolation causing test flakiness.

This is very handwavey—various issues like what?

> It will also make it much nicer to use Sentry test mode in user's test suites 🎉

I have no idea what this means. Why and how will this make it nicer?

solnic:
> This change makes me really sad.

Oh :( Sorry to hear that, I'll do my best to explain:
 
> `Sentry.Test` is **public API**, that we offer to users to make their testing experience better, faster, and concurrency safe.

This is exactly why this was done! You should be able to simply use `setup_sentry` in tests and use the new set of convenient helpers to easily assert on events/transactions/logs/check_ins. It also simplified production code so that there are no test-specific code paths.

> This change is likely going to have performance implications on people's test suites.

I haven't noticed any performance regressions in my testing. We'll also make it much simpler in 14.0 with a better transport-based architecture like we have in the Ruby SDK. This will allow more flexibility and simplify state isolation in tests.

I'll keep performance in mind too when working on these improvements.
 
> > Furthermore, there were various issues with config isolation causing test flakiness.
> 
> This is very handwavey—various issues like what?

Global configs were being changed during test executions causing flakiness. Now the test suite passes with no flaky errors, except some super rare cases where bypass crashes.
 
> > It will also make it much nicer to use Sentry test mode in user's test suites 🎉
> 
> I have no idea what this means. Why and how will this make it nicer?

This is all documented over here https://docs.sentry.io/platforms/elixir/testing/

---

I should also mention that this is an intermediate state before switching fully to telemetry processor and the transport abstraction - both refactorings will greatly simplify `Sentry.Test` 😄

PR #1031, opened 2026-04-02
refactor(tests): proper config isolation

solnic:
Introduces `Sentry.Test.Config` — a new module that provides per-test configuration isolation using `:persistent_term` keyed by test process PID, replacing the previous process-dictionary-based approach.

## Problem

The old `put_test_config/1` stored overrides in the process dictionary and duplicated them into `:persistent_term` (overwriting global config). This was fragile: spawned processes (GenServers, Tasks) couldn't reliably see test-specific config, and concurrent async tests could interfere with each other through shared `:persistent_term` keys.

## Solution

A pluggable `:namespace` resolver pattern in `Sentry.Config`:

- **`Sentry.Config`** gains a `:namespace` config option — a `{module, function}` tuple that resolves scoped config overrides. The default resolver always returns `:default` (use global config).
- **`Sentry.Test.Config`** is automatically activated when `test_mode: true` is set. It stores per-test overrides in `:persistent_term` as `{:sentry_config, test_pid, key}` and resolves them by walking the `$callers` chain, `allow/2` mappings, or (for `async: false` tests) active scope fallback.
- **`put_test_config/1`** in both test helper modules now delegates to `Sentry.Test.Config.put/1`, eliminating duplication.

### DIFF
--- a/lib/sentry/application.ex
+++ b/lib/sentry/application.ex
@@ -11,6 +11,7 @@ defmodule Sentry.Application do
   def start(_type, _opts) do
     config = Config.validate!()
     :ok = Config.persist(config)
+    :ok = Sentry.Test.Config.maybe_activate()
 
     Config.put_config(
       :in_app_module_allow_list,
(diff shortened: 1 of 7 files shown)


dingsdax (approved):

whatyouhide:
This is likely going to have performance implications; `erase` and `put` are *very* expensive operations. I assume you benchmarked this?

PR #1054, opened 2026-05-12
chore: Allow usage of hackney 4.x

sax:
### Description

Hackney has update several major versions in 2026, including security fixes. Looking at the hackney adapter, none of their breaking changes effect the sentry code.

### DIFF
--- a/lib/mix/tasks/sentry.install.ex
+++ b/lib/mix/tasks/sentry.install.ex
@@ -43,7 +43,7 @@ if Code.ensure_loaded?(Igniter) do
     def info(_argv, _composing_task) do
       %Igniter.Mix.Task.Info{
         group: :sentry,
-        adds_deps: [{:jason, "~> 1.2"}, {:hackney, "~> 1.8"}],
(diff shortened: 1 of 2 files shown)


sax:
Hmm the CI failure seems to be a setup issue with actions/create-github-app-token.

sax:
Ok, looks like there is a transitive dependency conflict. `:hackney` 4.0.0 split out its HTTP/2 and HTTP3 stacks to an [erlang_h2](https://github.com/benoitc/erlang_h2) library. Meanwhile, `:opentelemetry_exporter` depends on `:grpcbox`, which depends on `:chatterbox`... which implements a set of conflicting modules including `:h2_connection`. https://github.com/joedevivo/chatterbox/issues/154

So technically the `Sentry.HackneyClient` is fine, but in practice one still can't update hackney to 4.0 until other libraries sort out the conflicts.

grzuy:
For what is worth, note that about 10 hackney CVEs were made public today (https://github.com/benoitc/hackney/security) and 4.x seems to be the only supported version for security updates.

whatyouhide (approved):

PR #1062, opened 2026-05-17
fix(live_view_hook): handle Bandit.TransportError and URI struct in connect_info

### LINKED ISSUE #1025: LiveViewHook.on_mount crashes with Bandit.TransportError when client disconnects during mount
## Description

`Sentry.LiveViewHook.on_mount/4` crashes with `Bandit.TransportError: Unable to obtain peer_data` when a client disconnects before the LiveView mount completes.

## Root Cause

In `lib/sentry/live_view_hook.ex`, the `get_connect_info(socket, :peer_data)` call on line 82 is not wrapped in a try/rescue. When using Bandit as the web server, if the underlying WebSocket transport is already closed at the time of the call, Bandit raises `Bandit.TransportError`.

The existing `catch` block in `get_safe_ip_address/1` only protects the `:inet.ntoa` call — it does not protect `get_connect_info`.

## Reproduction

This occurs in production with mobile clients (e.g. Mobile Safari on iOS) that disconnect during LiveView mount — a common scenario on flaky mobile connections. Using Bandit as the HTTP server.

Observed on sentry 10.9.0, but the relevant code is unchanged through 12.0.3.

## Suggested Fix

Wrap the `get_connect_info(socket, :peer_data)` call in a try/rescue that catches `Bandit.TransportError` and returns `nil`, allowing the hook to continue without peer data rather than crashing.

## Stack Trace

```
Sentry.LiveView.on_mount hook errored out: ** (Bandit.TransportError) Unable to obtain peer_data
```

## Environment

- sentry: 10.9.0 (code unchanged through 12.0.3)
- bandit: latest
- phoenix_live_view: ~> 1.0
- elixir: 1.19.5 / OTP 28

### LINKED ISSUE #1040: LiveView Hook silently swallows errors
### Issue Description

I was having an issue with all errors related to LiveViews not showing up in sentry. After some debugging in a handle_event call running this:

```elixir
Sentry.capture_message("Manual sync test", result: :sync) |> IO.inspect()
```

I got:

```
2026-04-20 19:02:00.544 [warning] Failed to send Sentry event. Sentry failed to report event: the Sentry SDK could not encode the event to JSON: protocol JSON.Encoder not implemented for URI (a struct), the protocol must be explicitly implemented.

If you have a struct, you can derive the implementation specifying which fields should be encoded to JSON:

    @derive {JSON.Encoder, only: [....]}
    defstruct ...

It is also possible to encode all fields, although this should be used carefully to avoid accidentally leaking private information when new fields are added:

    @derive JSON.Encoder
    defstruct ...

Finally, if you don't own the struct you want to encode to JSON, you may use Protocol.derive/3 placed outside of any module:

    Protocol.derive(JSON.Encoder, NameOfTheStruct, only: [...])
    Protocol.derive(JSON.Encoder, NameOfTheStruct). This protocol is implemented for: Atom, BitString, Date, DateTime, Decimal, Duration, Ecto.Association.NotLoaded, Ecto.Schema.Metadata, Float, Integer, JSV.ValidationError, List, Map, NaiveDateTime, Oban.Met.Values.Gauge, Oban.Met.Values.Sketch, Oban.Pro.Producer, Oban.Pro.Producer.Meta.GlobalLimit, Oban.Pro.Producer.Meta.GlobalLimit.Partition, Oban.Pro.Producer.Meta.RateLimit, Oban.Pro.Producer.Meta.RateLimit.Partition, Time

Got value:

    %URI{
      scheme: "http",
      userinfo: nil,
      host: "localhost",
      port: 4001,
      path: nil,
      query: nil,
      fragment: nil
    }
```

After digging around in the actual source code for the live view hook, I found [this recommendation](https://hexdocs.pm/sentry/Sentry.LiveViewHook.html):

> To make this module work best, you'll need to fetch information from the LiveView's WebSocket. You can do that when calling the socket/3 macro in your Phoenix endpoint. For example:

```
socket "/live", Phoenix.LiveView.Socket,
        websocket: [connect_info: [:peer_data, :uri, :user_agent]]
```

However, it actually doesn't work at all without including `uri` there. Brand new phoenix apps do not include these keys by default, so it seems the documentation should be clearer about this and say "This module requires that you add..." instead of "To make this module work best".

### Reproduction Steps

Set up a brand new Phoenix app and follow the exact instructions at https://hexdocs.pm/sentry/setup-with-plug-and-phoenix.html

Then set up a live view with a `phx-click` event that triggers a server error.

### Expected Behavior

Server errors in a LiveView handler should be reported to Sentry, or at the very least the console/log output should show an error.

### Actual Behavior

Errors in LiveViews quietly fail without any log output indicating so.

### Elixir Version

1.19.5

### SDK Version

12.0.3

### Integration and Its Version

_No response_

### Sentry Config

_No response_

dl-alexandre:
## Summary

Fixes two related issues in `Sentry.LiveViewHook` that affect Bandit + LiveView production deployments:

1. **#1025** — `on_mount` crashes with `Bandit.TransportError` when the client disconnects during mount.
2. **#1040** — LiveView hooks silently swallow errors because `%URI{}` structs in `connect_info` fail JSON encoding.

### `lib/sentry/live_view_hook.ex`

- **Bandit disconnect safety**: Wrap `get_connect_info(socket, key)` in `get_connect_info_if_root/2` with a `try/rescue` that returns `nil` on any transport error. This mirrors the existing defensive pattern in `Sentry.PlugContext` (see `remote_port/1`).
- **URI struct normalization**: Convert `socket.host_uri` to a string with `URI.to_string/1` before storing it in Sentry request context. Previously, when `:uri` was omitted from `connect_info`, the raw `%URI{}` struct remained in context and caused JSON encoding to fail later in the pipeline.

### `lib/sentry/client.ex`

- **Defense-in-depth**: Run `sanitize_non_jsonable_values/2` on the `:request` field during event rendering, consistent with how `:extra`, `:user`, and `:tags` are already sanitized.

### Tests

- Added `SentryTest.EndpointNoUri` and `Sentry.LiveViewHookNoUriTest` to verify hook behavior when `:uri` is absent from `connect_info`.
- Added regression test for `Client.render_event/1` sanitizing non-JSON-encodable values in `:request`.

## Verification

- `mix test` — 813 tests, 0 failures
- `mix format --check-formatted` — clean
- `mix compile --warnings-as-errors` — clean

## Related Issues

- Closes #1025
- Closes #1040

### DIFF
--- a/lib/sentry/client.ex
+++ b/lib/sentry/client.ex
@@ -298,7 +298,12 @@ defmodule Sentry.Client do
     |> update_if_present(:breadcrumbs, fn bcs -> Enum.map(bcs, &Map.from_struct/1) end)
     |> update_if_present(:sdk, &Map.from_struct/1)
     |> update_if_present(:message, &sanitize_message(&1, json_library))
-    |> update_if_present(:request, &(&1 |> Map.from_struct() |> remove_nils()))
+    |> update_if_present(:request, fn req ->
+      req
+      |> Map.from_struct()
+      |> remove_nils()
+      |> sanitize_non_jsonable_values(json_library)
+    end)
     |> update_if_present(:extra, &sanitize_non_jsonable_values(&1, json_library))
     |> update_if_present(:user, &sanitize_non_jsonable_values(&1, json_library))
     |> update_if_present(:tags, &sanitize_non_jsonable_values(&1, json_library))
--- a/lib/sentry/live_view_hook.ex
+++ b/lib/sentry/live_view_hook.ex
@@ -156,7 +156,10 @@ if Code.ensure_loaded?(Phoenix.LiveView) do
 
     defp on_mount(params, %Phoenix.LiveView.Socket{} = socket) do
       Context.set_extra_context(%{socket_id: socket.id})
-      Context.set_request_context(%{url: socket.host_uri})
+
+      if uri = socket.host_uri do
+        Context.set_request_context(%{url: URI.to_string(uri)})
+      end
 
       Context.add_breadcrumb(%{
         category: "web.live_view.mount",
@@ -231,8 +234,15 @@ if Code.ensure_loaded?(Phoenix.LiveView) do
 
     defp get_connect_info_if_root(socket, key) do
       case socket.parent_pid do
-        nil -> get_connect_info(socket, key)
-        pid when is_pid(pid) -> nil
+        nil ->
+          try do
+            get_connect_info(socket, key)
+          rescue
+            _ -> nil
+          end
+
+        pid when is_pid(pid) ->
+          nil
       end
     end
 
--- a/lib/sentry/live_view_hook.ex
+++ b/lib/sentry/live_view_hook.ex
@@ -12,12 +12,21 @@ if Code.ensure_loaded?(Phoenix.LiveView) do
       * The user agent and user's IP address
       * Breadcrumbs for events that happen within LiveView
 
-    To make this module work best, you'll need to fetch information from the LiveView's
-    WebSocket. You can do that when calling the `socket/3` macro in your Phoenix endpoint.
-    For example:
+    To get the most complete request context, you'll need to fetch information from
+    the LiveView's WebSocket. You can do that when calling the `socket/3` macro in your
+    Phoenix endpoint. For example:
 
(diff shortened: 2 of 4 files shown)


solnic (approved):
Thanks for tackling this 🙇🏻

PR #1066, opened 2026-05-20
fix(test): handle deduped events when collecting

### LINKED ISSUE #1065: Message events no longer deduped
### Issue Description

Hello! When upgrading from 12.0.2 to 13.0.1 in one of my projects I found that the deduplication behavior changed. Before, `Sentry.Test.pop_sentry_reports()` returned 1 event in my test, now it returns multiple events

### Reproduction Steps

that particular test can be reduced to something like
```elixir
:ok = Sentry.Test.start_collecting_sentry_reports()

Enum.each(1..10, fn _ ->
  Sentry.capture_message("Some Message",
    extra: %{some: %{data: "structure"}},
    tags: %{some: "tags"}
  )
end)

errors = Sentry.Test.pop_sentry_reports()
assert length(errors) == 1
```

### Expected Behavior

On Sentry 12.0.3 the test passes. I assume that was the correct behavior because I couldn't find the change in the changelog

### Actual Behavior

on Sentry 13.0.1 the test fails

### Elixir Version

1.19.4

### SDK Version

13.0.1

### Integration and Its Version

_No response_

### Sentry Config

```
config :sentry,
  environment_name: Mix.env(),
  enable_source_code_context: true,
  root_surce_code_paths: [File.cwd!()],
  filter: Sentry.DefaultEventFilter,
  dedup_events: true,
  test_mode: true
```

solnic:
Fixes a regression in `Sentry.Test` that was introduced in 13.0.0 where duped events would be collected even when `dedup_events: true` was set.

Another thing that this addresses is adding `nimble_ownership` as a regular dep (not only in `test` env) - which is something that was reported in the comment thread under the issue.

Closes #1065


michallepicki:
I don't think `nimble_ownership` needs to be a regular dep if it's only used in `Sentry.Test` code. Following that logic would `bypass` also become a regular dep now? I don't want to have to ship cowboy in my bandit based Phoenix project :)

Instead, if `start_collecting_sentry_reports` is going to be still supported, I would suggest adding the same helpful runtime errors to it that `setup_sentry` has, basically [this line](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/test.ex#L136) to improve DX

solnic:
> I don't think `nimble_ownership` needs to be a regular dep if it's only used in `Sentry.Test` code

I 100% agree - unfortunately this is how it used to be in pre-`13.0.0` releases and because I added it back, the dep needs to be back too, otherwise it'd be a breaking change and I'd have to release `14.0.0` already 😢 

> Following that logic would `bypass` also become a regular dep now? I don't want to have to ship cowboy in my bandit based Phoenix project :)

Nope. We happen to use bypass but it's going away in `14.0.0` in favor of a new built-in abstraction (possibly a test transport or maybe just a test http client that captures requests for assertions, not sure yet).

> Instead, if `start_collecting_sentry_reports` is going to be still supported, I would suggest adding the same helpful runtime errors to it that `setup_sentry` has, basically [this line](https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/test.ex#L136) to improve DX

Thank you. This crossed my mind but then I realized that, as I explained above, sticking to `only: [:test]` would be a breaking change. Existing test suites that use `13.0.0` must continue to work when upgrading to `13.1.0`.

dingsdax (approved):

PR #1064, opened 2026-05-20
fix(oban): don't assume stacktrace is always a list

solnic:
Fixes #1063 

I haven't figured out where this is coming from, but here's a fix 😄

PR #1071, opened 2026-05-27
fix(oban): handle non-list stacktraces in Oban error reporter

whatyouhide:
### Problem

When an Oban job dies from a **gen-call exit**, Oban surfaces the `{module, function, args}` from the exit reason in the `:stacktrace` field of the `[:oban, :job, :exception]` telemetry metadata, and **not** a stacktrace list.

`Sentry.Integrations.Oban.ErrorReporter.report/5` passes that value straight into Sentry's `:stacktrace` option, which is typed as a list. This results in an error (that we saw in production):

```
** (NimbleOptions.ValidationError) invalid value for :stacktrace option: expected list, got: {NimblePool, :checkout, [#PID<0.493.0>]}
    (sentry) lib/sentry/event.ex:186: Sentry.Event.create_event/1
    (sentry) lib/sentry.ex: Sentry.capture_message/2
    (sentry) lib/sentry/integrations/oban/error_reporter.ex:114: Sentry.Integrations.Oban.ErrorReporter.report/5
```

The handler raises, and `:telemetry` then **permanently detaches it**. The affected node stops reporting *all* Oban job errors to Sentry until it restarts, so the first such exit silently creates an observability blind spot.

The existing normalization at the top of `report/5` only handled the empty-list (`[]`) case; any other non-list value fell straight through.

### DIFF
--- a/lib/sentry/integrations/oban/error_reporter.ex
+++ b/lib/sentry/integrations/oban/error_reporter.ex
@@ -78,6 +78,13 @@ defmodule Sentry.Integrations.Oban.ErrorReporter do
   end
 
   defp report(job, kind, reason, stacktrace, config) do
+    # Oban can hand us a non-list "stacktrace" when a job dies from a gen-call
+    # exit (such as a GenServer.call or NimblePool checkout timeout): the value
+    # is the `{module, function, args}` from the exit reason rather than a
+    # stacktrace. Sentry's :stacktrace option only accepts a list, so anything
+    # else would crash this handler (and get it detached). Coerce it away.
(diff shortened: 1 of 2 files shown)


whatyouhide:
Ah man @solnic just saw now that this overlaps https://github.com/getsentry/sentry-elixir/pull/1064 (which I had not seen). Tests and comments here are a bit more thorough but up to you.

solnic (approved):

solnic:
@whatyouhide no worries, great that we know what is causing it!

PR #1073, opened 2026-06-01
Exclude :bandit domain by default

dinokovac-river:
### Description

Like for `:cowboy`, we want to exclude the `:bandit` domain to avoid duplicate errors.

bandit is a drop-in replacement for cowboy and has been the default web server for new phoenix apps for a while now:
https://github.com/mtrudel/bandit

### DIFF
--- a/lib/sentry/logger_handler.ex
+++ b/lib/sentry/logger_handler.ex
@@ -26,7 +26,7 @@ defmodule Sentry.LoggerHandler do
     ],
     excluded_domains: [
       type: {:list, :atom},
-      default: [:cowboy],
+      default: [:cowboy, :bandit],
       type_doc: "list of `t:atom/0`",
       doc: """
       Any messages with a domain in the configured list will not be sent. The default is so as


whatyouhide (approved):

whatyouhide:
Thank you @dinokovac-river 🫶

martosaur:
@dinokovac-river @whatyouhide  Hi guys, a little late for the party but if I recall correctly we do not need to exclude `:bandit` domain. Duplicated errors come from using `PlugCapture` which is only needed for cowboy applications to workaround very specific cowboy quirks. This is highlighted in [the docs](https://sentry.hexdocs.pm/13.2.0/Sentry.PlugCapture.html). See https://github.com/getsentry/sentry-elixir/pull/900 and https://github.com/getsentry/sentry-elixir/issues/895 for the writeup.

dinokovac-river:
@martosaur thanks, totally missed that, sorry!
@whatyouhide should we revert the change then?

solnic:
Hey folks, thanks for dealing with this. I reckon we could do better and have `PlugCapture` emit a warning when it's added to a bandit-based app - WDYT? 😄

martosaur:
Oh that'd be great, especially if possible to do at compile time!

PR #1088, opened 2026-06-19
fix: improve Ranch errors in Sentry.LoggerHandler

whatyouhide:
### Description

These changes mean that we get much better errors when the range is stack trace and reason include an exception for example. We've actually seen this happen in production quite a bit, and it's a pretty annoying error to debug and whatnot.


solnic (approved):

solnic:
@whatyouhide any thoughts on expanding CI test suite with a matrix that includes different versions of popular (and/or critical) dependencies? ie we'd have cowboy, bandit, phoenix and ecto there with different major/minor combinations.

whatyouhide:
I personally wouldn't go that far 😬

PR #1151, opened 2026-08-05
fix(rate-limiting): drop only the rate-limited attachments from an envelope

solnic:
Rate-limit enforcement now happens before serialization and can drop items individually (as opposed to dropping entire envelope just because some attachments need to be dropped).

- `attachment` / `attachment_item` limited - send the error, drop its attachments
- `error` or global limit - drop the whole envelope, attachments included
- filtered down to zero items - send nothing

`ratelimit_backoff` client reports now describe exactly the items discarded rather
than the whole envelope.

Also adds `RateLimiter.global_rate_limited?/0` and makes `attachment` gate on `attachment_item` as well.


whatyouhide (approved):
Other than the BugBot comment this looks good.

PR #1162, opened 2026-08-07
test(tracing): add e2e scenarios for async trace continuation

solnic:
You can have as many unit tests as you want, but here's the real-deal coverage. This is also useful when doing manual smoke testing.


solnic:
Superseded by a PR with the correct base — a fix branch for #1166 was inserted beneath this one and GitHub's stacked-PR feature locks base changes.

PR #1187, opened 2026-09-03
refactor(config): remove `enable_metrics` toggle

solnic:
This removes `enable_metrics` toggle from the config, so calls to `Sentry.Metrics` will always send data to Sentry.


whatyouhide (approved):
Nice!

PR #1196, opened 2026-09-07
fix(scrubbing): apply the url scrubber to conn path and query fields …

### LINKED ISSUE #1195: Conn structs in stacktrace frame vars leak credentials via `query_string`, `request_path` and `path_info`
**SDK version:** 13.5.1 (`master` as of 2026-09-07); same on 13.3.0
**Setup:** Phoenix 1.7, `Sentry.PlugCapture` + `Sentry.PlugContext` with `url_scrubber`, `body_scrubber` and `header_scrubber` set

## Summary

Follow-up to #477 / #1068. A `%Plug.Conn{}` in a stacktrace frame arg is scrubbed with `Sentry.Scrubber.scrub/1` before being inspected into a frame var, but that scrub misses two things the request interface covers:

- `request_path` and `path_info` are not in `@scrubbable_conn_fields`, so a credential in a path segment (`/reset-password/:token`) renders verbatim. The registered `:url_scrubber` never runs on this path.
- `query_string` and `query_params` use the fixed `:query_string` / `:params` strategies, which only know the default keys (`password`, `passwd`, `secret`). `put_conn_scrubber/1` cannot extend them.

`scrub/2` takes per-call overrides, but `PlugCapture` calls `scrub/1` with none, so there is no hook. An app can register scrubbers for every credential it mints and still ship a working reset token to Sentry when a tokenized request raises with the conn in the top frame.

## Reproduction

```elixir
token = "SEKRIT-TOKEN-VALUE"

Plug.Test.conn(:get, "/reset-password/#{token}")
|> Sentry.Scrubber.scrub()
|> inspect()
|> String.contains?(token)
#=> true (request_path, path_info)

Plug.Test.conn(:get, "/unsubscribe?token=#{token}")
|> Sentry.Scrubber.scrub()
|> inspect()
|> String.contains?(token)
#=> true (query_string, query_params), whatever scrubbers are registered
```

Since #1093 removed SDK-side truncation, the frame var is not cut short before the value either.

## Suggestion

1. Run the registered `:url_scrubber` (or its default) over `request_path`, `path_info` and `query_string` when scrubbing a conn.
2. Let the registered conn scrubber reach the frame-var path, either by extending the key list the `:query_string` / `:params` strategies use, or by having `PlugCapture` pass overrides to `scrub/2`.

## Workaround

A `before_send` that regex-redacts `<key>=<value>`, `"<key>" => "<value>"` and the app's tokenized path segments out of the rendered frame vars. Works, but it is the wrong layer.

solnic:
Now `:url_scrubber` also handles the conn's own `request_path`, `path_info` and `query_string`, wherever the connection itself is reported. `path_info` is rebuilt against `script_name`, so forwarded routers stay correct.

### Before

The app's scrubber redacts the token segment of the URL, but the conn embedded in the exception still carries the raw path:

- `request_path` — `/scrubbing-demo/forwarded/reset-password/pathsecret-...`
- `path_info` — the secret as its own segment
- `query_string` — every parameter in the clear

pr1-conn-dump-before

### After

The same three fields now come from the scrubbed URL:

- `request_path` and `path_info` — the token segment replaced with the placeholder
- `script_name` — still `["scrubbing-demo", "forwarded"]`, showing `path_info` was rebuilt against the mount point rather than the whole path
- `query_string` — redacted alongside them

pr1-conn-dump-after

---

Closes #1195

### DIFF
--- a/lib/sentry/scrubber.ex
+++ b/lib/sentry/scrubber.ex
@@ -50,17 +50,19 @@ defmodule Sentry.Scrubber do
       default `scrub(conn, field)` clause when none is registered, or
     * a fixed tag — `:clear` replaces the field with `%{}`, `:params` scrubs the
       field as a params-shaped map, `:query_string` redacts sensitive params from
-      a raw query string, and `:private_allow_list` keeps only the registered
-      allow-listed keys of the field (see `default_private_allow_list/0` and the
-      `:private_allow_list` option of `put_conn_scrubber/1`), dropping everything
-      else.
+      a raw query string, `:url_scrubbed` derives the field from the URL the
+      registered `:url_scrubber` returns, and `:private_allow_list` keeps only
+      the registered allow-listed keys of the field (see
+      `default_private_allow_list/0` and the `:private_allow_list` option of
+      `put_conn_scrubber/1`), dropping everything else.
 
   By default `scrub/1` redacts `cookies`, `req_headers`, `params`, and
   `body_params` (the configurable fields — `body_params` shares the
   `:body_scrubber` with `params`, so it honors the same registered scrubber and
   is emptied when `body_scrubber` is `nil`), clears `req_cookies` and `assigns`
-  to `%{}`, scrubs `query_params` as a params-shaped map, and reduces `private`
-  to its allow-listed keys (`default_private_allow_list/0`). `assigns` is cleared
+  to `%{}`, scrubs `query_params` as a params-shaped map, derives `request_path`,
+  `path_info` and `query_string` from the scrubbed URL, and reduces `private` to
+  its allow-listed keys (`default_private_allow_list/0`). `assigns` is cleared
   wholesale because auth libraries (Guardian, Pow, Coherence) routinely store
   decoded tokens, full user structs, and session data there, where no key-based
   heuristic redacts safely. `private` keeps only the allow-listed framework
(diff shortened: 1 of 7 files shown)


lib/sentry/scrubber.ex
```diff
@@ -548,6 +572,57 @@ defmodule Sentry.Scrubber do
...
+  # when the scrubber hands back a non-binary, in which case every
+  # `:url_scrubbed` field keeps the conn's own value: scrubbing runs while an
+  # error is already being reported, so a broken user scrubber must not take the
+  # report down with it. `URI.parse/1` needs no such guard — it returns a `%URI{}`
+  # for any binary, and unparseable input lands in `:path`, which over-redacts
+  # rather than under-redacts.
+  defp scrubbed_uri(conn) do
+    case get(:url_scrubber).(conn) do
+      url when is_binary(url) -> URI.parse(url)
+      _other -> nil
```
whatyouhide, line 619: Is this the same behavior we had before? It's a pretty dangerous silent failure mechanism so I’m wondering if we'd at least need a warning log.

whatyouhide (approved):

lib/sentry/scrubber.ex
solnic: @whatyouhide ah great catch - this is now fixed via a658d85c443ba4692afa3083bdf375489143ff8b - I also added rescuing from potential failures which follows Sentry spec about user-provided callbacks.

solnic: edit: I rebased the stack so the fixup commit is 665d72367677bdaa07cc87a8d9b8d68e0bc6301f now, sorry for the noise

sl0thentr0py (approved):

PR #1198, opened 2026-09-07
feat(scrubbing): adopt denylist defined by Data Collection spec

solnic:
Now the default sensitive keys are the denylist the Data Collection spec requires, matched as a case-insensitive substring, and `config :sentry, scrubber: [param_keys: [...]]` extends that list with your own terms.

### Before

`internal_ref` is the row to focus on. Relay does not recognise that name, so it is the one value the panel shows exactly as the SDK left it — and there is no way to make the SDK redact it, because the `:param_keys` option does not exist yet:

- `internal_ref` — `cfgref-20260909-082441-before`, in both Body and Query String
- `token`, `api_key` — not in the old denylist
- `PASSWORD` — the same term as `password`, **missed because matching was case-sensitive**
- `user_password` — the same term again, missed because matching required exact equality

pr3-conn-dump-before

pr3-request-panel-before

### After

Every one of those is redacted, each by a different rule:

- `internal_ref` — `*********`, sensitive only because the app configured `scrubber: [param_keys: ["internal_ref"]]`
- `token` — a key the spec denylist adds
- `api_key` — matched as a substring of `key`
- `PASSWORD` — matched too
- `user_password` — matched as a substring

pr3-conn-dump-after

pr3-request-panel-after

### DIFF
--- a/lib/sentry/config.ex
+++ b/lib/sentry/config.ex
@@ -613,6 +613,32 @@ defmodule Sentry.Config do
       *Available since v13.2.0*.
       """,
       keys: [
+        param_keys: [
+          type: {:list, :string},
+          default: [],
+          type_doc: "list of `t:String.t/0`",
+          doc: """
+          Additional sensitive parameter keys, redacted wherever the SDK scrubs a
+          map or a query string: request and query params, the request URL,
+          `Sentry.LiveViewHook` breadcrumb data, Oban job args, and maps captured
+          into stacktrace frame variables.
+
+          These *extend* the SDK default (see `Sentry.Scrubber.default_param_keys/0`),
+          which is the denylist required by the
+          [Sentry Data Collection spec](https://develop.sentry.dev/sdk/foundations/client/data-collection/).
+          The spec defines custom deny-mode terms as additive, so there is no way
+          to shrink the default list — that is deliberate.
+
+          Terms are matched as case-insensitive substrings of the key name, so
+          `"ref"` redacts `"internal_ref"` and `"REF_ID"` alike.
+
+          This does not affect header scrubbing, which uses its own list — see
+          `Sentry.Scrubber.default_header_keys/0` and the `:header_scrubber`
+          option of `Sentry.PlugContext`.
+
+          *Available since 14.0.0*.
+          """
+        ],
         conn_private_allow_list: [
           type: {:list, :atom},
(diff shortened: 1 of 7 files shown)


lib/sentry/config.ex
```diff
@@ -627,6 +627,32 @@ defmodule Sentry.Config do
...
+          to shrink the default list — that is deliberate.
+
+          Terms are matched as case-insensitive substrings of the key name, so
+          `"ref"` redacts `"internal_ref"` and `"REF_ID"` alike.
+
+          This does not affect header scrubbing, which uses its own list — see
+          `Sentry.Scrubber.default_header_keys/0` and the `:header_scrubber`
+          option of `Sentry.PlugContext`.
+
+          *Available since 14.0.0*.
```
whatyouhide, line 639: Are we gonna have breaking changes because of this, since we're bumping the major version?

lib/sentry/config.ex
```diff
@@ -627,6 +627,32 @@ defmodule Sentry.Config do
...
+          Additional sensitive parameter keys, redacted wherever the SDK scrubs a
+          map or a query string: request and query params, the request URL,
+          `Sentry.LiveViewHook` breadcrumb data, Oban job args, and maps captured
+          into stacktrace frame variables.
+
+          These *extend* the SDK default (see `Sentry.Scrubber.default_param_keys/0`),
+          which is the denylist required by the
+          [Sentry Data Collection spec](https://develop.sentry.dev/sdk/foundations/client/data-collection/).
+          The spec defines custom deny-mode terms as additive, so there is no way
+          to shrink the default list — that is deliberate.
```
whatyouhide, line 630: Generally em-dashes are not spaces (en dashes are). Sadly agents love to space around em dashes


```suggestion
          to shrink the default list—that is deliberate.
```

lib/sentry/config.ex
```diff
@@ -627,6 +627,32 @@ defmodule Sentry.Config do
...
+
+          Terms are matched as case-insensitive substrings of the key name, so
+          `"ref"` redacts `"internal_ref"` and `"REF_ID"` alike.
+
+          This does not affect header scrubbing, which uses its own list — see
```
whatyouhide, line 635: Same here

whatyouhide (approved):

sl0thentr0py (approved):

PR #1200, opened 2026-09-09
fix(scrubbing): leave query params the SDK does not redact untouched

solnic:
Now `scrub_query_string/2` rewrites a parameter only when it redacts its value, so we won't end up with `%2A`, that would violate DC specs.

### Before

- `query_string` — the one redacted parameter reads `password=%2A%2A%2A%2A%2A%2A%2A%2A%2A`

pr4-conn-dump-before

### After

- `query_string` — the placeholder intact and everything the SDK keeps passed through

pr4-conn-dump-after

### DIFF
--- a/lib/sentry/scrubber.ex
+++ b/lib/sentry/scrubber.ex
@@ -332,15 +332,37 @@ defmodule Sentry.Scrubber do
     keys = param_keys(opts)
 
     query
-    |> URI.query_decoder()
-    |> Enum.map(fn {key, value} ->
-      cond do
-        sensitive_key?(key, keys) -> {key, @scrubbed_value}
-        is_binary(value) and value =~ credit_card_regex() -> {key, @scrubbed_value}
-        true -> {key, value}
-      end
-    end)
-    |> URI.encode_query()
+    |> String.split("&")
+    |> Enum.map_join("&", &scrub_query_pair(&1, keys))
+  end
+
+  # Only a redacted pair is rewritten. Everything the SDK keeps is passed
+  # through exactly as it arrived, so the reported query string still matches
+  # what the client sent rather than a re-encoding of it.
+  defp scrub_query_pair(pair, keys) do
+    case String.split(pair, "=", parts: 2) do
+      [raw_key, raw_value] ->
+        if redact_param?(decode_www_form(raw_key), decode_www_form(raw_value), keys) do
+          raw_key <> "=" <> @scrubbed_value
+        else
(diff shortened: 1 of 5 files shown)


lib/sentry/scrubber.ex
```diff
@@ -332,15 +332,37 @@ defmodule Sentry.Scrubber do
...
-    |> URI.query_decoder()
-    |> Enum.map(fn {key, value} ->
-      cond do
-        sensitive_key?(key, keys) -> {key, @scrubbed_value}
-        is_binary(value) and value =~ credit_card_regex() -> {key, @scrubbed_value}
-        true -> {key, value}
-      end
-    end)
-    |> URI.encode_query()
+    |> String.split("&")
```
whatyouhide, line 335: Are you not using `URI.decode_query/1` here (and below) because it'd lose ordering?

lib/sentry/scrubber.ex
```diff
@@ -332,15 +332,37 @@ defmodule Sentry.Scrubber do
...
+  defp redact_param?(key, value, keys) do
+    sensitive_key?(key, keys) or value =~ credit_card_regex()
+  end
+
+  # Scrubbing runs while an error is already being reported, so malformed
+  # percent-encoding must not raise. The raw form is still fine to match on.
+  defp decode_www_form(value) do
+    URI.decode_www_form(value)
+  rescue
+    ArgumentError -> value
```
whatyouhide, line 365: When does this raise?

whatyouhide (commented):
Just some comments here.

lib/sentry/scrubber.ex
solnic: Yes we want to preserve original state, order included, and also prevent cases like ending up with `*********` placeholder being converted to `%2A%2A%2A%2A%2A%2A%2A%2A%2A`.

lib/sentry/scrubber.ex
solnic: Good question - in my testing nowhere, but BugBot pointed out that if it did, we'd have an unwanted crash. It is very defensive, I have to admit - do you think it's 100% safe to remove that rescue?

whatyouhide: [I'd say so yes](https://github.com/elixir-lang/elixir/blob/v1.20.2/lib/elixir/lib/uri.ex#L483-L511).

sl0thentr0py (approved):

PR #1204, opened 2026-09-10
fix(telemetry_processor): flush buffered telemetry on shutdown

solnic:
Now Sentry flushes buffered telemetry during graceful application shutdown, before the telemetry processor and HTTP client stop.

The flush waits up to five seconds before allowing shutdown to continue.

This is to conform to the spec's requirement:

> When the application shuts down gracefully, the telemetry buffer MUST forward all data in memory to the telemetry scheduler.

See https://develop.sentry.dev/sdk/foundations/processing/telemetry-processor/#data-forwarding-scenarios

### DIFF
--- a/lib/sentry/test.ex
+++ b/lib/sentry/test.ex
@@ -615,9 +615,14 @@ defmodule Sentry.Test do
 
   ## Options
 
-    * `:type` - when set, only envelopes containing an item of this type
-      (e.g., `"event"`, `"transaction"`, `"log"`) are forwarded to the test
-      process. Envelopes not matching the type are silently dropped.
+    * `:type` - when set to a type or list of types, only envelopes containing a matching
+      item (e.g., `"event"`, `"transaction"`, `"log"`) are forwarded to the test
+      process. Envelopes not matching the type are silently dropped. Passing a list
+      of types is *available since 14.0.0*.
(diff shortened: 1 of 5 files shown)


lib/sentry/test.ex
```diff
@@ -615,9 +615,13 @@ defmodule Sentry.Test do
...
 
-    * `:type` - when set, only envelopes containing an item of this type
-      (e.g., `"event"`, `"transaction"`, `"log"`) are forwarded to the test
+    * `:type` - when set to a type or list of types, only envelopes containing a matching
+      item (e.g., `"event"`, `"transaction"`, `"log"`) are forwarded to the test
       process. Envelopes not matching the type are silently dropped.
+    * `:response` - optional one- or two-argument function receiving the `Plug.Conn` and,
+      when it accepts two arguments, the raw envelope body after collection. It must return
+      the response connection. Use it to simulate delayed or failed HTTP responses.
+      Defaults to a successful response.
```
whatyouhide: Would it be simpler to have just the two-argument version and ask users to use `conn, _envelope` instead? Also, should we mark this as "Available since ..."?

solnic: Done :)

sl0thentr0py (approved):

PR #1208, opened 2026-09-14
feat(telemetry_processor): enable all categories by default

solnic:
This sets telemetry processor categories to `[:error, :check_in, :transaction]` by default so that they can join `:log` and `:metric` group of event types that the SDK sends via `TelemetryProcessor`.

This is the minimal change to get them all handled by TP, full switch with no opt-out + related clean up will come in `15.0.0`.


lib/sentry/config.ex
```diff
@@ -552,7 +552,7 @@ defmodule Sentry.Config do
     ],
     telemetry_processor_categories: [
       type: {:list, {:in, [:error, :check_in, :transaction, :log]}},
```
sl0thentr0py, line 554: not sure I follow the changes, why is metric not in here?

solnic: @sl0thentr0py metrics always went via Telemetry Processor

sl0thentr0py (approved):

lib/sentry/config.ex
```diff
@@ -552,7 +552,7 @@ defmodule Sentry.Config do
...
-      default: [],
+      default: [:error, :check_in, :transaction],
```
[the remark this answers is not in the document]
solnic: TelemetryProcessor drops on overflowing buffers, that's the new behavior and it's compliant with the TelemetryProcessor spec with one caveat that the spec does not describe which items should be dropped, so here we drop the oldest.

